PR #1602, opened 2014-07-16
Don't include a SECRET_KEY in the repo

jean:
This key *will* be used as-is by newbies.
Force people to create a new key.

### DIFF
--- a/README
+++ b/README
@@ -62,6 +62,14 @@ here are the instructions for Ubuntu 12.04::
     # Compile GeoServer
     paver setup
 
+Before proceeding to the next step, edit ``geonode/settings.py`` 
+to unquote the line ``# SECRET_KEY = PLEASE DEFINE ME``,
+and change the placeholder text to a valid quoted string. 
+The key should be around 50 random alphanumeric characters.
+If you skip this step, startup will fail.
+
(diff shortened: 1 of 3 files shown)


ingenieroariel:
I am not too happy about requiring our users to come up with a random secret key before they start looking at GeoNode. Django does not require this for end users.

My preference would be for us to move to a workflow in which django-admin startproject is mandatory, and that would produce a fresh SECRET_KEY.

If others think it is a good idea, here is the workflow I suggest:

a) We change the serious looking key in settings.py to SECRET_KEY = 'your-site-could-get-hacked-if-you-do-not-override-this'
b) We move in 2.5 to a workflow that includes django-admin startproject instead of running a default one.
c) We make the documentation change based on ian's suggestion.

What do you think?

ischneider:
I also like the idea of putting the secret key in a single, external file and including it. Another quality options is to pull it in from the environment.

jean:
> My preference would be for us to move to a workflow in which django-admin
> startproject is mandatory, and that would produce a fresh SECRET_KEY.

Agreed, this is a good workflow.

--
jean                                              . .. .... //\\\oo///\\

PR #1611, opened 2014-07-30
Provide access to non-spatial data associated with a layer

gamesbook:
This pull request allows for additional data to be extracted from specific types of layers.

The initial motivation for this change stemmed from the need to spatially represent a Sensor Observation Service.  This can be done at a spatial database level, by creating a WFS view over an underlying Sensor Observation data.  In this view, individual sensors form features.  While some aspects of sensors can be handled through the feature attributes, what is missing is the ability to access time series data.

This pr sets up a 'layer_data' view which is enabled through settings in the layer's metadata of (a) the correct keyword for the layer (in this case: SOS), as well (b) a dictionary in the supplemental information field - this would look something like:
```
{"sos_url":"http://sos.server.org/test/sos",
"observedProperties":["urn:ogc:def:phenomenon:OGC:1.0.30:temperature"], 
"offerings":["WEATHER"]}
```
This pr should not impact on any existing code behaviour, as the code and template changes only take effect once the enablement described above happens.

This code is also designed to be extended in future to handle data from other sources; for example, a NetCDF file.

(We also have additional code that accesses this data to create, for example, time series charts but, based on previous discussions with the dev team, this will probably need to be housed in an app external to GeoNode core.)


geonode/layers/data.py
```diff
@@ -0,0 +1,111 @@
...
+        Instance is given as one time value. Periods of time (start and end) are
+        separated by "/". Example: 2009-06-26T10:00:00+01/2009-06-26T11:00:00+01
+
+    Returns
+    ------
+    HttpResponse with mimetype data, or None if an error is encountered.
+    """
+    sos_data = HttpResponse(content="Unable to process request", status=500)
+    offerings = supplementary_info.get('offerings')
+    url = sup_supplementary_info.get('sos_url')
```
jj0hns0n: This variable does not exist. It should probably be supplementary_info and this hints that the code may not be working properly.

Unfortunately we do not have an SOS server to test against and will have to pass on this.

PR #1635, opened 2014-08-06
Initial Preferences Features (File Formats)

state-hiu:
**Initial Preferences Features (File Formats)**

The included preferences features are part of a broader feature set that will enable users to specify their preferences for file formats, areas of interest, notifications, etc.  This will enable users to more seamlessly integrate GeoNode into their daily routine.

This PR adds preferences for metadata, vector, and raster file formats.  For example, most regular users of GeoNode have no need for GML 3.1.1.  Therefore, they should be given the option to hide it.  Other users might only use Google Earth, so they might want to hide the GeoJSON and "Zipped Shapefile" links.  Some organizations might require the use of a specific metadata format internally.  Finally, the user experience is fast right now, but as capabilities get added to GeoServer, showing all possible links will begin to overwhelm regular users.

Importantly, this user preferences filter is applied after the settings.py filters.  Also, the default when no preference is set, is to display all links (the current experience).

==
Fixes:
+ This PR also brings document metadata links under the settings.py DOWNLOAD_FORMATS_METADATA filter.


jj0hns0n:
@state-hiu closing this one. Can you reduce the scope per comments from @ingenieroariel above and reopen

state-hiu:
Except for PR #2004, there are no more fixes in this unmerged PR.  New features in code will be superseded by new TBD way for enabling user-level preferences.

Issue #1783, opened 2014-11-17
Wrong info displayed


al3mon:
--I'm on Documents tab, and when i try to share it, it shows "Share This Map" instead of "Share This Document"--

![share this map](https://cloud.githubusercontent.com/assets/9592424/5079354/8d1599e8-6e81-11e4-9a86-23ded652a57d.png)

PR #1841, opened 2014-11-26
Implementing sharing translation

teamgeode:
In ref to #1783 and #1802 

I made the translation, but I change the code to avoid gender problem for some language.

Can you check if you’re ok with the fact I put the sentence in the view file?


ingenieroariel:
Let's close this one and wait for a PR without .po and .mo for easier review. @teamgeode please take into account @capooti's last message in the new PR. Happy new year!

PR #1900, opened 2014-12-21
Dockerizing

milafrerichs:
## What does this PR do?
This PR adds a Dockerfile and fig file for easier development with the help of Docker. 
It is open for discussion and any comments are appreciated.
If someone is unfamiliar with: [fig](http://fig.sh/) or [docker](https://www.docker.com).
Using fig is much easier than using docker directly but might be replaced with a direct way in docker soon. 
### Possible Discussion/Changes
The pavement script could be replaced with additions into the docker file and another container running GeoServer. 
I did not tackle this just yet because I wanted to make sure you guys would like to have a dockerized way.
### Added
Option for start action of pavement script to run django in foreground. This is needed for Docker to detect something is running and does not quit. 
Dockerfile added to install all dependencies into the docker container.
Fig file for easier development, starting of docker etc.
### Changed
Update Readme
### Removed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:14.04
+
+RUN \
+  apt-get update && \
+  apt-get install -y build-essential && \
+  apt-get install -y libxml2-dev libxslt1-dev libjpeg-dev gettext git python-dev python-pip && \
+  apt-get install -y python-pillow python-lxml python-psycopg2 python-django python-bs4 python-multipartposthandler transifex-client python-paver python-nose python-django-nose python-gdal python-django-pagination python-django-jsonfield python-django-extensions python-django-taggit python-httplib2 && \
+  apt-get install -y --force-yes openjdk-6-jdk ant maven2 --no-install-recommends
+
+WORKDIR /geonode
+ADD . /geonode
+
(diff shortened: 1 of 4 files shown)


ingenieroariel:
I like this patch because it is clean and does not force us to switch to docker right away. Thanks @milafrerichs

ischneider:
For development purposes, this is a nice contribution but perhaps we should not recommend this for production use.

A useful enhancement would be to split provisioning a base image create a base image.

> I like this patch because it is clean and does not force us to switch to docker right away

Hopefully there will never be a time when we're `forced` to use docker :smile:

ingenieroariel:
I agree with you that this should not be recommended for production use.
(and agree with the last point too!)

milafrerichs:
Thanks for merging. 
And yes this is not meant for production use. 
If you want I can work on a production version.

gamesbook:
Our experience in "Dockerizing" GeoNode is that having GeoNode and the custom version of GeoServer in a single container is the way to go.  My experience of trying to host them in separate containers shows its feasible but creates all kinds of stresses on deployment and configuration (at least for the dev installation and usage).  An "official" version maintained by the dev team would be awesome; we'd be happy to share our approach if it would help create this tool.

PR #1918, opened 2015-01-15
Social Links Enhancements

state-hiu:
The social links are no longer hard coded, but added as a list to the settings.py file.  Admins can now easily remove or add their social networks of choice.  This will be useful for managing customization of multiple deployments.

Minor Changes
- prepended a mailto email link.
- context_processors.py defaults SOCIAL_BUTTONS to False if var is not found in settings.py.  Most settings.py files come with SOCIAL_BUTTONS included.
- included https check when sharing url

More importantly, this will open up a simple way to leverage "Web Intents" / APIs to integrate GeoNode services into other systems, such as discovery platforms, url shorteners, etc.  Are there other web intents that should be added to the default settings.py file?

### DIFF
--- a/geonode/documents/views.py
+++ b/geonode/documents/views.py
@@ -20,7 +20,7 @@ from geonode.base.models import TopicCategory, ResourceBase
 from geonode.documents.models import Document
 from geonode.documents.forms import DocumentForm, DocumentCreateForm, DocumentReplaceForm
 from geonode.documents.models import IMGTYPES
-from geonode.utils import format_urls
+from geonode.utils import build_social_links
 
 ALLOWED_DOC_TYPES = settings.ALLOWED_DOCUMENT_TYPES
 
@@ -91,13 +91,7 @@ def document_detail(request, docid):
             'related': related}
 
         if settings.SOCIAL_ORIGINS:
-            social_url = "{protocol}://{host}{path}".format(
-                protocol=("https" if request.is_secure() else "http"),
-                host=request.get_host(),
-                path=request.get_full_path())
-            context_dict["social_links"] = format_urls(
-                settings.SOCIAL_ORIGINS,
-                {'name': document.title, 'url': social_url})
+            context_dict["social_links"] = build_social_links(request, document)
 
         return render_to_response(
             "documents/document_detail.html",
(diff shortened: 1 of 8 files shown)


geonode/documents/views.py
```diff
@@ -83,15 +84,23 @@ def document_detail(request, docid):
...
             Document.objects.filter(id=document.id).update(popular_count=F('popular_count') + 1)
 
+
+        context_dict = {
+            'permissions_json': _perms_info_json(document),
+            'resource': document,
+            'imgtypes': IMGTYPES,
+            'related': related}
+
+        if settings.SOCIAL_ORIGINS:
```
simod: Would it be possible to avoid the repetition of this piece of code in the three views?

garnertb:
Can you document the `SOCIAL_ORIGINS` here: https://github.com/GeoNode/geonode/blob/master/docs/reference/developers/settings.txt

state-hiu:
SOCIAL_ORIGINS is now documented in https://github.com/GeoNode/geonode/blob/master/docs/reference/developers/settings.txt.

In regards to DRY, the only two keys for context_dict in common among the detail views for layers, maps, and documents are resource and permission_json.

```
'resource': document,
'permissions_json': _perms_info_json(document),
```
I could create a function in utility.py that returns an initial dict with those values.  However, I'd then need to add additional lines of code to append additional key-value pairs as needed, such as imgtypes, related, documents, metadata, is_layer, config, and layers.

simod:
sorry @state-hiu I was referring to this block https://github.com/GeoNode/geonode/pull/1918/files#diff-36bc05580e37aef61b66c480db47cc38R257 until to https://github.com/GeoNode/geonode/pull/1918/files#diff-36bc05580e37aef61b66c480db47cc38R264

state-hiu:
Okay, so should I add that to utils.py or create a new social/utils.py?

simod:
Something in utils should be fine. :) thanks!

simod:
although it's quite social specific....hm dunno if it's worth to create a social/utils.py file just for that but maybe better.

state-hiu:
Just added to utils.py for this PR.  Looking forward, I'm interested in refractoring/modularizing the utility functions during the code sprint.

Issue #1919, opened 2015-01-15
On Windows paver start fails


george-silva:
Hello guys and gals,
Following the setup instructions clearly, paver start command fails. It is still looking for /dev/null for start_geoserver task.

Check this line:

https://github.com/GeoNode/geonode/blob/master/pavement.py#L346

In my specific installation, the command ```java -version``` appears to work, so we never consider this to be on windows or not, maintaining the ```longernullpath = '/dev/null'```.

We should not rely on this behavior to determine if the user is on Windows or other OS, so, my proposal here is that we create a new function inside pavament.py that informs the script of the user's OS, and then determines the ```longernullpath```.

I plan to provide a patch. Not sure how I can provide unit tests for this, but I'll think of something.

garnertb:
Maybe you can use [os.devnull](https://docs.python.org/2/library/os.html#os.devnull)?

george-silva:
@garnertb that seems to be simpler! I'll use that approach!

jj0hns0n:
Can this now be closed?

george-silva:
sure
Em 02/02/2015 07:46, "Jeffrey Johnson" <notifications@github.com> escreveu:

nathanhilbert:
I can verify that this can be closed.  Can the "Windows" label be added to this for others to find in the future?

garnertb:
Closing, per @nathanhilbert's comments.

PR #1920, opened 2015-01-15
Checks if /dev/null actually exists, and if not, reset it to something Windows understands

george-silva:
This is a simple patch to solve bug issue #1919.

The change is simple. I kept the things that check Java support the same, but now we do not rely on the behavior of java being configured to set the loggernullpath. In a test I made with a Windows Server 2008, ```java -version``` worked, and by consequence, loggernullpath was still ```/dev/null```, causing ```paver start``` to fail, when we could not find that filestream.

Now, we just check ```os.devnull```. If it's equals to ```nul```, it means we are on Windows. I did not wanted to change line 344, that configures loggernullpath to be the output of ```os.devnull``` because I'm not aware of the consequences this might have on Mac OSX. So, this could be further optimized as (if someone can guarantee that it will work for Mac OSX - Linux I know it will work):

```python
# function start_geoserver
# ...
with pushd(data_dir):
        javapath = "java"
        loggernullpath = os.devnull

        # checking if our loggernullpath exists and if not, reset it to something manageable
        if loggernullpath == "nul":
            open("../../downloaded/null.txt", 'w+').close()
            loggernullpath = "../../downloaded/null.txt"

        try:
            sh(('java -version'))
        except:
            if not options.get('java_path', None):
                print "Paver cannot find java in the Windows Environment.  Please provide the --java_path flag with your full path to java.exe e.g. --java_path=C:/path/to/java/bin/java.exe"
                sys.exit(1)
            # if there are spaces
            javapath = 'START /B "" "' + options['java_path'] + '"'

        sh((
            '%(javapath)s -Xmx512m -XX:MaxPermSize=256m'
            ' -DGEOSERVER_DATA_DIR=%(data_dir)s'
            # workaround for JAI sealed jar issue and jetty classloader
            ' -Dorg.eclipse.jetty.server.webapp.parentLoaderPriority=true'
            ' -jar %(jetty_runner)s'
            ' --log %(log_file)s'
            ' %(config)s'
            ' > %(loggernullpath)s &' % locals()
        ))
# ...
```

That's pretty much it. I'm not sure how we can test this script. If anyone can point me to the right direction, I can create the UT.

### DIFF
--- a/pavement.py
+++ b/pavement.py
@@ -342,6 +342,12 @@ def start_geoserver(options):
     with pushd(data_dir):
         javapath = "java"
         loggernullpath = "/dev/null"
+
+        # checking if our loggernullpath exists and if not, reset it to something manageable
+        if os.devnull == "nul":
+            open("../../downloaded/null.txt", 'w+').close()
+            loggernullpath = "../../downloaded/null.txt"
+
         try:
             sh(('java -version'))
         except:
@@ -350,10 +356,6 @@ def start_geoserver(options):
                 sys.exit(1)
             # if there are spaces
             javapath = 'START /B "" "' + options['java_path'] + '"'
-            # cmd log file needs to exist in windows
-            # using folder from .gitignore
-            open("../../downloaded/null.txt", 'w+').close()
-            loggernullpath = "../../downloaded/null.txt"
 
         sh((
             '%(javapath)s -Xmx512m -XX:MaxPermSize=256m'
--- a/pavement.py
+++ b/pavement.py
@@ -341,10 +341,10 @@ def start_geoserver(options):
     # prevents geonode security from initializing correctly otherwise
     with pushd(data_dir):
         javapath = "java"
-        loggernullpath = "/dev/null"
+        loggernullpath = os.devnull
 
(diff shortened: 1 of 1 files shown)


garnertb:
On a mac:
```
$ python -c "import os; print os.devnull"
/dev/null
```

george-silva:
Good. I'll update this again to use os.devnull setting up loggernullpath.

pavement.py
```diff
@@ -341,7 +341,13 @@ def start_geoserver(options):
     # prevents geonode security from initializing correctly otherwise
     with pushd(data_dir):
         javapath = "java"
-        loggernullpath = "/dev/null"
+        loggernullpath = os.devnull
+
+        # checking if our loggernullpath exists and if not, reset it to something manageable
+        if loggernullpath == "nul":
```
ingenieroariel: Is that "nul" a typo?

garnertb:
Nope, I checked the docs after I saw that but its correct.

garnertb:
Merged in 340d6e0db6eaffeca37685028fe2081bed84335b.

PR #1921, opened 2015-01-15
Bump Django to 1.6.10.

garnertb:
@ingenieroariel  - This replaces `Django >=1.6.1, <=1.6.5` with `Django==1.6.10`.  Both @simod and I seem to remember that there was a reason why we had to version it that way but not the specific reason why.  Do you remember why?


ingenieroariel:
Thanks for updating it.

The reason is that we want to support the package that ships with Ubuntu
14.04 and pull the latest Django that we have tested against. All with the
goal to make deployment easier (less packages maintained by us).

-a

Issue #1922, opened 2015-01-15
Page explore Layers, missing edit data and edit style permission to layers


collaer:
In the  /layers/ page :

When editing permessions to the selected layers (button Set Permission), two options are missing : "edit data" and "edit style". This page is really usefull to set permission to several layers at once. 

regards,

julien

PR #1927, opened 2015-01-19
add perm choice for #1922

teamgeode:
PR in ref to #1922

### DIFF
--- a/geonode/layers/templates/layers/layer_list.html
+++ b/geonode/layers/templates/layers/layer_list.html
@@ -21,7 +21,9 @@
   {% endwith %}
   {% endwith %}
   {% endwith %}
+  {% with is_layer_list='true' %}
   {% include "_bulk_permissions_form.html" %}
+  {% endwith %}
 {% endblock %}
 
 {% block extra_script %}
--- a/geonode/layers/templates/layers/layer_list.html
+++ b/geonode/layers/templates/layers/layer_list.html
@@ -21,7 +21,7 @@
   {% endwith %}
   {% endwith %}
   {% endwith %}
-  {% with is_layer_list='true' %}
+  {% with is_layer='true' %}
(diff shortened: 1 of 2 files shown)


geonode/layers/templates/layers/layer_list.html
```diff
@@ -21,7 +21,9 @@ <h2 class="page-title">{% trans "Explore Layers" %}</h2>
   {% endwith %}
   {% endwith %}
   {% endwith %}
+  {% with is_layer_list='true' %}
```
simod: could we just use is_layer here so that we don't have to modify the other file?

teamgeode:
Yes we can.
Will change that if you want. But is_layer_list it's more to make a difference between one layer and a list of layer.

simod:
I would be +1 on just is_layer as from the permissions perspective there's no difference whether is one layer or many layers :)

simod:
@teamgeode are those layer bulk permissions actually working?
The issue is that the edit layer style and layer data permissions are set on Layer objects and not on ResourceBase while the bulk permissions view is only acting on ResourceBase, see here https://github.com/GeoNode/geonode/blob/master/geonode/security/views.py#L102. So I suspect that those two perms are not actually applied (and we don't have tests for those which should be added as well)

simod:
@teamgeode forgive me just noticed now this https://github.com/GeoNode/geonode/blob/master/geonode/security/models.py#L186 that should make everything working :)
Thanks!

teamgeode:
I was tested with some layer.

User_A make layer and no one have perm on it except viewing.
User_B try to modify style but can have the style panel.
User_A add User_B on the style perm list by list layer page.
User_B have access on the style panel.

That’s look working for me.

I will make more test tomorrow and tell you inform.

PR #1930, opened 2015-01-21
Web Intents / CKAN / Initial Support

state-hiu:
Building on https://github.com/GeoNode/geonode/pull/1918, this PR adds initial support for using "web intents" to register geospatial services with CKAN.  This PR adds support for the optional `abstract` querystring parameter to SOCIAL_ORIGINS.  Commented out by default, this PR adds a block to the settings.py file for managing CKAN "social links", to enable publishing.  To enable, remove the comment from the following line in the settings.py: `SOCIAL_ORIGINS.extend(CKAN_ORIGINS)`.  The commented out example in the settings.py file points to Humanitarian Data Exchange (https://data.hdx.rwlabs.org/).

Additional querystring parameters would need to be supported on the CKAN side to build out the workflow.

### DIFF
--- a/docs/reference/developers/settings.txt
+++ b/docs/reference/developers/settings.txt
@@ -403,6 +403,18 @@ Default::
 
 A list of dictionaries that is used to generate the social links displayed in the Share tab.  For each origin, the name and and url format parameters are replaced by the actual values of the ResourceBase object (layer, map, document).
 
+CKAN_ORIGINS
+--------------
+Default::
+
+    CKAN_ORIGINS = [{
+        "label":"Humanitarian Data Exchange (HDX)",
+        "url":"https://data.hdx.rwlabs.org/dataset/new?title={name}&notes={abstract}",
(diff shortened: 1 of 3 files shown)


state-hiu:
Are we clear to merge?  CKAN_ORIGINS are commented out by default.

simod:
Should we document somewhere this feature for who wants to link ckan?

state-hiu:
CKAN_ORIGINS is appended to SOCIAL_ORIGINS.  It never gets accessed independently outside of settings.py.  Should I document in https://github.com/GeoNode/geonode/blob/master/docs/reference/developers/settings.txt?

I'd hold off on broader documentation of this functionality until we get a better grasp of how GeoNode-CKAN integration is going to work in the future.

garnertb:
Yes, https://github.com/GeoNode/geonode/blob/master/docs/reference/developers/settings.txt seems like the appropriate place for now.

PR #1942, opened 2015-01-24
Layout Changes / Group Members, Avatar Change

state-hiu:
Layout Changes for groups members and avatar change pages.  Added tabs to filter by role and some styling to members list (see image below).  Minor style fixes to avatar change.

image

### DIFF
--- a/geonode/groups/models.py
+++ b/geonode/groups/models.py
@@ -134,8 +134,11 @@ class GroupProfile(models.Model):
     def join(self, user, **kwargs):
         if user == user.get_anonymous():
             raise ValueError("The invited user cannot be anonymous")
-        GroupMember.objects.get_or_create(group=self, user=user, **kwargs)
-        user.groups.add(self.group)
+        member, created = GroupMember.objects.get_or_create(group=self, user=user, defaults=kwargs)
+        if created:
+            user.groups.add(self.group)
+        else:
+            raise ValueError("The invited user \"{0}\" is already a member".format(user.username))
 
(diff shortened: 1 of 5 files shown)


garnertb:
@state-hiu, any chance you can also fix the messed up "User identifier" select2 box in this PR?  Also, "User identifier" should be renamed to something more human friendly.  If not, go ahead and merge this, its still a big improvement.

794a1ebe-a34e-11e4-8971-3e8d25933dac

state-hiu:
Not really sure how to fix that.  The `form-control` class at the select2-container is adding a `padding: 6px 12px;`.  When padding is 0 the input looks good.  The form is generate with: `{{ member_form|as_bootstrap }}`, so I don't have easy control over the styling.  Unless you know of an easy solution, I'd merge for now.

teamgeode:
I think the best way to resolve this layout issue is to place https://github.com/GeoNode/geonode/blob/master/geonode/groups/templates/groups/group_members.html#L31 into a div with an id.

      <div id="add_id_user_indentifiers">
         {{ invite_form|as_bootstrap }}
      </div>

And you modify the CSS by heritage like:

      #add_id_user_indentifiers div{
         padding: 0px;
      }

state-hiu:
The newest commit fixes the dropdown.  Can someone run the static build.  I'm not sure how to do it for release (I just use `make css`).  I've also fixed some logic (`get_or_create`) and added a check so you can't add existing members to the group.  You need to remove a member and add back if you want to elevate his/her from member to manager.

Issue #1826, opened 2014-11-24
Create a Country Filter for Layer / Map Explore Pages


jj0hns0n:
It would be nice to use the Region field and make a filter that appeared in the Layer and Map Explore pages.

state-hiu:
I'm going to close this issue.  Adding a spatial browse functionality or more advanced use of administrative hierarchies shoould be added to a new issue.

PR #1943, opened 2015-01-24
HOLD: Region Filter

state-hiu:
For issue #1826, this PR creates a region filter to be displayed right above the extent filter.  The filter is collapsed by default.  The filter uses an autocomplete field against the full list of regions.

@simod, any suggestions/issues?

**Need to Fix**
- [x] geonode/api/api.py
- [x] geonode/templates/search/_region_filter.html
- [ ] Merge TagResource and RegionResource

**Known Issues**
- [x] The text and region autocomplete inputs both work.  However, when I type into text field the text appears in the other one.  Not sure how this is happening.

**Future Enhancements**
- [ ] Filter autocomplete list by regions with count > 0
- [ ] Select multiple regions.  Display selections.
- [ ] Leverage region hierarchies for filtering

### DIFF
--- a/geonode/api/api.py
+++ b/geonode/api/api.py
@@ -9,6 +9,7 @@ from guardian.shortcuts import get_objects_for_user
 
 from geonode.base.models import ResourceBase
 from geonode.base.models import TopicCategory
+from geonode.base.models import Region
 from geonode.layers.models import Layer
 from geonode.maps.models import Map
 from geonode.documents.models import Document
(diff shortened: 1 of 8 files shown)


geonode/api/api.py
```diff
@@ -87,6 +87,42 @@ class Meta:
...
+            ctype = ContentType.objects.get_for_model(self.type_filter)
+            count = bundle.obj.taggit_taggeditem_items.filter(
+                content_type=ctype).filter(object_id__in=resources_ids).count()
+        else:
+            count = bundle.obj.taggit_taggeditem_items.filter(object_id__in=resources_ids).count()
+
+        return count
+
+    class Meta:
+        queryset = Tag.objects.all().order_by('name')
```
simod: why it's using the Tag as models? Shouldn't be Region?

geonode/templates/search/_region_filter.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}
+{% load base_tags %}
+<nav class="filter">
+
+    <h4><a href="#" class="toggle toggle-nav"><i class="fa fa-chevron-right"></i>{% trans "Regions" %}</a></h4>
+    <ul class="nav closed" id="regions">
+        <li>
+            <div class="input-group">
+            <input name="region_search_input" id="region_search_input" ng-model="text_query" type="text"
```
simod: ng-model="text_query" is the reason why the text appears in both places. You have to remove this ng-model.

geonode/api/api.py
```diff
@@ -87,6 +87,42 @@ class Meta:
         }
 
 
+class RegionResource(TypeFilteredResource):
```
simod: I would avoid the code duplication in this class and TagResource. At least the dehydrate_count method should be in common.

state-hiu:
Fixed binding and count.

@simod, the dehydrate_count logic for region and tag resources are quite different.

https://github.com/GeoNode/geonode/pull/1943/files#diff-1c999635c7de2fc5309948254e872d37R109

Issue #1873, opened 2014-12-01
GeoServer timeout (layers/acls)


menegon:
I'm testing GeoNode 2.4.0+beta4  (istalled by apt-get).
It works fine if there are just a few layers. However, if the number
of layers increase, GeoServer doesn't show any layer and the user is
not logged in.

The problem is the internal request for the acls
(http://localhost/layers/acls). The response is ok but it's too slow
and GeoServer doesn't wait.

jj0hns0n:
@menegon can we close this now that #1950 was merged. Guess we should test again on the latest packages?

jj0hns0n:
Lets look at this with @afabiani tomorrow when he is at the sprint.

simod:
@jj0hns0n can we close this one?

PR #1950, opened 2015-01-27
improved layer_acls performaces

menegon:
If the number of layers increases the view layer_acls becomes very slow. 
This patch significantly reduces the number of SQL queries using the polymorphic selectors/functions.
This partialy fixes the #1873 issue.

### DIFF
--- a/geonode/geoserver/views.py
+++ b/geonode/geoserver/views.py
@@ -18,6 +18,7 @@ from django.utils.translation import ugettext as _
 
 from guardian.shortcuts import get_objects_for_user
 
+from geonode.base.models import ResourceBase
 from geonode.layers.forms import LayerStyleUploadForm
 from geonode.layers.models import Layer, Style
 from geonode.layers.views import _resolve_layer, _PERMISSION_MSG_MODIFY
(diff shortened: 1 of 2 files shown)


simod:
@menegon could you fix the pep8 violations so that we can merge? thanks.

capooti:
@menegon if you have a long acls consider to cache it (that is what we do, actually)

capooti:
nice pr, by the way :)

geonode/geoserver/views.py
```diff
@@ -477,21 +478,16 @@ def layer_acls(request):
...
-        x.layer.typename for x in all_writable if x in all_readable and hasattr(
-            x,
-            'layer')]
+    # Include permissions on the anonymous user    
+    # use of polymorphic selectors/functions to optimize performances
+    layer_readable = get_objects_for_user(acl_user, 'view_resourcebase', ResourceBase.objects.instance_of(Layer)) 
+    layer_writable = get_objects_for_user(acl_user, 'change_layer_data', Layer.objects.all())
+
+    _read  = [l.typename for l in layer_readable.get_real_instances()]
+    _write = layer_writable.values_list('typename', flat=True)
```
ischneider: For the size of the `_read` and `_write` lists, it probably doesn't make a huge performance difference (especially when compared to the gain made by reducing DB queries) but if these were `set` instead of `list`, the operations below would be faster and could expressed simpler:

```python
read_only = read ^ write
write_only = read & write
```

menegon: I have modified the code, thanks!

menegon:
Sets are unordered collections thus the test_layer_acls doesn't work. It's need to change the test.

ischneider:
@menegon thanks for your work on this and sorry for adding more. 

The test seems kinda fragile anyway. Instead of using JSON object comparison to verify the response, it seems like checking the specific expected properties would be easier, e.g.

```python
self.assertEquals('bobby', response_json['name'])
```

And use a set operation for the layers to avoid dependence on order.

menegon:
.. and what about this:
self.assertEquals(sorted(expected_result), sorted(response_json))
it seems to work properly.

ischneider:
Works for me

PR #1954, opened 2015-01-29
research user-group insensitive and exclude AnonymousUser

teamgeode:
This PR permit to search user with insensitive case.

ins
(test on Manage Group Members page for a group)

At the same time I exclude AnonymousUser on the list of result.


simod:
@teamgeode thanks for this, could you just fix the pep8 violation?

Issue #1895, opened 2014-12-19
Text entered in Profile field doesn't show on profile page


ScottEvil:
When a user edits their profile, there is a field named "profile" where they're supposed to say something about themselves. Fill this in, save an go to the profile page. That text will not appear with the other profile information.  GeoNode 2.4

Issue #1792, opened 2014-11-18
Addition to the "Uploading Error" message on Documents


al3mon:
--It would be nice for the users if you specify wich documents are allowed to upload''--

![allowed](https://cloud.githubusercontent.com/assets/9592424/5080308/0064507e-6e8c-11e4-97d7-a56f4090d794.png)

PR #1985, opened 2015-02-03
added profile information to profile detail page

oware:
profile information displays if user is logged in. commit resolves issue #1895

### DIFF
--- a/geonode/documents/templates/documents/document_upload_base.html
+++ b/geonode/documents/templates/documents/document_upload_base.html
@@ -10,6 +10,7 @@
 <div class="page-header">
   <a href="{% url "document_browse" %}" class="btn btn-primary pull-right">{% trans "Explore Documents" %}</a>
   <h2 class="page-title">{% trans "Upload Documents" %}</h2>
+  <p>Allowed document types [doc, docx, gif, jpg, jpeg, ods, odt, odp, pdf, png, ppt, pptx, rar, tif, tiff, txt, xls, xlsx, xml, zip, gz]
 </div>
 <div class="row">    
   {% block body %}{% endblock body %}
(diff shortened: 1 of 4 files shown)


oware:
allowed document types listed. commit resolves #1792

geonode/documents/templates/documents/document_upload_base.html
```diff
@@ -10,6 +10,7 @@
 <div class="page-header">
   <a href="{% url "document_browse" %}" class="btn btn-primary pull-right">{% trans "Explore Documents" %}</a>
   <h2 class="page-title">{% trans "Upload Documents" %}</h2>
+  <p>Allowed document types [doc, docx, gif, jpg, jpeg, ods, odt, odp, pdf, png, ppt, pptx, rar, tif, tiff, txt, xls, xlsx, xml, zip, gz]
```
ischneider: Any way to get this list from the actual code/configuration?

capooti: I was thinking the same, but I don't think there is a better way other than setting this in a context variable or from context processor.
Does such an overhead justifies the effort?

garnertb:
I think so, since the admin can arbitrarily restrict this at anytime and the template will be incorrect.

geonode/documents/templates/documents/document_upload_base.html
ischneider: > I was thinking the same, but I don't think there is a better way other than setting this in a context variable or from context processor.

Using a templatetag would be the most straight-forward.

A simple tag could just spit out the list of extensions (as configured in settings/wherever).

jj0hns0n:
This PR has a bunch of extraneous stuff. @oware can you close and resubmit in a more atomic way.

PR #1993, opened 2015-02-03
Clean upload errors

### LINKED ISSUE #1753: Layer with error during upload was not removed.
Layer with error during upload was not removed from uploaded layers directory.  (Working on version 2.4a17 and Master)

Use case:
- Try to upload a layer shapefile with error in it.
- Geonode detect the error (that’s cool).
- Looking into geonode/uploaded/layers/ directory and you can see some files from the failed layer upload.

I think this was not a bug but can be considered like an enhancement to remove these files.

menegon:
This fix #1753


ingenieroariel:
Let's look more into this. After looking at errors in uploading I think it
is best to keep the upload session.

-a

Issue #1970, opened 2015-02-02
Improve paver for windows


nathanhilbert:
Improved errors
------
If a user runs paver start --java_path="C:/something/java.exe", everything runs fine.  If the user does a crtl-c on the django dev server, it closes and returns to the prompt, but does not kill geoserver's java instance.  When user tries to do paver start --java_path="C:/something/java.exe" they get a permission denied to downloads/null.txt because the file is already open using the first geoserver, and paver is trying to make sure the file exists to pass off to geoserver.
Suggestion is to check if downloads/null.txt is open at the os.devnull check, and suggest to user to run paver stop_geoserver with exit.  Alternative is to not start a second geoserver and use the one that is currently running.

Try to find java in environment variables
------------
Use os.environ["JRE_HOME"] to try to find java for user before requiring flag.

Check environment variables
--------------
Verify Geos.dll can be found and GDAL_LIBRARY_PATH environment variables is set. If not, throw error and don't start.

nathanhilbert:
Closed with PR #1996

PR #1996, opened 2015-02-03
Paver and settings improvements for Windows users

nathanhilbert:
This checks if the JAVA_HOME variable is set before complaining to add the --java_path flag to paver start. 
Also, running paver start the second time without doing a paver stop first would throw a "permission denied" to downloaded/null.txt.  Permission was denied because it was being used by the currently running Geoserver.  A better message is thrown now.
Both outlined in #1970

### DIFF
--- a/docs/tutorials/admin/install/quick_install.txt
+++ b/docs/tutorials/admin/install/quick_install.txt
@@ -66,12 +66,12 @@ Windows
 To install in Windows it is assumed you're familiar with python development and virtualenv on Windows and that you're familiar with the windows command prompt. You will need the follow prerequists installed:
 
     * Java JDK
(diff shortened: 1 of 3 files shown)


nathanhilbert:
I threw in some settings logic to check for GDAL and GEOS.  I know it is preferred to keep all logic out of settings.py, if there is a more suitable place for this check, please let me know.

nathanhilbert:
Please hold on this.  I'm cleaning it up.  I didn't mean to rebase commits that weren't mine.

nathanhilbert:
Any ideas why Travis is failing?

garnertb:
Looks like a hiccup, I restarted the build.

PR #2018, opened 2015-02-04
Edited developer mode documentation adding postgresql installtion and other edits

eomwandho:
Edited developer mode documentation adding postgresql installtion and other edits


docs/tutorials/devel/install_devmode/index.txt
```diff
@@ -8,16 +8,21 @@ In order to install Geonode 2.0 in developing mode on Ubuntu 12.04 the following
 
 For Windows: (:ref:`_install_win_devmode`)
```
capooti: maybe Ubuntu?

docs/tutorials/devel/install_devmode/index.txt
```diff
@@ -45,6 +50,33 @@ For Windows: (:ref:`_install_win_devmode`)
...
+
+     .. code-block:: console
+
+      $ sudo apt-get install postgresql-9.3-postgis-2.1 postgresql-9.3-postgis-scripts
+
+     Change postgres UNIX password
+
+     .. code-block:: console
+
+      $ sudo passwd -u postgres # change password expiry infromation
```
capooti: why do we need to set a password at all?

capooti:
nice, please check my comments.
As you changed just one doc file, it would be useful to squash the commits as described here: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

docs/tutorials/devel/install_devmode/index.txt
eomwandho: Postgresql doesn't ask users to set password during initial install on Ubuntu, so its necessary to create new password for postgres. Though one can also user thier default system user, but postgres will through an error, that database doesn't exist.

Issue #1890, opened 2014-12-15
Adding (or move) geometry editing capabilities into Layer


collaer:
Hello,

Not a bug, just a question about relocating a functionnality.

I find out a bit confusing the place where the feature geometries can be edited in geoexplorer.

Personnaly i think the best place for that capabilities should be located (also or only ?) into the "layer" menu / options rather than in the maps.

At least for first users adding some help or link in to the "Layer : Edit layer" options should be interresting because the first place you look when you want to edit a layer is in the layer menu and then in the 'edit layer' option.

My first choice should be to add an option in the edit layer menu to open a simple geoexplorer map to edit geometries.

What do you think ?

jj0hns0n:
The button labeled 'create map' on the layer explore page performs this function loading geoexplorer with this layer and allowing for editing. The button could be renamed and/or moved around on the page to make it clear this is how to 'edit this layer'.

collaer:
ok for me,
I think it should be clearer for newcomers to have both options/buttons 'edit geometries' and 'create map'.

capooti:
we could add an "Edit data" link in the menu for this purpose.

PR #2026, opened 2015-02-04
Adding link to edit data into the layer edit menu

collaer:
just an new "edit data" entry in the existing menu :
image

for #1890


geonode/layers/templates/layers/layer_detail.html
```diff
@@ -204,6 +204,9 @@ <h4 class="modal-title" id="myModalLabel">{% trans "Edit Layer" %}</h4>
                 <h4>Layer</h4>
                 {% if "change_resourcebase" in perms and not resource.service %}
                 <a class="btn btn-default btn-block btn-xs" href="{% url "layer_replace" resource.service_typename %}">{% trans "Replace" %}</a>
+                {% if resource.storeType == 'dataStore'  %}
```
capooti: needs also a check if the user has layer edit permissions

PR #2030, opened 2015-02-04
removed some required win_install_dep packages since compiled wheels are now used

nathanhilbert:
pypi is now using compiled wheels which has better support for pip than the previously used .exe files, which required easy_install.  As packages like pycsw are brought around to use wheel supported packages, win_deps_install will no longer be required per the comments in code.
Also a note in code that 64bit python users are on their own to install everything.

### DIFF
--- a/pavement.py
+++ b/pavement.py
@@ -142,8 +142,7 @@ def setup(options):
 
 
 def grab_winfiles(url, dest, packagename):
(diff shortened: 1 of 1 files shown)


nathanhilbert:
FYI [What are Wheels?](http://pythonwheels.com/)

PR #2033, opened 2015-02-04
Fixes #1307

### LINKED ISSUE #1307: Layer feature not queriable on geonode 2.0c5
In Geonode 2.0c5, a layer feature cannot be queried if the column names have spaces. Geonode 1.x knew how to handle column names with spaces in them.

It would be great if Geonode 2.x can handle column names with spaces or special characters.

capooti:
Fixes #1307 by avoiding invalid column names when uploading shapefiles.
Adds a pyshp dependency


capooti:
oops looks like the regulare expressions filter out  "_", working on this

PR #2041, opened 2015-02-04
styler does not show up when not logged #2011

### LINKED ISSUE #2011: The style editor in the layer detail page should be disabled if not logged in
If the user is not logged in, the style editor should be completely disabled since they cant actually save the styles. What happens now is an error is thrown when the app tries to save the style. It would be much better to just not allow the user to try to edit the style (as on the map detail page) or prompt them to login (as in the map composer / geoexplorer)

san_andres_y_providencia_water_ _example_com

@bartvde any ideas on this one? (sorry, you are the only geoexplorer guy we have access to)

ndufrane:
This pr resolve #2011 by not showing the button if not logged

### DIFF
--- a/geonode/templates/geonode/geo_header.html
+++ b/geonode/templates/geonode/geo_header.html
@@ -24,11 +24,13 @@ GeoExplorer.Viewer.prototype.loadConfig = function(config) {
 
     // Do not include the layer styler on the map-detail view.
     {% if not map %}
+        {% if user.is_authenticated %}
         config.viewerTools.splice(5, 0, {
             id: "styler",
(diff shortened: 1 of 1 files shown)


geonode/templates/geonode/geo_header.html
```diff
@@ -24,11 +24,13 @@
 
     // Do not include the layer styler on the map-detail view.
     {% if not map %}
+        {% if user.is_authenticated %}
```
bartvde: should this be when the user is not authenticated?
Since splice will remove the tool.

Issue #1412, opened 2014-04-09
WPS failures on attribute statistics generation


tomkralidis:
FYI when uploading a layer with columns on which statistics can be performed, GeoServer WPS is returning `ows:ExceptionReport`:
```
<?xml version="1.0" encoding="UTF-8"?>
<wps:ExecuteResponse xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xml:lang="en" service="WPS" serviceInstance="http://localhost:8080/geoserver/ows?" version="1.0.0">
<wps:Process wps:processVersion="1.0.0">
<ows:Identifier>gs:Unique</ows:Identifier>
<ows:Title>Unique</ows:Title>
<ows:Abstract>Returns the unique values of a given attribute in a feature collection.</ows:Abstract>
</wps:Process>
<wps:Status creationTime="2014-04-09T19:54:18.333Z">
<wps:ProcessFailed>
<ows:ExceptionReport version="1.1.0"><ows:Exception exceptionCode="NoApplicableCode">
<ows:ExceptionText>Process failed during execution
java.lang.NullPointerException
null</ows:ExceptionText>
</ows:Exception>
</ows:ExceptionReport>
</wps:ProcessFailed>
 </wps:Status>
</wps:ExecuteResponse>
```

jj0hns0n:
Is this still an issue?

tomkralidis:
Looks like there's a test in https://github.com/GeoNode/geonode/blob/master/geonode/layers/tests.py#L111

Basically when the WPS returns Exception, we pass and set attribute stat columns to `NA`.

Digging deeper, tracing this through shows that GeoServer WPS is throwing error:

```xml


<?xml version="1.0" encoding="UTF-8"?>
<wps:ExecuteResponse xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xml:lang="en" service="WPS" serviceInstance="http://localhost:8080/geoserver/ows?" version="1.0.0">
<wps:Process wps:processVersion="1.0.0">
    <ows:Identifier>gs:Aggregate</ows:Identifier>
    <ows:Title>Aggregate</ows:Title>
[...]

PR #2059, opened 2015-02-05
Using label and description in layer attributes when WPS_ENABLED is set to False

capooti:
In layer details page displaying the attributes label and description in place of WPS indicators when OGC WPS_ENABLED is set to False.
By default, WPS_ENABLED is now set to False. Refs #1412


jj0hns0n:
Can we show BOTH when WPS is enabled? Seems like we ALWAYS want to show the label and description and only show the stats if they are available.

capooti:
sure, I was thinking the same. Working on this right now

PR #2068, opened 2015-02-10
allowed document types list

oware:
dynamic list of allowed document types. resolves issue #1792


simod:
@oware thanks for this. The  build is failing for the indentation. Although, since the ALLOWED_DOCUMENT_TYPES is used just in one view I would pass it as context parameter in the view here https://github.com/GeoNode/geonode/blob/master/geonode/documents/views.py#L120 overriding the get_context_data method, see here https://docs.djangoproject.com/en/1.7/ref/class-based-views/mixins-simple/#contextmixin.

simod:
thanks!

Issue #2060, opened 2015-02-05
trying to fix column name when uploading a shp


collaer:
Now a check is performed and refusing shp with blank spaces into column names.
Should be better to try to fix them and replacing by underscores ?
do you agree ?

vdeparday:
+1 
I think we should sanitize them automatically and warn the user that they were sanitized and how, and give them a link to download the new one.

PR #2081, opened 2015-02-20
Geode issue 2060

teamgeode:
Commit in reference to issue #2060

Columns names from layer upload are checked before the upload and are replaced if they included error.

The rule on the regex are verified and if they are incorrect the columns name are modified to match it.

The new parameter who was transmitted is a dictionary for inform the user of the modification of the columns names.

trad
And you can see the modification on the layer page:
layer

If the upload fail, a new message are shown to user, asking him to check if the file was correct and check the charset encoding of it.

For the rename of column we use [awesome-slugify](https://pypi.python.org/pypi/awesome-slugify) who was added on the setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,7 +92,7 @@ setup(name='GeoNode',
         "Paver==1.2.1", # python-paver
         "nose <=1.0, <=1.3.1", # python-nose
         "django-nose==1.2", # python-django-nose
-        "awesome-slugify",
+        "awesome-slugify==1.6.2",
 
         # Django Apps
         "django-pagination >=1.0.5, <=1.0.7", # python-django-pagination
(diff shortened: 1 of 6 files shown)


simod:
@teamgeode this is cool. Could you verify the tests.
paver run_tests locally to run all of them.

Thanks!

setup.py
```diff
@@ -92,6 +92,7 @@ def fullsplit(path, result=None):
         "Paver==1.2.1", # python-paver
         "nose <=1.0, <=1.3.1", # python-nose
         "django-nose==1.2", # python-django-nose
+        "awesome-slugify",
```
capooti: it would be better to add a specific version here.
Well done anyway :)

PR #2087, opened 2015-02-25
Add support for zip upload

### LINKED ISSUE #1319: Zipped file upload does not work
This is with the GeoServer REST upload:

1. Upload a zip file containing a shapefile (demolayer.shp, demolayer.dbf, etc)
2. After completion, try uploading the same zipfile again, or a renamed zipfile with the same contents:

>GeoNode encountered problems when creating layer demolayer_1.It cannot find the Layer that matches this Workspace.try renaming your files.

Possible solution: create a new zipfile,  rename and copy over the contents of the old zipfile (demolayer.shp -> demolayer_1.shp, etc), and send the new zipfile to Geoserver for importing.  I can create a PR for this quickly if it's an acceptable fix.

simod:
One shapefile per zip, fixes #1319

### DIFF
--- a/geonode/layers/forms.py
+++ b/geonode/layers/forms.py
@@ -21,6 +21,8 @@
 import os
 import tempfile
 import autocomplete_light
+import zipfile
+
 
 from django import forms
 from django.utils import simplejson as json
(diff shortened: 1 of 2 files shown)


vesnikos:
Hi, 
since tempfile.mkdtemp() doesn't clean up itself (it's the users resposibility to clean it up) isn't possible for geonode to bloat the tmp folder with unwanted data? 

A suggestion to help workarounds on the 'issue' would be to add a prefix on the generated tmp folder.

simod:
We have this line https://github.com/GeoNode/geonode/blob/master/geonode/layers/views.py#L199 that should clean up. Seems to be working for me. Will doublecheck

state-hiu:
Great!  It is not needed in this pull request, but we should add `.tar.gz` support, too.  The implementation would be similar, but would require some refractoring to maintain DRY.  See:

https://docs.python.org/2/library/tarfile.html

Issue #2022, opened 2015-02-04
Layer search is no longer available in the GXP map viewer when adding a layer


vdeparday:
When adding a layer with the green + sign in the map viewer, there is the list of layers (can be several hundred) and there is no way to filter/search it:
![image](https://cloud.githubusercontent.com/assets/1036189/6042018/7b62df92-ac4f-11e4-9dd7-7e7937cb478a.png)

It seems like the search is still in the code but not enabled or not showing up
https://github.com/GeoNode/geonode/blob/master/geonode/static/geonode/js/extjs/GeoNode-GeoExplorer.js

@bartvde an idea?

bartvde:
looking at app.sources it doesn't seem like a search source is defined in the config?

bartvde:
ah it actually there and named search will look a bit further

bartvde:
@vdeparday can you try changing search: true to:

```
search: {
  selectedSource: 'search'
}
```

bartvde:
So in this spot: https://github.com/GeoNode/geonode/blob/master/geonode/static/geonode/js/extjs/GeoNode-GeoExplorer.js#L439

jj0hns0n:
@bartvde that did the trick and enabled the widget again. The search API has changed dramatically since this last worked. Can you point us at where we set the Search API URL and what is the expected output format. We will need to figure out how to adapt to the new API.

bartvde:
sure url is set here: https://github.com/GeoNode/geonode/blob/master/geonode/static/geonode/js/extjs/GeoNode-GeoExplorer.js#L451

bartvde:
The protocol code is here: https://github.com/boundlessgeo/gxp/blob/master/src/script/plugins/GeoNodeCatalogueSource.js

vdeparday:
I am trying to follow: http://docs.geonode.org/en/master/tutorials/devel/debug/geoexplorer.html to change and debug the GeoNodeCatalogueSource.js to adapt to the new Search api but https://github.com/GeoNode/suite/tree/geonode/geoexplorer no longer compiles because it seems to refer to http://repo.opengeo.org/

@ingenieroariel , how can I get past that?

''''
Downloading: http://dev.mapfish.org/maven/repository/org/ringojs/ringojs/0.8.1/ringojs-0.8.1.jar
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 37.721 s
[...]

PR #2107, opened 2015-03-31
Re-enable layer searcher in the add layer tool 

ndufrane:
Proposition to re-enable the layer searcher in the add layer tool of the map composer (issue #2022 ).

This code use the geonode api (/api/layers) and a copy of the GeonodeCatalogSource adapted to match the geonode api

### DIFF
--- a/geonode/maps/templates/maps/map_geoexplorer.js
+++ b/geonode/maps/templates/maps/map_geoexplorer.js
@@ -2,6 +2,7 @@
 {% include 'geonode/geo_header.html' %}
 <link href="{{ STATIC_URL}}geonode/css/geoexplorer/map_geoexplorer.css" rel="stylesheet"/>
 <script type="text/javascript" src="{{ STATIC_URL}}geonode/js/extjs/GeoNode-mixin.js"></script>
+<script type="text/javascript" src="{{ STATIC_URL}}geonode/js/extjs/Geonode-CatalogueApiSearch.js"></script>
 <script type="text/javascript" src="{{ STATIC_URL}}geonode/js/extjs/GeoNode-GeoExplorer.js"></script>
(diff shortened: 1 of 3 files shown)


capooti:
thanks for this PR, it would be very useful.
The correct workflow to add a geoexplorer module such as Geonode-CatalogueApiSearch.js would be to rebuild GeoNode-Geoexplore-js
For example, add it as done here: https://github.com/GeoNode/suite/commit/da46d9479fecd46933e93f0023b6e60ea54d1c5d
and then rebuild geoexplorer as explained here:
https://github.com/GeoNode/suite
Let me know if you need any help, and thanks so much for this really useful effort! ;)

ndufrane:
Oops i did not saw the Geonode/suite fork ...

Just one question : is geoexplorer/app/static/script/ux the correct directory to put the Geonode-CatalogueApiSearch.js ? I mean, there is a spécific place where i can put geonode spécifics javascript ?

capooti:
Sorry, I misread your questions, I thought you want to add a generic tool, but this is specific to GeoNode, then the way you do is correct to me.
@bartvde could you confirm this? thanks

bartvde:
ux is used for Ext JS user extensions only IIRC

ndufrane:
I think too, that the javascript inside the suite repo needs to be related to generic tools.

I admit that placing the js inside the extjs folder of geonode is not the best way. Because the javascript is not minified and not packaged in one big javascript file (The same problem applies for GeoNode-GeoExplorer.js and GeoNode-mixin.js) causing more http request to be issued when loading the page.

So, this is ok like this or this PR need some modification to put the Javascript in a better place ?

BULIAD:
Thank You V.Much!!

PR #2116, opened 2015-04-09
Additional options for importlayers command

mbertrand:
Added command line options for specifying category, title, and 'private' (making layer viewable only to owner).


capooti:
nice! Could you add some documentation here [1]?
thanks so much

[1] http://geonode.readthedocs.org/en/latest/tutorials/admin/commands.html

mbertrand:
Updated the documentation and fixed some test/flake8 issues too.

PR #2120, opened 2015-04-13
Fix error with keyword filtering when settings.SKIP_PERMS_FILTER=True

mbertrand:


capooti:
hi Matt, published was based on a custom manager:
https://github.com/GeoNode/geonode/pull/1763/files#diff-c29f4555cb81f97adce989f4907b15c0R626

now I think for some reasons it has been removed, therefore your pr is fine with me.
Unless we want to restore the PublishedResourceBaseManager, not sure when it was removed and by who.

simod:
+1 on merging

capooti:
oops, it was me LOL
https://github.com/GeoNode/geonode/commit/0a419a6ecd1d847e2c39e4d1b30e1febf4fe584b#diff-c29f4555cb81f97adce989f4907b15c0

Thanks for fixing this @mbertrand

PR #2128, opened 2015-04-17
Creates a full geonode db for permormance testing purposes

capooti:
This script helps to test GeoNode performances (including GeoServer side) by generating a number of layers and documents on a GeoNode instance.


capooti:
sorry, made same mistake with squashing commits.
I will create another PR.

Issue #2130, opened 2015-04-17
Performance issues when uploading layers to GeoNode


capooti:
Using this script [1], it looks like GeoServer is initially very fast but then the upload time quickly increases with the layers number (time is in seconds, the uploaded shapefile is ~5k):

Layer 0 generated in: 1.83858203888
Layer 1 generated in: 1.85110998154
Layer 2 generated in: 2.00699400902
...
Layer 48 generated in: 4.82827591896
Layer 49 generated in: 4.83818912506
Layer 50 generated in: 5.1511900425
...
Layer 71 generated in: 8.78467202187
Layer 72 generated in: 9.61577486992
Layer 73 generated in: 10.5833621025
...
Layer 293 generated in: 25.4465839863
Layer 294 generated in: 27.8491809368
Layer 295 generated in: 36.5090360641
...

[1] https://github.com/capooti/geonode/blob/script-to-test-performance/scripts/misc/create_full_geonode_db.py

pjdufour:
@simod, close issue?

Issue #2122, opened 2015-04-14
Performance issues with geonode.security.views.resource_permissions


capooti:
I am testing GeoNode 2.4 with a large number of resource base objects (about 2000).
When accessing to the detail page of one resource, for example a layer, the view in the subject [1]  is extremely slow.
After some debugging it looks like this line [2] is taking a long time to execute (about 1 minute).

Same behavior from python shell:

get_users_with_perms(layer, attach_perms=True, with_superusers=True)
time: about 45 sec

It looks like django-guardian try to run this query, that keeps postgres working at 100% for all that time:

```sql
SELECT DISTINCT "people_profile"."id", "people_profile"."password", "people_profile"."last_login", "people_profile"."is_superuser", "people_profile"."username", "people_profile"."first_name", "people_profile"."last_name", "people_profile"."email", "people_profile"."is_staff", "people_profile"."is_active", "people_profile"."date_joined", "people_profile"."organization", "people_profile"."profile", "people_profile"."position", "people_profile"."voice", "people_profile"."fax", "people_profile"."delivery", "people_profile"."city", "people_profile"."area", "people_profile"."zipcode", "people_profile"."country" FROM "people_profile" LEFT OUTER JOIN "guardian_userobjectpermission" ON ( "people_profile"."id" = "guardian_userobjectpermission"."user_id" ) LEFT OUTER JOIN "people_profile_groups" ON ( "people_profile"."id" = "people_profile_groups"."profile_id" ) LEFT OUTER JOIN "auth_group" ON ( "people_profile_groups"."group_id" = "auth_group"."id" ) LEFT OUTER JOIN "guardian_groupobjectpermission" ON ( "auth_group"."id" = "guardian_groupobjectpermission"."group_id" ) WHERE (("guardian_userobjectpermission"."object_pk" = '4453'  AND "guardian_userobjectpermission"."content_type_id" = 38 ) OR ("guardian_groupobjectpermission"."object_pk" = '4453'  AND "guardian_groupobjectpermission"."content_type_id" = 38 ))
```
[1] https://github.com/GeoNode/geonode/blob/master/geonode/security/views.py#L51
[2] https://github.com/GeoNode/geonode/blob/master/geonode/security/models.py#L56

simod:
I can confirm that it slows down but not so much, 3.6 seconds with 1400 + resources. Better to investigate though.

capooti:
[...]

PR #2133, opened 2015-04-21
Improve geonode performances

simod:
This PR solves the major issues described in #2130 and #2122.
The guardian permissions are checked faster avoiding complex joins
The categories, regions and keyword counts are done in the post-process in bulk and not per object anymore
The layer_upload uses gsconfig so to not query geoserver for all the layers available
The layer_acls check is improved by ~75%

A proper test to load lot of data has to be done in a realistic way. Loading many times the same layer will hit the db to find an available name so there is still a slight slow down in layer upload like it becomes 1s slower every 300 layer loaded.

### DIFF
--- a/geonode/api/api.py
+++ b/geonode/api/api.py
@@ -233,17 +233,17 @@ class ProfileResource(ModelResource):
     def dehydrate_layers_count(self, bundle):
         obj_with_perms = get_objects_for_user(bundle.request.user,
                                               'base.view_resourcebase').instance_of(Layer)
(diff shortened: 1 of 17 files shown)


ingenieroariel:
:+1:

capooti:
it looks perfect to me, great work! ;)

ischneider:
Nice work, @simod.

As per my line comment, it looks like there are several other places where the use of `values_list` can be replaced.

simod:
thanks! fixing them

PR #2136, opened 2015-04-22
improve visual of document_detail page

teamgeode:
A little improvement for visualization of the document preview.

Verified on smartphone and web browser.
The main problem was that the preview was taking too much space.

50% look right for me.


ischneider:
I don't have time at this exact moment to test, but does this work?
http://bootstrapdocs.com/v3.0.0/docs/css/#overview-responsive-images

teamgeode:
@ischneider I tested the responsive tag and that’s work fine.
So now we have a width on 500px for normal screen (better look than 100%) and a 100% on small screen.

PR #2165, opened 2015-05-20
port existing MapStory Favorites feature to GeoNode as optional module.

jwood106:
Uncomment 'geonode.contrib.favorite' entry in settings.py to enable.
Once enabled, UI allows authenticated user to favorite Document, Layer and Map.

### DIFF
--- a/geonode/contrib/favorite/models.py
+++ b/geonode/contrib/favorite/models.py
@@ -1,0 +1,81 @@
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.contenttypes import generic
+from django.db import models
+
+from geonode.documents.models import Document
+from geonode.layers.models import Layer
+from geonode.maps.models import Map
+
+
+class FavoriteManager(models.Manager):
+    def favorites_for_user(self, user):
+        return self.filter(user=user)
+
+    def _favorite_ct_for_user(self, user, model):
+        content_type = ContentType.objects.get_for_model(model)
+        return self.favorites_for_user(user).filter(content_type=content_type).prefetch_related('content_object')
+
+    def favorite_documents_for_user(self, user):
+        return self._favorite_ct_for_user(user, Document)
+
+    def favorite_maps_for_user(self, user):
+        return self._favorite_ct_for_user(user, Map)
+
+    def favorite_layers_for_user(self, user):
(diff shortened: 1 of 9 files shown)


garnertb:
Please squash this into a single commit so that we can review it.

simod:
What is the Favorite functionality?
I know that we have some if statements spread around like the one of NOTIFICATION, but we are trying to get rid of them (and we will in the near future). It would be preferable to have non invasive contrib apps so that they can be plugged in without the need add code in core. On the long (and even short) term it will become very difficult to maintain that code.

jwood106:
Per garnertb's request, squashed into single commit.
Per simod's question, while squashing, added better favorite functionality description in the commit description.
Re: simod's comment about minimizing changes necessary in core for an optional contrib app, I agree.  Definitely valuable to make the minimum changes to the core, keep as much isolated in the contrib module for clarity and maintainability.  I made the minimum 'hook' changes to core that I know to do.  You mentioned you are looking at how to get rid of existing hooks, makes sense to apply the same here, if applicable.

ingenieroariel:
I understand @simod's suggestion and also believe the notification one is a particularly intrusive third party app and requires conditionals all over the place.

That said, I see why favorites is someting that should be added to document detail, layer detail and map detail pages. In that case, I believe it is better to instruct the user to add the relevant HTML snippet manually instead of writing the conditional code. This is more in line to how Django apps work (they ask you to add things to templates, settings.py and urls.py) and end up in more flexibility for the end user.

@simod, what do you think about the compromise of having code that is commented out and can be uncommented to enable the functionality instead of the context variables that were introduced?

simod:
I think is better, especially if the template is just a snippet that lives in the contrib app, with the instructions, that can be {% included %} in the desired pages.

jwood106:
Per GeoNode developer feedback above (btw, thank you for the feedback), I removed all previously submitted changes to GeoNode core, now this pull request contains only code in the new contrib module.
I added a readme.txt with instructions for changes developer will need to make to GeoNode core if want to enable the Favorite module in the app.
Squashed the changes into a single commit.
Thank you to Tyler Garner for superior Git support.  He helped me overcome several issues I had and improve my workflow for this pull request.  Much appreciated!

ingenieroariel:
I am :+1: on merging

simod:
+1 too

PR #2172, opened 2015-05-29
fix encoding problem when using Spanish lang (accents)

cristianzamar:
When you set GeoNetwork as the CSW backend and Spanish language, it crashes with error:

`UnicodeDecodeError: 'ascii' codec can't decode byte 0xf3 in position ##: ordinal not in range(128)`

I found that the problem were the accents included in the metadata content.

### DIFF
--- a/geonode/catalogue/backends/generic.py
+++ b/geonode/catalogue/backends/generic.py
@@ -154,7 +154,7 @@ class Catalogue(CatalogueServiceWeb):
 
     def csw_request(self, layer, template):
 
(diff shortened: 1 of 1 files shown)


ingenieroariel:
Do you know if this affects master too?

cristianzamar:
Not yet, but I will test it soon and make another pull request if it applies to v2.4. Otherwise, I'll comment the failure here.

cristianzamar:
I confirm it applies to master too, thus created PR #2185

PR #2192, opened 2015-06-17
Added SRID Flag

pjdufour:
Adds `SRID_DETAIL` flag for specifying if you want to show the SRID on a layer/map detail page above/below the fold.  Current default of not showing is maintained.


capooti:
It would be nice to add some documentation here.

PR #2198, opened 2015-06-24
added filter by owner and SEARCH_FILTERS configuration

menegon:


geonode/api/api.py
```diff
@@ -156,6 +157,24 @@ class Meta:
         serializer = CountJSONSerializer()
 
 
+class OwnerResource(TypeFilteredResource):
```
simod: @menegon would it be possible to reuse https://github.com/GeoNode/geonode/blob/master/geonode/api/api.py#L185?

menegon: I have just sent a pull request in order to reuse the ProfileResource class. 
But I'm not sure if it is better because Profile Resource makes more database queries (by dehydrate functions)

PR #2205, opened 2015-06-28
Exif Support

pjdufour:
This PR adds an Exif contrib app.  This app is disabled by default.  If enabled by the `EXIF_ENABLED` settings flag, this app parses Exif tags from JPEGs uploaded as documents and automatically generates metadata, including date, keywords, and a brief abstract.  It uses `from PIL import Image, ExifTags`, so doesn't require any changes to dependencies.


simod:
@pjdufour there may be a conflict with the previous PRs here, could you fix it?

pjdufour:
Sure thing.

pjdufour:
Hold please.

pjdufour:
Also, I just added in tab for document detail pages that shows a few Exif tag values.

pjdufour:
Let me squash.

PR #2210, opened 2015-07-03
Fix #2209

### LINKED ISSUE #2209: Context/template variable for permissions ("perms")
Unless it's by design for some reason, we should not overwrite the variable `perms` in templates to make it represent the list of perms from django-guardian.  For example, 

```Python
[u'add_resourcebase', u'change_resourcebase', u'change_resourcebase_metadata', u'change_resourcebase_permissions', u'delete_resourcebase', u'download_resourcebase', u'publish_resourcebase', u'view_resourcebase']
```

The variable name `perms` is already used by django.contrib.auth and [represents a `PermWrapper` object](https://docs.djangoproject.com/en/1.7/ref/templates/api/#django-contrib-auth-context-processors-auth).

```
perms – An instance of django.contrib.auth.context_processors.PermWrapper, representing the permissions that the currently logged-in user has.
```
I'd suggest changing the name globally (for all templates) to `perms_list` (or similar name) and passing in as a context variable using the django-guardian function [get_perms](http://django-guardian.readthedocs.org/en/latest/userguide/check.html?highlight=get_perms#get-perms) rather than calling `{% get_obj_perms request.user for resource.get_self_resource as "perms" %}` directly in the template.

This has no known security vulnerabilities (hence not critical).  However, de-conflicting these names would make it easier to debug and add features to the permissions interface.

pjdufour:
This PR is a fix for #2209 and other minor changes. Discussion on issue page.

### DIFF
--- a/geonode/documents/templates/documents/document_detail.html
+++ b/geonode/documents/templates/documents/document_detail.html
@@ -31,9 +31,7 @@
 
 <div class="row">
   <div class="col-md-8">
-
-    {% get_obj_perms request.user for resource.get_self_resource as "perms" %}
-    {% if "download_resourcebase" in perms %}
+    {% if "download_resourcebase" in perms_list %}
       {% if resource.extension|lower in imgtypes and resource.doc_file %}
       <div id="embedded_map">
         <a style="text-decoration:none;" href="{% url "document_download" resource.id %}" target="_blank">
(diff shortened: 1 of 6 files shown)


capooti:
some tests is failing, could you check? thanks

pjdufour:
Reverted some code and rebased.  Tests are passing now.

PR #2230, opened 2015-07-16
Intitial NLP Support

pjdufour:
This PR add a `nlp` contrib app.  This app is disabled by default.  If enabled by the `NLP_ENABLED` settings flag, the app will use [MITIE](https://github.com/mit-nlp/MITIE) and natural language processing (NLP) techniques to infer additional metadata from the raw contents of new text documents.  This contrib app will help fill key metadata gaps.

Specifically, the app will identify locations mentioned within the document and crosscheck them against regions (https://github.com/GeoNode/geonode/blob/master/geonode/base/models.py#L117) to autofill the regions metadata field.  Keywords are generated for identified organizations but only if they match "existing" keywords.

In the future this app can be enhanced in a few ways.  Additional structured metadata could be inferred for layers by reading in the abstract, title, and purpose from a metadata file.  The remote services feature could use the nlp app to fill in structured metadata for remote services that have none or little to begin with.  Lastly, a separate library agnostic to GeoNode's models should be built outside of the core so [CKAN](https://github.com/CKAN/ckan) and other platforms can reuse some of the opensource code.

### DIFF
--- a/docs/reference/developers/settings.txt
+++ b/docs/reference/developers/settings.txt
@@ -558,6 +558,30 @@ Default: ``False``
 
 A boolean that specifies whether the Exif contrib app is enabled.  If enabled, metadata is generated from Exif tags when documents are uploaded.
 
+NLP_ENABED
+...........
+Default: ``False``
+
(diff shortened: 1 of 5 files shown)


simod:
Nice! The failure was due to geonode and not the PR. (flake8). I'm +1

jj0hns0n:
@pjdufour can you bump this so travis builds again.

garnertb:
I manually restarted the build.

PR #2233, opened 2015-07-16
add OpenSearch autodiscovery support

tomkralidis:
cc @pjdufour 

This PR provides [OpenSearch](http://www.opensearch.org/Home) support to GeoNode in the following areas:

- all webpages have an [autodiscovery](http://www.opensearch.org/Specifications/OpenSearch/1.1#Autodiscovery_in_HTML.2FXHTML) link in `<head>`.
- the discovery document has a link to the basic HTML search as well as a pycsw powered Atom output search.  The reason for two different endpoints in the document is because pycsw does not provide HTML output, which is an OpenSearch autodiscovery requirement for web browser based search addons

### DIFF
--- a/geonode/catalogue/templates/catalogue/opensearch_description.xml
+++ b/geonode/catalogue/templates/catalogue/opensearch_description.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+    <ShortName>{{ shortname }}</ShortName>
+    <LongName>{{ shortname }}</LongName>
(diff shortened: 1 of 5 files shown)


pjdufour:
Cool!  Any other search engine optimization to do?

tomkralidis:
@pjdufour I think we're good for now for SEO.

Aside: I'm getting an unrelated travis error at https://travis-ci.org/GeoNode/geonode/builds/71342467#L1796 which seems to be related to the recent EXIF enhancements.

pjdufour:
That error had to do with loading templates for contrib apps.  Fixed in https://github.com/GeoNode/geonode/commit/58b5ff280faae66bd91bbc21b82ccd45183fe650.  You should just need to restart the Travis-CI job on your branch.

PR #2239, opened 2015-07-18
Fix #2166

### LINKED ISSUE #2166: Email addresses not added for new users added through admin
When adding a new user through the admin (Users model), the email address is not added to the Email Addresses table.     Consequently, if you forget your password it cannot find your email address.

pjdufour:
Fix #2166.  Create emailadress when creating new user through admin and keep in sync.


pjdufour:
@simod, can you put eyes on this and make sure this is the preferred solution.  I don't have prior experience dealing with the django user accounts package.

capooti:
+1 for me

Issue #2285, opened 2015-09-28
document upload fails on very long filenames


senoadiw:
Hi all,

I tried uploading a document file with a very long name and got the following error log:

DataError: value too long for type character varying(100)

I think doc_file field should have max_length set to 255 like the file field in LayerFile => https://github.com/GeoNode/geonode/blob/master/geonode/layers/models.py#L268

Regards,

SN

capooti:
I agree with you and I think there is another issue with this, but cannot find it.
I think this case of schema change is allowed, should we change this before releasing 2.4?

jj0hns0n:
@capooti is this implemented? Moving to 2.7 in any case.

PR #2288, opened 2015-09-29
Set max_length=255 for doc_file and doc_url in document model

capooti:
Set max_length=255 for doc_file and doc_url in document model as per #2285

### DIFF
--- a/geonode/documents/models.py
+++ b/geonode/documents/models.py
@@ -37,6 +37,7 @@ class Document(ResourceBase):
     doc_file = models.FileField(upload_to='documents',
                                 null=True,
                                 blank=True,
+                                max_length=255,
                                 verbose_name=_('File'))
 
     extension = models.CharField(max_length=128, blank=True, null=True)
@@ -46,6 +47,7 @@ class Document(ResourceBase):
     doc_url = models.URLField(
         blank=True,
         null=True,
+        max_length=255,
         help_text=_('The URL of the document if it is external.'),
         verbose_name=_('URL'))
 


simod:
@capooti wouldn't this break existing databases?

capooti:
I don't think so, as it is just setting the field length to an higher value for new synced database.
Though best practice here it would be to have a migration script, so we may postpone this to 2.5.

simod:
I'm afraid that the db would complain if we allow more that 100 but the it has a constraint of 100. Unless we will provide a migrations script.

capooti:
apparently this is already happening according to #2285. We need to make
sure that the form does not accept file names longer than 100 chars then.

On Tue, Sep 29, 2015 at 9:39 AM, Simone Dalmasso <notifications@github.com>
wrote:

> I'm afraid that the db would complain if we allow more that 100 but the it
> has a constraint of 100. Unless we will provide a migrations script.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GeoNode/geonode/pull/2288#issuecomment-143974342>.
>



-- 
Paolo Corti
Geospatial software developer
web: http://www.paolocorti.net
twitter: @capooti
skype: capooti

jj0hns0n:
@capooti should we go ahead and merge this one now?

capooti:
Yes, given that now we can do schema changes.
Should we start using Django migrations for schema changes, by the way? (this way will be easier to update production systems)

simod:
@capooti +1 for the migrations, not sure that django 1.6 has migrations yet but we should also upgrade it to the latest.

capooti:
Yes, now I think south is integrated in django.
let's do this: when we will updated django to the latest (time to do it, imho, let me know if I may help), I will add the migration to this PR, and then we merge it

jj0hns0n:
@capooti Can we go ahead and merge this now?

capooti:
Sure! Go ahead

PR #2289, opened 2015-09-29
added additional filetypes for geonode.importer - gdal specific

BerryDaniel:
Requires the GeoServer GDAL extension to be added to GeoServer.

GDAL filetypes added are CADRG (Compressed ARC Digitized Raster Graphics) / CIB Raster Product Format, NITF (National Imagery Transmission Format), ERDAS Imagine, MrSID (Multiresolution Seamless Image Database), JPEG2000.

The .sid and .jp2 files require gdal to be built with MrSID and JP2 support.


pjdufour:
I like this idea alot but think this is a good opportunity to refractor `FileTypes.js` a bit.  Since https://github.com/BerryDaniel/geonode/blob/gdal_import/geonode/static/geonode/js/upload/FileTypes.js#L5, is just returning a JSON object, could we build the object using some for loops rather than hardcoding each option such as CIB_127 and CIB_128 where the only difference is the number.

```Javascript
    var output = {
        SHP: new FileType({
            name: gettext('ESRI Shapefile'),
            format: 'vector',
            main: 'shp',
            requires: ['shp', 'prj', 'dbf', 'shx']
        }),
        TIF: new FileType({
            name: gettext('GeoTiff File'),
            format: 'raster',
            main: 'tif',
            requires: ['tif']
        }),
    ...
    ...
    ...
    };
for (var i = 41; i < 50; i++)
{
  output['CIB'+i] = new FileType({
            name: gettext('Controlled Image Base - 1 meter'),
            format: 'raster',
            main: 'i'+i,
            requires: ['i'+i]
        }),
}
```

BerryDaniel:
@pjdufour Took your advice and refactored FileTypes.js a bit.

capooti:
Could you make it configurable?
For example with a settings named something like GEOSERVER_GDAL_ENABLED, set by default to False?

garnertb:
@capooti, any reason why we shouldn't include the gdal extension in the geoserver build and default this to true?

capooti:
if it is stable, I have nothing against :)
Just we need to make sure to handle scenarios where GeoServer does not come with such an extension imho

pjdufour:
Not to push refractoring too much, since other parts of GeoNode could use a good scrub, too.  However at https://github.com/GeoNode/geonode/blob/master/geonode/static/geonode/js/upload/FileType.js#L19 could we "default" this.requires to [this.main] like (pseudo-code):

```
this.requires = this.requires == null ? [this.main] : this.requires
```

Therefore we could cut down on the

```
main: 'foo',
requires: ['foo']
```

repetition and only keep the requires line for shapefiles mostly.

BerryDaniel:
@pjdufour made the adjustment

garnertb:
https://github.com/GeoNode/geonode/pull/2340.

PR #2293, opened 2015-10-01
fixed issue with date before 1900

menegon:
This fix an issue of strftime when the year of the resource data is before 1900.


pjdufour:
Could you actually add a comment to the file right before the fix saying why we aren't using strftime, so we don't delete this later by accident?

PR #2298, opened 2015-10-06
remove obsolete references to category form in labels/views.py

ccosse:
remove obsolete references to category form in labels/views.py


simod:
Thanks!

Issue #2304, opened 2015-10-13
Include detailed layer information in maps API endpoint


cstephen:
Currently, the maps API endpoint at `/api/maps/<map-number>` includes a count of the layers a map has, but no information about the layers themselves. My understanding is this is because the maps API is currently only used to return search results.

At the Scenarios Network for Alaska + Arctic Planning, we are developing an isolated web application that needs to use the API to load WMS information for all of the layers belonging to a particular map. We went ahead and added a new API endpoint for detailed map layer information and gave it a new path, `/api/maplayers/<map-number>`. This new endpoint includes everything returned by the original "maps" endpoint but also tacks on all of the information returned by the Map model's `layer_config()` method. This includes each layer's title, GeoServer WMS URL, bounding box, keywords, etc.

I'd like to contribute this work to the GeoNode project, but does this approach make sense? Or would it be better to add map layer information to the existing `/api/maps/<map-number>` endpoint?

simod:
@cstephen please go ahead with a pull request. The apis for search are very light by design for performance purpose and retrieve only the information available in the respective tables without following relations. The tastypie dehydration cycle has been disabled. But the apis are also designed to be extended and we did already with a Owners light api and a Profile heavier api. So we can absolutely include a new more verbose one for maps and layers.

capooti:
Alternatively you could extend the API to your needs in your GeoNode custom
project, without forking GeoNode.

Here is a quick code sample that extend the Layer API with a new attribute
(not tested but should put you on the way):

class ExtendedLayerResource(LayerResource):

    extent = fields.TextField()

    def dehydrate_extent(self, bundle):
        return '%s %s %s %s' % (bundle.obj.bbox_x0, bundle.obj.bbox_x1,
bundle.obj.bbox_y0, bundle.obj.bbox_y1)

then from template call the ExtendedLayerResource end point in place of the
LayerResource
hope it helps
p

On Tue, Oct 13, 2015 at 8:27 AM, Simone Dalmasso <notifications@github.com>
wrote:

> @cstephen <https://github.com/cstephen> please go ahead with a pull
> request. The apis for search are very light by design for performance
> purpose and retrieve only the information available in the respective
> tables without following relations. The tastypie dehydration cycle has been
> disabled. But the apis are also designed to be extended and we did already
> with a Owners light api and a Profile heavier api. So we can absolutely
> include a new more verbose one for maps and layers.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #2305, opened 2015-10-13
Added a new MapLayers API endpoint that adds layer information to bas…

cstephen:
…ic Maps API.

This is in reference to issue #2304.

### DIFF
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -539,8 +539,11 @@ class MapLayersResource(CommonModelApi):
     def dehydrate(self, bundle):
         map_id = bundle.data['id']
         layers = []
-        for maplayer in MapLayer.objects.filter(map_id=map_id):
-            layers.append(maplayer.layer_config())
+        for maplayerObj in MapLayer.objects.filter(map_id=map_id):
+            layerObj = Layer.objects.filter(typename=maplayerObj.name)[0]
+            layer = maplayerObj.layer_config()
+            layer['id'] = layerObj.id
+            layers.append(layer)
         bundle.data['layers'] = layers
         return bundle
 
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -18,6 +18,7 @@ from tastypie.utils.mime import build_content_type
 
 from geonode.layers.models import Layer
 from geonode.maps.models import Map
+from geonode.maps.models import MapLayer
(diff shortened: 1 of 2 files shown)


geonode/api/resourcebase_api.py
```diff
@@ -525,6 +526,25 @@ class Meta(CommonMetaApi):
...
+
+    """Includes everything from Maps API plus details for each layer in map"""
+
+    class Meta(CommonMetaApi):
+        queryset = Map.objects.distinct().order_by('-date')
+        if settings.RESOURCE_PUBLISHING:
+            queryset = queryset.filter(is_published=True)
+        resource_name = 'maplayers'
+
+    def dehydrate(self, bundle):
```
simod: @cstephen did you do any performance test on this api? Dehydrate loops over each result and for each result you are looping over each maplayer. The performances may decrease quickly with the grow of the number of maps.

cstephen: Good thinking, @simod.

Here are some performance test results for this new API endpoint. These tests were run against a virtual machine with 4 CPU cores and 8gb of RAM. I can run the same tests on real hardware if needed but it will take more time.

I created some example maps with 10, 20, 50, and 100 layers. For each of these, I accessed the new endpoint at `/api/maplayers/<map-id>` and:

- Took the average of ten timed `wget` requests while GeoNode was sitting idle.
- Sent requests to the endpoint in rapid succession with jMeter and took the average after 500 requests (plenty of time for the average to stabilize).

Here are the results:

*10 layers*
Idle: 0.10 seconds
jMeter: 0.51 seconds

*20 layers*
Idle: 0.18 seconds
jMeter: 0.69 seconds

*50 layers*
Idle: 0.26 seconds
jMeter: 1.29 seconds

*100 layers*
Idle: 0.46 seconds
jMeter: 2.29 seconds

If you access the new endpoint without specifying a map ID (just `/api/maplayers/`) it loads the same thing that `/api/maps/` does, so there is no performance impact there.

cstephen:
The numeric ID for each layer has been included in the `maplayers` endpoint so it could be used to fetch more detailed layer information from the existing `layers` endpoint.

simod:
thanks @cstephen we will discuss this PR during the code sprint.

simod:
@cstephen this API is not used by GeoNode itself at the moment because it is specific for your project. We will keep this PR here for the 2.5 but for your case it would be better if you can use it in your downstream project.

PR #2308, opened 2015-10-14
Changes to the spherical projections and better error handling and lo…

jmwenda:
…gging


simod:
@jmwenda there are some tests failures, could you take a look?

Issue #2314, opened 2015-10-19
Adding apiKey in local_settings.py for Google and Bing maps


AnnalisaS:
It's possible to add to MAP_BASELAYERS entry, in local_settings.py, also Bing and Google basemaps setting my own key, if available?
Thank you

capooti:
I am working on this. The admin can set the BING api key in settings, and in such a case the bing basemap is enabled.

jj0hns0n:
The BING key is not respected by geoexplorer, but this has been implemented.

PR #2341, opened 2015-11-17
Now bingmaps base layer will be enabled if a BING_API_KEY is defined

capooti:
Refs #2314


capooti:
should make easy for the end user to enable bingmaps, but if you don't like it we do not merge it :)

garnertb:
I like the functionality, its just preferable to implement in a way that users can still override the other map configuration settings.

Issue #1022, opened 2013-08-07
Add json endpoints for open data policy compliance.


jj0hns0n:
One thing that might be good to add to the feature wish list (at least for federal customers) is to have GeoNode satisfy the new 'open data' policy for its own content.  That would mean providing their JSON output format at geonode.example.com/data.  

http://project-open-data.github.io/schema/
http://project-open-data.github.io/examples/catalog-sample-extended.json

pjdufour:
See the following page to learn how to get geospatial data into Data.Gov and GeoPlatform.

http://www.digitalgov.gov/resources/how-to-get-your-open-data-on-data-gov/#federal-geospatial-data

The CSW endpoint is still accepted (and I would say preferred to data.json).  For many agencies the data.json endpoint is controlled by a non-geospatial office and is managed in a legacy system (SharePoint, or manually written, etc.), which creates a significant bottleneck.  A `/data.json` endpoint on GeoNode would not be used.

To help agencies fulfill their OMB **requirements**, fixing metadata and CSW bugs associated with CKAN integration (https://github.com/GeoNode/geonode/labels/ckan-integration) will enable direct harvest from GeoNode into data.gov and agency level CKANs.

tomkralidis:
@kalxas has implemented the data.gov CSW and could perhaps speak to CKAN harvesting.  It's worth  articulating the gaps in GeoNode's metadata model on CSW output so we can apply and achieve harvesting/federation.

pjdufour:
Yes, hopefully we can do a more thorough review of the gaps at some point (Comparing ISO 19115 vs. actual GeoNode implementation vs. OMB/data.json requirements).

CC @JJediny

kalxas:
Hi,
Harvesting from CSW is the best option in my opinion (large US federal organizations are deploying pycsw to cover this requirement), so we need to fix all CSW related issues. 
Of course It would be great to have OpenDataJSON support. 
For data.gov we implemented a harvester from OpenDataJSON to ISO 19115 in order to get all those records in pycsw.
How about an OpenDataJSON output format plugin for pycsw to do the reverse?

mwengren:
All,

FWIW, I did some testing before FOSS4G last year on GeoNode CSW -> CKAN harvesting via CS-W.  The shortcomings that I noticed on the CKAN side were:

1) would not present a WMS 'Preview Map' link without a full GetCapabilties URL present within GeoNode CS-W/ISO 199115 OnlineResource links
[...]

PR #2342, opened 2015-11-17
add support for data.json output (#1022)

tomkralidis:
Per #1022.


JJediny:
hey now... so @tomkralidis whats the relative path to the endpoint?

tomkralidis:
@JJediny `/data.json`.  You can see an example at http://demo.geonode.org/data.json which is linked off http://demo.geonode.org/developer/

Issue #2351, opened 2015-11-18
xyz.shp.xml is not supported as an extension for metadata


jj0hns0n:
This is very common and apparently not supported. @tomkralidis can you look into this?

![screenshot 2015-11-18 15 03 47](https://cloud.githubusercontent.com/assets/506078/11254476/86fd7846-8e06-11e5-893f-a70784b9b987.png)

PR #2352, opened 2015-11-18
support XML metadata uploads with filenames like xyz.shp.xml (#2351)

tomkralidis:
Per #2351.


tomkralidis:
cc @capooti for ref.

PR #2377, opened 2015-12-03
Added management command script to change titles of layers on particu…

cstephen:
…lar maps.


cstephen:
This script allows you to change the title for a layer on a specific map. It does this by changing the title in the two places where it appears in the "Layer Params" field of the MapLayer model.

PR #2393, opened 2015-12-17
Added option to provide a custom date to importlayers management comm…

cstephen:
…and.

### DIFF
--- a/geonode/layers/utils.py
+++ b/geonode/layers/utils.py
@@ -354,7 +354,7 @@ def extract_tarfile(upload_file, extension='.shp', tempdir=None):
 
 
 def file_upload(filename, name=None, user=None, title=None, abstract=None,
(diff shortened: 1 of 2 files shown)


geonode/layers/utils.py
```diff
@@ -502,14 +502,18 @@ def file_upload(filename, name=None, user=None, title=None, abstract=None,
         if len(regions_resolved) > 0:
             layer.regions.add(*regions_resolved)
 
+    if date is not None:
+        layer.date = date
```
simod: does it work even if the date is not  converted to a DateTime object?

cstephen: It did work, but I took a closer look and converting the provided date into a DateTime object first is definitely the correct way to do this. Good catch! I've fixed it.

PR #2420, opened 2016-02-06
Migration from20 to 24

manaetches:


simod:
Not sure we want to include this in master. others opinion?

gamesbook:
A patch to the 2.0 branch?

manaetches:
Sorry Simone,

I think my query may be posed towards the wrong group. Would you know of an
appropriate group of users who require assistance for geonode 2.0 to 2.4
migration. That way I don't have to bother the experts.

Cheers,

M

simod:
@manaetches as said in the other ticket, Q and A can be asked on geonode-users@lists.osgeo.org. There is the correct place where we all try to answer.

Issue #2422, opened 2016-02-11
investigate removing distribution_url and distribution_description from ResourceBase


tomkralidis:
As per http://lists.osgeo.org/pipermail/geonode-devel/2016-February/000752.html, these fields are duplicative at best given `geonode.base.models.Link`, and we don't really use them anywhere downstream.  Propose to remove.

capooti:
+1

ingenieroariel:
+1 on removing them while adding the corresponding Link by default

PR #2424, opened 2016-02-12
remove unused fields (#2422)

tomkralidis:


capooti:
+1 to merge. Developers: we can now do schema modification like this one, correct?

PR #2428, opened 2016-02-16
Not a pull request - distribution url & description

brucecrevensten:
This is an anti-pull-request!  Point is to show how this code works so we can discuss other ways to implement the same functionality.


brucecrevensten:
Wrong org, sorry!

Issue #2432, opened 2016-02-23
Map's Center Value Forced to Web Mercator


darinacosta:
If a map contains a layer with a bounding box, the map's center value is forced to Web Mercator, regardless of the DEFAULT_MAP_CRS value in settings.py:

https://github.com/GeoNode/geonode/blob/master/geonode/maps/views.py#L540

abersh:
when I upload layers on GeoNode 2.8 production it doesn't display the layer on center even I changed the DEFAULT_MAP_CENTER=(37,8) it doesn't work. please help anyone

PR #2438, opened 2016-02-24
Improve WGS84 support

darinacosta:
To resolve issue https://github.com/GeoNode/geonode/issues/2432.

_(Issue: If a map contains a layer with a bounding box, the map's center value is forced to Web Mercator, regardless of the DEFAULT_MAP_CRS value in settings.py.)_

Rewritten so that the map's center point is returned in coordinates if DEFAULT_MAP_CRS is EPSG:4326; otherwise, it defaults to Web Mercator.

### DIFF
--- a/geonode/maps/views.py
+++ b/geonode/maps/views.py
@@ -537,7 +537,11 @@ def new_map_config(request):
                 x = (minx + maxx) / 2
                 y = (miny + maxy) / 2
 
-                center = list(forward_mercator((x, y)))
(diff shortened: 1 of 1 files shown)


simod:
@darinacosta could you fix this flake8 issue https://travis-ci.org/GeoNode/geonode/builds/111528001#L2454?

darinacosta:
@simod sorry about that. fixed! thanks.

PR #2442, opened 2016-03-08
Metadata read button

davicustodio:
I would like to contribute to the project by providing a solution that tested here in my organization.

We concluded that for the processing and visualization of metadata we should get rid of the dependence of use of Geonetwork. For this thought to implement a specific page for metadata visualization. The idea was to provide more this feature within the GeoNode so that users could enjoy a complete interface to interact with the metadata (editing and reading).

The implementation covers the inclusion of a specific button for access to a page that displays metadata of a layer, or map, or document.

While the display is functioning in most cases, the metadata display page still has some bugs related to the layout, which may appear depending on the registered data, but I think this kind of correction can be easy for experts in CSS and HTML.

Anyway, I hope this suggestion will be useful for the project.

David.

Related Issue :  #2443

### DIFF
--- a/geonode/documents/templates/documents/document_detail.html
+++ b/geonode/documents/templates/documents/document_detail.html
@@ -84,6 +84,12 @@
 
   <div class="col-md-4">
 
+    <li class="list-group-item">
+        <a href="{% url "document_metadata_detail" resource.id %}">
+            <button class="btn btn-primary btn-md btn-block">{% trans "Metadata Detail" %}</button>
+        </a>
+    </li> 
+
     <ul class="list-group">
       <li class="list-group-item">
       {% if "download_resourcebase" in perms_list %}
--- a/geonode/documents/templates/documents/document_metadata_detail.html
+++ b/geonode/documents/templates/documents/document_metadata_detail.html
@@ -1,0 +1,1 @@
+{% extends "metadata_detail.html" %}
--- a/geonode/documents/urls.py
+++ b/geonode/documents/urls.py
@@ -38,5 +38,7 @@ urlpatterns = patterns('geonode.documents.views',
                        url(r'^(?P<docid>\d+)/remove$', 'document_remove', name="document_remove"),
                        url(r'^upload/?$', login_required(DocumentUploadView.as_view()), name='document_upload'),
                        url(r'^search/?$', 'document_search_page', name='document_search_page'),
+                       url(r'^(?P<docid>[^/]*)/metadata_detail$', 'document_metadata_detail',
+                           name='document_metadata_detail'),
                        url(r'^(?P<docid>\d+)/metadata$', 'document_metadata', name='document_metadata'),
                        )
--- a/geonode/documents/views.py
+++ b/geonode/documents/views.py
@@ -449,3 +449,16 @@ def document_remove(request, docid, template='documents/document_remove.html'):
             content_type="text/plain",
             status=401
         )
+
+
+def document_metadata_detail(request, docid, template='documents/document_metadata_detail.html'):
+    document = _resolve_document(
+        request,
+        docid,
+        'view_resourcebase',
+        _PERMISSION_MSG_METADATA)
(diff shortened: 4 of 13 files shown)


tomkralidis:
@davicustodio.  Thanks for this valuable contribution!  A few questions/comments:

- so this functionality will allow for a more detailed view of a layer/map/document metadata record (over/above/better than what is provided on layer info pages like http://demo.geonode.org/layers/geonode%3Aosm_1)?  If yes, any thoughts/objections on calling things `*metadata_detail*` instead of `*metadata_read*`?
- can you fix the [flake 8 errors](https://travis-ci.org/GeoNode/geonode/builds/114768956#L2454-L2470)
- please squash your commits?

simod:
Would be also possible to generalise the logic? Looks that the three views are doing the same thing, same with the templates. Maybe this could just go be a resolve ResourceBase and in one template just have a "object" entity for all instead of a "layer" entity (which is a little bit confusing when working with documents and maps).

geonode/templates/metadata_read.html
```diff
@@ -0,0 +1,373 @@
...
+        {% endif %}
+        {% endif %}
+
+        {% if layer.regions.all %}
+        <dt>{% trans "Regions" %}</dt>
+        <dd>{{ layer.regions.all|join:", " }}</dd>
+        {% endif %}
+
+        <dt>{% trans "Published" %}</dt>
+        <dd>{% if layer.is_published %} Sim {% else %} Não {% endif %}</dd>
```
simod: Could you translate it in english?

geonode/templates/metadata_read.html
```diff
@@ -0,0 +1,373 @@
...
+    
+    {% endif %}
+
+    <span class="subtitle">{% trans "References" %}</span>
+    <div class="sep_title"><hr></div>
+
+    <dl class="dl-horizontal">
+
+        <dt>{% trans "Link Online" %}</dt>
+        <dd><a href="{{ SITEURL }}{{ layer.get_absolute_url }}">{{ SITEURL }}{{ layer.get_absolute_url }}</a></dd>
```
simod: SITEURL should be not necessary in the link.

davicustodio:
@tomkralidis 
Thank you for your consideration. I agree with you about changing the references to metadata_detail.
Just make the changes in the names and also made a splash in the old commit.
How is the first time I submit a pull request here in the project, I am struggling to figure out how to solve the "flake 8 errors". My branch is fully updated with the master of GeoNode. My machine is configured to use a hybrid installation taking advantage of the data and the configuration of a production machine (so that I can best do the tests). Have an installed Virtualenv but'm not using effectively, I'm just telling apache point to the folder of my branch. I do not know if the fact that I do not have installed on my GeoNode virtualenv is that is causing the flakes.
Could anyone help me with any tips on how to solve it?
Thank you.

geonode/templates/metadata_read.html
capooti: it should also be in a trans block, as for "Published" in the line above

davicustodio:
@simod you're right, I also think we could improve the process logic. I'll be analyzing the alternatives you suggested and I will try to find the best way to implement.

geonode/templates/metadata_read.html
tomkralidis: agree, propose something like
```html
<a href="{{ layer.get_absolute_url }}">{{ SITEURL }}{{ layer.get_absolute_url }}</a></dd>
```

tomkralidis:
@davicustodio thanks for the changes.  Can you fully replace metadata_read with metadata_detail, to be consistent?  This includes renaming templates files, etc.

pjdufour:
I'm noticing changes to windows-related files in your commit.  Did you squash other commits into this commit?  Your commit should just include the functionality described.  Thanks!

davicustodio:
@pjdufour I did cleaning in the last commit. We are now only with the changes that matter. Thank you.

Issue #2453, opened 2016-03-22
Metadata details view using XSL


etj:
In order to provide a “light” customized output for the resource metadata, XSL transformation may be used to easily setup such output.

When in a resource page, the “Download metadata” will also present a “XML+XSL” choice, which will provide the resource metadata in ISO19115/19139 format, together with an header in the form

    <?xml-stylesheet type="text/xsl" href="http://url.to.transfromation.xsl/this.is.the.xsl"?>

The XSL will transform the metadata in a human readable format. [This](http://prototipo-catalogo.ipma.pt/layers/geonode:_5008E7C6_F6B3_4A82_A30E_0785BAC3F862/xml) implementation can be taken as a working example.

All the rendering part is processed through the XSL file; this means that the labels documenting the various fields are embedded into the XSL file. In order to have a localized output, we need different XSL for the different languages.

The different XSL files should be built from the command line, using a template for the XSL and the usual ``.po``/``.mo`` files. 
A command like: 

    python manage preparestatic

would prepare the localized XSL files, which would then be used as static content (and moved to the proper dirs by the ``collectstatic`` command).

The metadata XML would be returned by a proper service, which will 
- retrieve the bare metadata XML 
- build the URL for the localized XSL
- create the output xml with the xml-stylesheet header

In order to create the XSL URL to the localized XSL file, the service will have to choose which language should be used. As a first try, it will try to look for the language selected by the geonode user, if a session is active, otherwise the ``Accept-Language`` HTTP header will be parsed, otherwise the XSL in a default language will be provided.

etj:
How to enable the ISO+XSL metadata output
=============================

1. **Enable the contrib module**
   This functionality is implemented as a contrib module to be less invasive as possible.
   In the ``settings.py``(or local-settings) file, enable the ``geonode.contrib.metadataxsl`` app.

1. **Update the existing resources**
   In order to make it possibile for a resource to have its metadata exported as ISO+XSL, it needs to 
   have a new entry in the ``base_link`` table.
   The implementation will create the needed links every time a new resource is added, but we need to 
   post-process the existing resources and add the missing links.

   In order to add the missing links, you need to enter the geonode directory and issue the command:

        python manage.py addmissinglinks

    For each resource having a link for the ISO metadata, a new link for the ISO+XSL will be created.

1. **Do the export!**
    The "Download Metadata" dialog will show a new entry "ISO with XSL", which will point to the new service:   
[...]

PR #2475, opened 2016-04-29
#2453 Add metadata export as ISO+XSL

etj:


pjdufour:
Could you squash your commits?  Thanks.

etj:
Oky, commits squashed into one.

Issue #2478, opened 2016-05-17
Add Mapbox Streets and Satellite to basemap/background layer


maning:
Want to try this, but not sure where to start in the code.  
Is it here: https://github.com/GeoNode/geonode/blob/12e9c252fe1c96096a36e36da7c724dd7e6a4f7f/geonode/settings.py#L643

Any ideas?

JJediny:
You can add mapbox service as a XYZ basemap - See #1286

**Replace ######## with Mapbox Map ID**
```
{
    "source": {"ptype": "gxp_olsource"},
    "type":"OpenLayers.Layer.XYZ",
    "args":[
    "Mapbox Satellite", ["http://a.tiles.mapbox.com/v3/energy.i04dnk87/${z}/${x}/${y}.png",
    "http://b.tiles.mapbox.com/v3/energy.########/${z}/${x}/${y}.png",
    "http://c.tiles.mapbox.com/v3/energy.########/${z}/${x}/${y}.png",
    "http://d.tiles.mapbox.com/v3/energy//${z}/${x}/${y}.png"],
    {"transitionEffect": "resize","attribution": "Mapbox"}],
    "visibility": True,
    "fixed": True,
    "group":"background"
  }, {
    "source": {"ptype": "gxp_olsource"},
    "type":"OpenLayers.Layer.XYZ",
    "args":[
    "Mapbox Terrian", ["http://a.tiles.mapbox.com/v3/energy.map-ayrdk7iy/${z}/${x}/${y}.png",
    "http://b.tiles.mapbox.com/v3/energy.########/${z}/${x}/${y}.png",
    "http://c.tiles.mapbox.com/v3/energy.########/${z}/${x}/${y}.png",
    "http://d.tiles.mapbox.com/v3/energy.########/${z}/${x}/${y}.png"],
    {"transitionEffect": "resize","attribution": "Mapbox"}],
    "visibility": True,
    "fixed": True,
    "group":"background"
  }, 
```

jj0hns0n:
@BerryDaniel can we close this now with the recent baselayers work?

BerryDaniel:
#2514 fixed this issue, closing now.

PR #2483, opened 2016-05-26
add mapbox satellite and streets in the MAP_BASELAYERS

maning:
Related to #2478, this adds Mapbox Satellite and Streets with [Openlayers](https://www.mapbox.com/help/mapbox-with-openlayers/).  

After installing geonode, you have to copy the code to your local_settings.py based on [this instructions](https://geonode.readthedocs.io/en/1.2/deploy/customize.html#changing-the-default-basemap) and include your Mapbox access token.

Ideally, access token should be provided during the post-install setup. Any idea how to to do this?

geonode-compressor

Tested on geonode 2.4

### DIFF
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -639,9 +639,26 @@ MAP_BASELAYERS = [{
     "name": "naip",
     "group": "background",
     "visibility": False
-}, 
+},
 {
-    "source": {"ptype": "gxp_mapboxsource"},
+    "source": {"ptype": "gxp_olsource"},
+    "type":"OpenLayers.Layer.XYZ",
+    "args":[
+  "Mapbox Satellite", ["https://api.mapbox.com/v4/mapbox.satellite/${z}/${x}/${y}.png?access_token=<your access token here>"],
+        {"transitionEffect": "resize","attribution": "© Mapbox © OpenStreetMap © DigitalGlobe"}],
+    "visibility": False,
+    "fixed": True,
+    "group":"background"
+},
+{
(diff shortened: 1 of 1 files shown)


pjdufour:
I would handle the access token by creating a new settings for it; like we do for `BING_API_KEY`.

See https://github.com/GeoNode/geonode/blob/master/geonode/settings.py#L895.

Actually, I think the best way is to create the token (`MAPBOX_ACCESS_TOKEN=None`) before the local_settings.py import.  If it is set in local_settings.py, then it will override.  And then check if the token is not None (just `if MAPBOX_ACCESS_TOKEN:`).

Otherwise, you can do it like we do for SLACK and have a boolean flag too (https://github.com/GeoNode/geonode/blob/master/geonode/settings.py#L818).

Whichever way you think makes the most sense will likely be fine.

simod:
Hi, what is the behaviour if the token is not set? Or will it be mandatory to set the token?

maning:
@pjdufour, Thanks for the tip.  Will work on this.
@simod, access token is required similar to Bing and Google.

simod:
As Patrick highlighted it's better to have geonode ignore it if not configured. I personally would prefer an optional base layers setting template with the various blocks that a user can manually add to the real map baselayers settings if he wants and has a token.

jj0hns0n:
@maning can you update the PR based on @simod's comments?

jj0hns0n:
Can we close this on with @BerryDaniel's recent commits?

Issue #2489, opened 2016-06-01
Saving Metadata Overwrites Layer's Custom Thumbnail


darinacosta:
A new thumbnail is generated each time a layer's metadata is updated here: https://github.com/GeoNode/geonode/blob/master/geonode/geoserver/signals.py#L387

This overwrites custom thumbnails that are created here:
https://github.com/GeoNode/geonode/blob/master/geonode/static/geonode/js/utils/thumbnail.js

jj0hns0n:
@darinacosta can we close this?

PR #2492, opened 2016-06-07
check for existing thumbs

travislbrundage:
Addressing Issue #2489

### DIFF
--- a/geonode/geoserver/signals.py
+++ b/geonode/geoserver/signals.py
@@ -20,6 +20,7 @@
 
 import errno
 import logging
+import os
 import urllib
 
 from urlparse import urlparse, urljoin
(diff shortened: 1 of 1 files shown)


ingenieroariel:
@travislbrundage what do you think about adding that logic to the same create_thumbnail method with an `overwrite` parameter that defaults to False?

travislbrundage:
Sounds good to me. I've rewritten it like so here: https://github.com/GeoNode/geonode/pull/2494
(This was dacosta's code, so I had to make a new commit myself)

From what I can see there's only two instances of create_thumbnail for layers being called, and I believe in both cases it is fine to not manually override with `overwrite=True`, since it will execute normally if it doesn't find the thumbnail set anyway. Please let me know if I'm mistaken. The instances are:
https://github.com/GeoNode/geonode/blob/1b2ed5ba935d01b9f1e7066cea3fd93e4bedafc9/geonode/geoserver/signals.py#L387
https://github.com/GeoNode/geonode/blob/1b2ed5ba935d01b9f1e7066cea3fd93e4bedafc9/geonode/services/views.py#L1317

PR #2494, opened 2016-06-09
Check for existing thumbs before overwriting

travislbrundage:
Improvement upon [PR#2492](https://github.com/GeoNode/geonode/pull/2492)

### DIFF
--- a/geonode/layers/utils.py
+++ b/geonode/layers/utils.py
@@ -653,51 +653,59 @@ def upload(incoming, user=None, overwrite=False,
     return output
 
 
-def create_thumbnail(instance, thumbnail_remote_url, thumbnail_create_url=None, check_bbox=True, ogc_client=None):
-    if not ogc_client:
-        ogc_client = http_client
-    BBOX_DIFFERENCE_THRESHOLD = 1e-5
+def create_thumbnail(instance, thumbnail_remote_url, thumbnail_create_url=None,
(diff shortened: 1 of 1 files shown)


geonode/layers/utils.py
```diff
@@ -653,51 +653,59 @@ def upload(incoming, user=None, overwrite=False,
...
-                                   url=thumbnail_remote_url,
-                                   defaults=dict(
-                                       extension='png',
-                                       name="Remote Thumbnail",
-                                       mime='image/png',
-                                       link_type='image',
+def create_thumbnail(instance, thumbnail_remote_url, thumbnail_create_url=None,
+                     check_bbox=True, ogc_client=None, overwrite=False):
+    thumbnail_dir = os.path.join(settings.MEDIA_ROOT, 'thumbs')
+    thumbnail_name = 'layer-' + instance.uuid + '-thumb.png'
```
sarasafavi: Trivial nitpick: for readability & consistency, how about switching this from concatenating ("+") to substituting instance.uuid in with `%s`? That way it'll match how we form `filename` elsewhere in this same file (https://github.com/travislbrundage/geonode/blob/315cd1c7ab6a7a7de441ff43f1ea1410171d244e/geonode/layers/utils.py#L710)

travislbrundage: Fixed and will resubmit. Thanks for the feedback.

sarasafavi:
LGTM :) :+1:

Issue #2495, opened 2016-06-09
Assigning icons to TopicCategories


etj:
As descripted in #2408, TopicCategories shoud have an icon associated.

Since there may be problems in handling the size in uploaded images, here's an implementation which uses the icons from [Font Awesome](http://fontawesome.io/). FontAwesome [is already used in GeoNode](https://github.com/GeoNode/geonode/tree/master/geonode/static/geonode/fonts) so there's no further data to add to the repo.

This is a first rendering of the TopicCats in the existing editor:
![topcat00](https://cloud.githubusercontent.com/assets/717359/15923870/98e067a2-2e2f-11e6-983b-7540507cf3ca.png)
and this is an example of how they are presented in the search page:

![topcat01](https://cloud.githubusercontent.com/assets/717359/15923918/d3fbd344-2e2f-11e6-8df1-bd6b43c786e0.png)

The admin can change the associated icon in the TopicCategory administration page by defining the FA class:

![topcat02](https://cloud.githubusercontent.com/assets/717359/15923962/1600e09a-2e30-11e6-8152-f3eb476b327b.png)

The default icon association has been copied from GeoNetwork, in order to keep coherent the graphic representation.

capooti:
I love it :)

capooti:
can you send a PR with this?

etj:
Already done, and already merged as far as I can see :)

jj0hns0n:
@artists974 has this been merged? can we close this?

afabiani:
Yep, closed.

Best Regards,
Alessio Fabiani.

==
GeoServer Professional Services from the experts!
Visit http://goo.gl/it488V for more information.
==

Ing. Alessio Fabiani
@alfa7691
Founder/Technical Lead

GeoSolutions S.A.S.
Via di Montramito 3/A
55054  Massarosa (LU)
Italy
phone: +39 0584 962313
fax:     +39 0584 1660272
mob:   +39 331 6233686

http://www.geo-solutions.it
http://twitter.com/geosolutions_it

-------------------------------------------------------

*AVVERTENZE AI SENSI DEL D.Lgs. 196/2003*

Le informazioni contenute in questo messaggio di posta elettronica e/o
nel/i file/s allegato/i sono da considerarsi strettamente riservate. Il
loro utilizzo è consentito esclusivamente al destinatario del messaggio,
per le finalità indicate nel messaggio stesso. Qualora riceviate questo
[...]

PR #2496, opened 2016-06-09
#2495: Add icons associated to TopicCategories

etj:


capooti:
sorry, next time I will need to read emails from newest to oldest :)

PR #2510, opened 2016-06-16
added mapbox api basemaps functionality

BerryDaniel:
Added mapbox api basemaps functionality if MAPBOX_ACCESS_TOKEN is set.

fix for #2478 

- created an api_basemaps app under geonode.contrib
- moved bing basemap to api_basemaps
- set default value for both to None and check for os.environ for both

### DIFF
--- a/geonode/contrib/api_basemaps/__init__.py
+++ b/geonode/contrib/api_basemaps/__init__.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+#########################################################################
+#
+# Copyright (C) 2016 OSGeo
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


ingenieroariel:
:+1:

JJediny:
Just note that XYZ tile basemaps at least in pre-2.4ish - dependent on Geoexplorer version = loaded/requested tiles whether they were activated or not. Someone should test that XYZ basemaps arent loading in the background on zoom from web browsers development console

BerryDaniel:
@JJediny That does not appear to be the case

BerryDaniel:
@mtirwin

mtirwin:
@BerryDaniel flagging that [our tiles-from-styles API serves 512 x 512 px tiles by default](https://www.mapbox.com/api-documentation/#retrieve-raster-tiles-from-styles). If OL is expecting 256 x 256 px tiles, you'll want to flag that in the query param like this:

`https://api.mapbox.com/styles/v1/mapbox/%s/tiles/256/${z}/${x}/${y}?access_token=%s`

cc/ @tmcw @lyzidiamond @maning

BerryDaniel:
added the 256 addition, thanks @mtirwin

PR #2515, opened 2016-06-17
Update docs with database initialization step

sarasafavi:
Ref this thread: http://lists.osgeo.org/pipermail/geonode-devel/2016-June/000988.html


capooti:
wonderful, thanks a lot :)

PR #2517, opened 2016-06-17
Update documentation for the MAPBOX_ACCESS_TOKEN setting

capooti:
Small documentation update.


BerryDaniel:
@capooti Good catch!

PR #2518, opened 2016-06-17
Update to how thumbnails are saved

davisc:
Refactor thumbnail files to be stored in the location configured for media storage so that whether uploaded or auto-generated, the locations are consistent when the application is configured for remote storage.


pjdufour:
Is this for S3 storage?

davisc:
It's in support of alternate storage backends. I am submitting another pull request this morning that includes the settings to enable another storage backend, in this case S3, should a user decide to. It also makes sure all other media files can work with remote storage unless explicitly opted out.

PR #2522, opened 2016-06-20
Support S3 Remote Storage 

davisc:
This PR adds support for s3 storage and forces layer uploads to use the FileSystemStorage.


pjdufour:
Very cool!  I don't have a perfect understanding of how os.environ works, but should

```
```

```
```

capooti:
+1 to merge, after integrating @pjdufour suggestion. Thanks a lot :)

davisc:
`os.environ.get ` returns `None` if it doesn't exist, however, I agree @pjdufour with your suggestion to explicitly set the optional default value. I've tested the configuration on Centos and Ubuntu but not Windows.

capooti:
It would be useful to document all of the new settings in the settings section of the documentation: http://docs.geonode.org/en/master/reference/developers/settings.html
Thanks!

PR #2525, opened 2016-06-23
Region in title search fields

artists974:
These modifications enable to display the 'Region' of the resource (from metadata) in the autocomplete results : 

title_of_the_resource     region_of_the_resource

### DIFF
--- a/geonode/base/autocomplete_light_registry.py
+++ b/geonode/base/autocomplete_light_registry.py
@@ -23,13 +23,18 @@ import autocomplete_light
 from taggit.models import Tag
 from models import ResourceBase, Region
 
+class ResourceBaseAutocomplete(autocomplete_light.AutocompleteModelTemplate):
+    choice_template = 'autocomplete_response.html'
 
 autocomplete_light.register(Region,
                             search_fields=['name'],
                             autocomplete_js_attributes={'placeholder': 'Region/Country ..', },)
 
 autocomplete_light.register(ResourceBase,
+                            ResourceBaseAutocomplete,
                             search_fields=['title'],
+                            order_by=['title'],
+                            limit_choices = 100,
                             autocomplete_js_attributes={'placeholder': 'Resource name..', },)
 
 autocomplete_light.register(Tag,
--- a/geonode/documents/autocomplete_light_registry.py
+++ b/geonode/documents/autocomplete_light_registry.py
@@ -21,9 +21,15 @@
 import autocomplete_light
 from models import Document
 
+class DocumentAutocomplete(autocomplete_light.AutocompleteModelTemplate):
+    choice_template = 'autocomplete_response.html'
+
 autocomplete_light.register(
     Document,
-    search_fields=['^title'],
+    DocumentAutocomplete,
+    search_fields=['title'],
+    order_by=['title'],
+    limit_choices = 100,
     autocomplete_js_attributes={
         'placeholder': 'Document name..',
     },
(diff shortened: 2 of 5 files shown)


simod:
Hi, what is the use case for having the region in there?
I have also rarely seen the region correctly filled in the metadata as it is not used by many users.

simod:
In any case, the PR is good and we could include anyway everything apart from the region addition despite what this thread decides.

artists974:
Hi !

Displaying the region there can be useful if you have duplicate titles, but for different regions. To have a concrete example, if you go on our [website](http://aware.cirad.fr), and you type "ign", you will have many administrative resources which have the same title (like "Roads 2015", "Cities 2014", etc. 

If region is not displayed; the user has to press 'enter' key to see all the resources list details, whereas if he sees the title + region he knows which one he is looking for and can click to be instantly redirected to the resource by clicking on it. 

The other thing is that GeoNode propose on the left panel a "Region" fitler, so if nobody fills the region field in metadata, then I think this Region filter will always stay be empty, and then useless...

Finally, I guess many people don't fill the 'Region' field, but anyway if it is not filled in metadata it won't be displayed in the search bar, so in this case it will be just like before.

Thanks !

artists974:
I just noticed there was a little error in the first pull request I did, so I did a modification, I don't know if you can see it or not... ("changed condition to choice.region")

simod:
I'm -1 on having the region together with the title as this could be unwanted to users. I would see more unwanted that wanted. I'm +1 on all other improvements made here. Let's wait for others considerations. For whatever decision could you then fix the pep8 violations and squash the commits? Thanks a lot!

capooti:
I am with @simod here. @artists974 you should keep separated customization specific for your instance from core GeoNode. The other improvements look really useful, thanks

artists974:
Hi, we will delete the "region display" part then for core Geonode then. 

I have another question also : I will do a new pull request to enable the text search according to the catalogue where you are (layers, documents, maps), but this new one needs the files edited here. 
So what should I do ?
- Correct Pep8 problems here, delete Regions customizations, and then send back and let you merge it to master so I can send only needed files in the new pull request,
- Or put all the files needed in the new pull request, (including these files, with Pep8 problems corrected, region customizations deleted + the new files) ?

If my question is not clear, let me know !

Thanks

simod:
I would go for the second as this PR without the regions stuff contains
some general improvements.

2016-06-27 8:17 GMT+02:00 AWARE - ARTISTS team <notifications@github.com>:

> Hi, we will delete the "region display" part then for core Geonode then.
>
> I have another question also : I will do a new pull request to enable the
> text search according to the catalogue where you are (layers, documents,
> maps), but this new one needs the files edited here.
> So what should I do ?
>
>    - Correct Pep8 problems here, delete Regions customizations, and then
>    send back and let you merge it to master so I can send only needed files in
>    the new pull request,
>    - Or put all the files needed in the new pull request, (including
>    these files, with Pep8 problems corrected, region customizations deleted +
>    the new files) ?
>
> If my question is not clear, let me know !
>
> Thanks
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/GeoNode/geonode/pull/2525#issuecomment-228662455>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAvkdKbRxKr332AAjvTxjVWUtavpFPJUks5qP2sEgaJpZM4I8i1O>
> .
>



-- 
Simone

PR #2527, opened 2016-06-27
Gdal bindings installation from dev docs

### LINKED ISSUE #2526: GDAL bindings from ubuntu dev docs
The issue is well described in this thread [osgeo module install problem](http://osgeo-org.1560.x6.nabble.com/osgeo-module-install-problem-td5209620.html) and happening by far

The only way to overcome that was to install the pygdal module which is undocumented and would be the recommended

francbartoli:
Fix #2526


jj0hns0n:
You need to squash this into a single commit against master.

PR #2540, opened 2016-07-06
addition to previous patch: format port as str

sarasafavi:
See https://github.com/GeoNode/geonode/pull/2539


geonode/geoserver/helpers.py
```diff
@@ -987,7 +987,8 @@ def _create_db_featurestore(name, data, overwrite=False, charset="UTF-8", worksp
          'min connections': '1',
          'fetch size': '1000',
          'host': db['HOST'],
-         'port': db['PORT'],
+         'port': db['PORT'] if isinstance(
+             db['PORT'], basestring) else str(db['PORT']) or '5432',
```
garnertb: Could these both cleaned up to just: `str(db.get('PORT', 5432))`?

PR #2548, opened 2016-07-12
Ol3 preview

lukerees:
Allow users to change the layer and map previews to use OpenLayers3


BerryDaniel:
@lukerees is it possible to also pass it the default projection https://github.com/GeoNode/geonode/blob/master/geonode/settings.py#L609

jj0hns0n:
@lukerees can you update this against master so we can merge during the sprint.

PR #2553, opened 2016-07-19
Show whole catalogue and allow actions according to permissions

artists974:
Hi,

These modifications lead to a deep change of GeoNode experience : 
- The whole catalogue can be seen by every user, logged in or not
- Each user - logged in or not - has access to resources abstract and to download metadata ( this button can be hidden easily in the templates with Django tags, if this is not the desired behavior, anyway)
- The resources list is sorted by permissions : first the authorized ones - sorted by most recent to less recent ; and then the restricted ones - sorted by most recent to less recent too, with a "RESTRICTED ACCESS" label
- If "Who can view/download the resource ?" - "Anyone" are not checked, then anonymous users can see the thumbnail and the abstract in the resources lists, and then can click on the resource and see the abstract and the "download metadata" button. They cannot see the resource : a blocking message appears "You don't have the required permissions to visualize the resource". If a user doesn't have permissions to view the resource and would like to see it,, it is necessary for him to contact the Point of Contact
- The "add to cart" button is hidden if the user is not a contributor : because it gives him the capacity to change permissions, and to create maps that will appear on the Geonode maps list
  if the user is a contributor : he can change permissions if he has these rights on the resource, or if he is the owner of the resource
  Also : he cannot create a map if he doesn't have the permissions
- Buttons : if the user is a contributor, then he has access to all the "add" buttons (layers, maps, documents). A "user_can_add_resourcebase" tag is created is base_tags.py. If the user doesn't have the "add_resource" permission, then only "Explore ressource" buttons will be shown, and "add" buttons will be hidden. 

It is still possible to make this instance work with the same behavior than the original GeoNode instance, but with more flexibility. It is important to well define users and groups permissions, but once it's done, it's a real comfortable instance to use.

I guess these modifications are too deep to be merged directly with Master branch of GeoNode, so I don't know how some of them could be included (maybe as a new fork ?).
A good solution could be to create something like : 'DEFAULT_VIEW_ALL_CATALOGUE = True' in settings.py for example, in the same way than 'SKIP_PERMS_FILTER = True'.

Don't hesitate if you need more informations,

Thank you !

### DIFF
--- a/geonode/api/api.py
+++ b/geonode/api/api.py
@@ -252,17 +252,17 @@ class ProfileResource(TypeFilteredResource):
     def dehydrate_layers_count(self, bundle):
         obj_with_perms = get_objects_for_user(bundle.request.user,
                                               'base.view_resourcebase').instance_of(Layer)
-        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values('id')).distinct().count()
+        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values_list('id')).distinct().count()
 
     def dehydrate_maps_count(self, bundle):
         obj_with_perms = get_objects_for_user(bundle.request.user,
                                               'base.view_resourcebase').instance_of(Map)
-        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values('id')).distinct().count()
+        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values_list('id')).distinct().count()
 
     def dehydrate_documents_count(self, bundle):
         obj_with_perms = get_objects_for_user(bundle.request.user,
                                               'base.view_resourcebase').instance_of(Document)
-        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values('id')).distinct().count()
+        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values_list('id')).distinct().count()
 
     def dehydrate_avatar_100(self, bundle):
         return avatar_url(bundle.obj, 240)
--- a/geonode/api/authorization.py
+++ b/geonode/api/authorization.py
@@ -20,7 +20,6 @@
 
 from tastypie.authorization import DjangoAuthorization
 from tastypie.exceptions import Unauthorized
-
 from guardian.shortcuts import get_objects_for_user
 
 
@@ -30,11 +29,17 @@ class GeoNodeAuthorization(DjangoAuthorization):
     permission system"""
 
     def read_list(self, object_list, bundle):
-        permitted_ids = get_objects_for_user(
+        permitted_view_ids = get_objects_for_user(
             bundle.request.user,
-            'base.view_resourcebase').values('id')
+            'base.view_resourcebase').values_list(
+            'id',
+            flat=True)
+        return object_list.filter()
 
-        return object_list.filter(id__in=permitted_ids)
+    def explore_resource(self, object_list, bundle):
+        return bundle.request.user.has_perm(
+            'explore_resourcebase',
+            bundle.obj.get_self_resource())
 
     def read_detail(self, object_list, bundle):
         return bundle.request.user.has_perm(
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -38,6 +38,7 @@ from tastypie.utils.mime import build_content_type
 
 from geonode.layers.models import Layer
 from geonode.maps.models import Map
+from geonode.maps.utils import is_map_viewable_by_user_utils
 from geonode.documents.models import Document
 from geonode.base.models import ResourceBase
 
@@ -46,6 +47,7 @@ from .authorization import GeoNodeAuthorization
 from .api import TagResource, RegionResource, ProfileResource, \
     TopicCategoryResource, \
     FILTER_TYPES
+
 
 if settings.HAYSTACK_SEARCH:
     from haystack.query import SearchQuerySet  # noqa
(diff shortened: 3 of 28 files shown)


geonode/api/api.py
```diff
@@ -252,17 +252,17 @@ def dehydrate_email(self, bundle):
     def dehydrate_layers_count(self, bundle):
         obj_with_perms = get_objects_for_user(bundle.request.user,
                                               'base.view_resourcebase').instance_of(Layer)
-        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values('id')).distinct().count()
+        return bundle.obj.resourcebase_set.filter(id__in=obj_with_perms.values_list('id')).distinct().count()
```
simod: django is fine in using values('id') to filter with __in, adding the values_list is a small overhead

geonode/api/authorization.py
```diff
@@ -30,11 +29,17 @@ class GeoNodeAuthorization(DjangoAuthorization):
...
 
     def read_list(self, object_list, bundle):
-        permitted_ids = get_objects_for_user(
+        permitted_view_ids = get_objects_for_user(
             bundle.request.user,
-            'base.view_resourcebase').values('id')
+            'base.view_resourcebase').values_list(
+            'id',
+            flat=True)
+        return object_list.filter()
```
simod: any missing filter here?

geonode/api/resourcebase_api.py
```diff
@@ -304,25 +306,27 @@ def build_haystack_filters(self, parameters):
...
-            sqs = (SearchQuerySet() if sqs is None else sqs).order_by(
-                "-title_sortable")
-        elif sort.lower() == "-popular_count":
-            sqs = (SearchQuerySet() if sqs is None else sqs).order_by(
-                "-popular_count")
-        else:
-            sqs = (
-                SearchQuerySet() if sqs is None else sqs).order_by("-date")
-
+        # if sort.lower() == "-date":
```
simod: do we loose sorting capabilities when using haystack?

geonode/api/resourcebase_api.py
```diff
@@ -435,17 +441,57 @@ def get_list(self, request, **kwargs):
...
+            'id',
+            flat=True)
+
+        obj_no_perms = []
+        obj_perms = []
+
+        # retrieving in the json response the permisions
+        # to view resource 'can_view_resource'
+        # to change resource perms 'can_change_perms'
+        for idx, obj in enumerate(objs):
```
simod: If I understand correctly here there is an additional sorting on the whole queryset. We disabled such kind of loops over ResourceBase because of poor performances, do you think that this would keep the api fast as they are now?

geonode/api/resourcebase_api.py
```diff
@@ -481,6 +536,20 @@ def create_response(
             content_type=build_content_type(desired_format),
             **response_kwargs)
 
+    def get_resource_type(self, obj):
```
simod: Could we reuse this logic here https://github.com/GeoNode/geonode/blob/master/geonode/base/templatetags/base_tags.py#L125

geonode/api/resourcebase_api.py
```diff
@@ -435,17 +441,57 @@ def get_list(self, request, **kwargs):
...
+                c_map = Map.objects.all().get(resourcebase_ptr_id=obj['id'])
+                obj['can_view_resource'] = is_map_viewable_by_user_utils(request.user, c_map)
+
+                # for a map we only check if user has right to change perms for the map itself
+                if obj['id'] in permited_change_perms:
+                    obj['can_change_perms'] = True
+                else:
+                    obj['can_change_perms'] = False
+
+                if obj['can_view_resource']:
```
simod: the permissions was view_resourcebase, maybe this is a small typo

geonode/base/models.py
```diff
@@ -354,6 +354,9 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin):
     detail_url = models.CharField(max_length=255, null=True, blank=True)
     rating = models.IntegerField(default=0, null=True, blank=True)
 
+    # check
+    can_view_resource = False
```
simod: What is this for?

artists974:
Hi Simone, 

All these modifications are the results of a team work. For that reason I will need a few time to give you precise explanations about these different parts of the code. I will really do my best to answer you as soon as I can.

Thanks !

simod:
there's no rush at all. those are just considerations. Thanks!

geonode/api/resourcebase_api.py
artists974: I did not write this part of the code, so I will give you an answer asap. 
Does this can help you : [http://lists.osgeo.org/pipermail/geonode-users/2015-June/000519.html](url) ? 
Thanks

jj0hns0n:
@artists974 can you update us on this and update the branch against master?

geonode/api/authorization.py
artists974: We intentionally deleted the filter here, in order to show the whole catalogue for every user.

geonode/api/resourcebase_api.py
artists974: In our instance, we don't use the Haystack. Unfortunately, i don't know the behaviour with the Haystack, and i can't tell if it we lose sorting capabilities when using it.

geonode/api/resourcebase_api.py
artists974: We implemented a sort by permissions : 
First, the authorized resources for the user, then the restricted ones. 

Our catalogue includes about 300 resources, and we didn't notice any lower performance than on the original GeoNode instance, it is exactly the same. However, I cannot tell if the performance will decrease with a 2000 resources catalogue.

geonode/api/resourcebase_api.py
artists974: Yes, I think we could reuse the same logic. I don't know if the person from our team who wrote this function in our code was inspired by the function in base_tags, or if he didn't see it. Anyway, it is the same logic.

geonode/api/resourcebase_api.py
artists974: Actually there is no mistake here. We use can_view_resource to create the sorting, but it has no effect and impact on the permissions. 
For the permissions, we always use view_resourcebase, as in original GeoNode instance. view_resourcebase is reserved to users who can view the resource (and its data) really, whereas the can_view_resource is just used to create the sorting.

geonode/base/models.py
artists974: This was created to init the can_view_resource. Yet, I think we could delete it because I'm not sure it is really used...

artists974:
Hi Simone, first of all sorry for the delay.

I tried to give you answers here, hope it will help you. I cannot really update this branch against master, because there was no change on this one since I did the pull request, it would just be the same than the one here.

By the way, we would really like to show you some new improvements we made on a new branch. Our instance of GeoNode is now able to put a resource in standby, which will be only viewable by the superuser and by the owner. A special catalogue is available, containing "standby" resources, in layers, maps and documents (see the attached file). In this way, our instance has a behavior more close to the GeoNode one, and a great flexibility in terms of confidentiality. 

What is the best way to show this instance to the community, doing a pull request on Master here ? Or adding 
standby_layers
the instance on a new branch on our github Profile artists974 ?

Thanks a lot ! Hope you will enjoy.

Antoine Rolland

jj0hns0n:
@simod can you take a look at this again.

simod:
Not much else to say, this functionality is very nice but this PR needs update. @artists974 the layer review process would be very welcome, I think it would be an excellent contrib app.

tomkralidis:
+1 with @simod.  This PR needs updating.

simod:
we are going to close this for now. It would be good to have a discussion on the workflow before resubmitting.

PR #2557, opened 2016-07-19
Default to public schema if not defined

travislbrundage:
Some database settings may not explicitly define the schema, so default to using public.

### DIFF
--- a/geonode/upload/utils.py
+++ b/geonode/upload/utils.py
@@ -45,11 +45,11 @@ class UnsavedGeogigDataStore(UnsavedDataStore):
             datastore = settings.OGC_SERVER['default']['DATASTORE']
             pg_geogig_db = settings.DATABASES[datastore]
             message["dbHost"] = pg_geogig_db['HOST']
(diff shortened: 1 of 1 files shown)


pjdufour:
I think you can just do the following, right?

```
message["dbSchema"] = pg_geogig_db.get('SCHEMA', 'public')
```

travislbrundage:
Thanks for the python advice, I will refactor that. Looks much neater.

geonode/upload/utils.py
```diff
@@ -45,11 +45,11 @@ def message(self):
             datastore = settings.OGC_SERVER['default']['DATASTORE']
             pg_geogig_db = settings.DATABASES[datastore]
             message["dbHost"] = pg_geogig_db['HOST']
-            message["dbPort"] = pg_geogig_db['PORT'] or '5432'
+            message["dbPort"] = pg_geogig_db.get('PORT', '5432')
```
travislbrundage: This 'or' statement wasn't doing anything; get() with the default value is what was intended.

PR #2578, opened 2016-08-05
Added documentation for configuring an AWS S3 Bucket with GeoNode

travislbrundage:
Currently having an issue with the added documentation section in the table of contents. It's not appearing correctly when viewing the siblings, although it appears in the ToC fine at the root level.


capooti:
Thanks a lot!
I am not sure why we have that Java error in trails, but it does not look to depend from your changes.

Issue #2566, opened 2016-07-27
Admin template exception when viewing/adding maps, layers, documents


sarasafavi:
When viewing an individual map/layer/document, a NoneType exception is thrown, e.g.:

```
TypeError at /admin/maps/map/2/
'NoneType' object is not callable
```
```
/vagrant/.venv/lib/python2.7/site-packages/autocomplete_light/widgets.py in render
                    autocomplete = self.autocomplete(values=value) 
```

The same error occurs when adding via admin as well - this prevents the use of these admin views. 

Seems to be related to django-autocomplete-light & the Django 1.8 GeoNode upgrade: reference this change [geonode/base/admin.py:133](https://github.com/GeoNode/geonode/blob/master/geonode/base/admin.py#L133) made within https://github.com/GeoNode/geonode/commit/fc4085bf6ffe8f09a307309e5f22ef2371892ef1.

pjdufour:
For DAL (django-autocomplete-light), see:  https://github.com/GeoNode/geonode/issues/2451

jj0hns0n:
@sarasafavi is this still occuring? Can we look during the sprint.

sarasafavi:
@jj0hns0n this was fixed in https://github.com/GeoNode/geonode/pull/2580; closing

Issue #2451, opened 2016-03-21
Upgrade django-autocomplete-light (DAL) to latest version


karakostis:
https://django-autocomplete-light.readthedocs.org/en/master/
Needed to support genericForeignKey in relation to the Documents app.
Its trickier than it seems. Some changes are not backwards compatible.

pjdufour:
This is hard.  Documentation on upgrading is not great and factory functions have been removed.  Happy to take suggestions on maintaining DRY.  Might need to wait a few more months until broader adoption by open source community of version 3.x.

Working on it https://github.com/pjdufour/geonode/commits/dal-3.1.5

jj0hns0n:
@karakostis @pjdufour whats the status on this?

rob-metalinkage:
was looking at status - so just noting its still pinned to 2.3.3 in
https://github.com/GeoNode/geonode/blob/master/requirements.txt

Is this effort given up, on the backburner, superceded?

afabiani:
With the upgrade of Django we should get rid of DAL at all imho

rob-metalinkage:
Yep - seems to be built into 2.0 - is there a planned timescale for that?

PR #2580, opened 2016-08-09
fix exception during renders of admin forms using autocomplete_light

harts-boundless:
This is a proposed fix for #2566. 

To recap, when accessing admin URLs pertaining to certain models (e.g. /admin/maps/map/2/), template rendering raises "TypeError: 'NoneType' object is not callable" (I'll spare readers the full traceback because I'm certain you can easily replicate it against any version of Geonode going back to the great Django 1.8 upgrade commit). So these admin views for these models do not load.

To diagnose, this relates to `class ResourceBaseAdminForm` in `base/admin.py`, where `TaggitField` is instantiated with an instance of `TaggitWidget`. Before this patch, no value is specified for the first arg `autocomplete`; therefore, during the template render, that `None` value gets called. 

The first step of the fix is suggested by a commented-out line where `autocomplete` is specified as `'TagAutocomplete'`. However, when using this line, the boot-time instantiation of `TaggitField` fails, preventing the app from running at all. This probably explains why that line was commented out, because it is a step toward solving the admin problem but it leaves the app as a whole broken.

The basis of this PR (and admittedly it is a kludge) is to let TaggitField do its thing as currently, by not giving it a widget with autocomplete set, and then afterward setting such a widget so that the template render can work. I chose this solution to try to make a minimal patch and because django-autocomplete-light should probably be upgraded to its rewrite v3 someday anyway. If someone else has a better solution in the near term I'd be happy to see that solution merged instead of this PR.

### DIFF
--- a/geonode/base/admin.py
+++ b/geonode/base/admin.py
@@ -130,5 +130,10 @@ admin.site.register(License, LicenseAdmin)
 
 
 class ResourceBaseAdminForm(autocomplete_light.ModelForm):
-    # keywords = TaggitField(widget=TaggitWidget('TagAutocomplete'), required=False)
-    keywords = TaggitField(widget=TaggitWidget(), required=False)
+    # We need to specify autocomplete='TagAutocomplete' or admin views like
+    # /admin/maps/map/2/ raise exceptions during form rendering.
+    # But if we specify it up front, TaggitField.__init__ throws an exception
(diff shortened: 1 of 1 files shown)


sarasafavi:
Tested locally - works for me. I'm happy with this kludge pending https://github.com/GeoNode/geonode/issues/2451 hopefully getting done in the near future.

PR #2587, opened 2016-08-18
No specific versions in setup.py, only minimum

ingenieroariel:
For downstream projects we need to be able to override some dependency versions, like pycsw.

We should only filter out versions that are not compatible with GeoNode and let downstream projects decide which specific version to pin.


jj0hns0n:
@ingenieroariel can you look why the test is failing so we can get this merged during the sprint.

PR #2596, opened 2016-08-22
Override Django configuration with environment variables

ingenieroariel:


pjdufour:
Could you explain why no more local_settings.py support?   Seems like big change in workflow.

geonode/settings.py
```diff
@@ -898,14 +914,8 @@
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
     MEDIA_URL = "https://%s/%s/" % (AWS_S3_BUCKET_DOMAIN, MEDIAFILES_LOCATION)
 
-import djcelery
-djcelery.setup_loader()
 
-# Load more settings from a file called local_settings.py if it exists
```
jj0hns0n: This doesnt seem like a good idea.

jj0hns0n:
We will need to update the documentation to reflect these changes.

geonode/settings.py
```diff
@@ -21,62 +21,75 @@
...
 from kombu import Queue
-import geonode
+from geonode import __file__ as geonode_path
+from geonode import get_version
 from geonode.celery_app import app  # flake8: noqa
+import djcelery
+import dj_database_url
+
+
+def str2bool(v):
```
sarasafavi: instead of creating `str2bool` here to try & anticipate all the varieties of "True" that might be used in settings files, why don't we just drop this custom handling & just enforce consistency from the start (i.e., comparing some_settings_value == "True")

ingenieroariel: Good idea.

geonode/settings.py
ingenieroariel: `local_settings.py` is the bad idea, I have had to chase my own tail in production several times because of `local_settings.py`.[1][2]

[1] http://nedbatchelder.com/blog/201112/duplicitous_django_settings.html
[2] https://zeth.net/archive/2011/12/31/django-and-the-localsettings-anti-pattern/

ingenieroariel:
Actually, Patrick and Jeff are right - this has direct implications to packaging.

I would like us to move away from the local_settings.py paradigm and move towards the use of either environment variables for small changes or downstream projects for big changes.

Since the local_settings.py is a big change, I'll revert it so we can test both approaches and decide in the future.

geonode/settings.py
ingenieroariel: I ended up importing another implementation of `strtobool` from distutils, that way we don't have to reinvent the wheel.

ingenieroariel:
@pjdufour @jj0hns0n Can you look at the latest version?

jj0hns0n:
happy to merge if you update with latest master.

ingenieroariel:
this has no conflicts with latest master - despite what the github ui says. The last commit in master was an apache config change and this changes another file.

ingenieroariel:
I just pushed and travis started building. In around 15 minutes it may be possible to merge it.

Issue #2555, opened 2016-07-19
Search functionality no longer works on people/groups


sarasafavi:
Search only searches layers, maps, and documents -- should also search across people & groups

sarasafavi:
Can't assign to myself: working this now

artists974:
Maybe it doesn't work because of 

- this part in search_scripts.html :

  {% get_context_resourcetype as pathc %}
	{% if pathc == "layers" %}
		AUTOCOMPLETE_URL_RESOURCEBASE = '{% url "autocomplete_light_autocomplete" "LayerAutocomplete" %}';  
	{% elif pathc == "maps" %}
		AUTOCOMPLETE_URL_RESOURCEBASE = '{% url "autocomplete_light_autocomplete" "MapAutocomplete" %}';  
	{% elif pathc == "documents" %}
		AUTOCOMPLETE_URL_RESOURCEBASE = '{% url "autocomplete_light_autocomplete" "DocumentAutocomplete" %}';  
	{% elif pathc == "search" %}
		AUTOCOMPLETE_URL_RESOURCEBASE = '{% url "autocomplete_light_autocomplete" "ResourceBaseAutocomplete" %}';  
	{% else %}
		AUTOCOMPLETE_URL_RESOURCEBASE = '{% url "autocomplete_light_autocomplete" "ResourceBaseAutocomplete" %}';  // AR - Comment : This line is the original configuration for Geonode's search bar display results. 
	{% endif %}

Which only works for Layers, Documents and Maps. Must be necessary to add People and Groups. 

- Also I think it is necessary to add People and Groups in this function, in base_tags.py : 

def get_context_resourcetype(context):

- Also add this class in People and Groups' autocomplete_light_registry.py (check how autocomplete_light_registry.py is filled for Layers, Maps and Documents) : 

class PeopleAutocomplete(autocomplete_light.AutocompleteModelTemplate):
     choice_template = 'autocomplete_response.html'
class GroupAutocomplete(autocomplete_light.AutocompleteModelTemplate):
     choice_template = 'autocomplete_response.html'





With these modifications, I think this should work.

sarasafavi:
@artists974 so those fixes mostly work to add People/Groups back into DAL config (so that people/group autocomplete results show up after typing in the header's search box). Ref #2582 for a proposed patch that does that.

[...]

PR #2597, opened 2016-08-22
Add context-specific search for people, groups

sarasafavi:
_Feel free to squash prior to merging; left as-is for info only_

Ref: https://github.com/GeoNode/geonode/issues/2555

Adds a context-dependent search (filter box) to the `/people` and `/groups` pages. This search works similarly to those on the other resource pages (map, layer, documents). In the case of `/people` search, searches as executed on the `username` property (as people profiles do not have a `title`).

group_search
_Re-styled group page showing all groups_

group_search_result
_search results using context-specific group search: autocomplete results are shown in drop-down from search box_

people_search
_Same; for /people_

people_search_result
_Same; for /people -- text may be entered to search first name, last name, or username values: autocomplete shows the username of matching profile(s)_

### DIFF
--- a/geonode/groups/templates/groups/group_list.html
+++ b/geonode/groups/templates/groups/group_list.html
@@ -14,33 +14,12 @@
     <h2>{% trans "Explore Groups" %}</h2>
   </div>
 
-  <div class="btn-group pull-right" role="group" aria-label="sort">
-    <div class="btn-group pull-right" role="group">
-      <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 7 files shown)


stan4gis:
> This  search works similarly to those on the other resource pages (map, layer, documents). 

> 

As I can see in 2.4.1, auto-complete is not context-dependent in map, layer and documents pages

sarasafavi:
> As I can see in 2.4.1, auto-complete is not context-dependent in map, layer and documents pages

Hi @stan4gis: in GeoNode 2.4.1, auto-complete within the resource filter (that is, the search box embedded within the page itself, and not the search box in the site header) is indeed context-dependent: filter results will depend on whether you are viewing a map, layer, or document listing.

Issue #2332, opened 2015-11-09
Use `isPublished` field to control whether a Resource is exposed in pycsw


JJediny:
I believe `isPublished` field in ResourceBase Metadata is only a placeholder and/or only controls Geonode API filter... if this is correct could this field also be used to restrict layers/documents/maps being exposed by pyCSW?

The need for this is to avoid data/documents that are DRAFT or in-developement from being exposed to other CSW services/harvests before "they are ready".

JJediny:
@tomkralidis any constraints from doing this or recommendations of an approach we can/should take?

tomkralidis:
@JJediny fair enough.  We will need to update [pycsw](https://github.com/geopython/pycsw/blob/master/pycsw/plugins/repository/geonode/geonode_.py) to (optionally) filter based on publish state.

I would also recommend an option in pycsw config (which we manage in GeoNode) called `show_published` or something which can direct pycsw to display unpublished records or not.  *Unless* there is a GeoNode workflow rule that says unpublished never gets put to CSW/OAI/SRU/OpenSearch.

tomkralidis:
FYI I have this working now.  Using pycsw's [repository filter](http://docs.pycsw.org/en/latest/repofilters.html) mechanics, we are able to define this in https://github.com/GeoNode/geonode/blob/master/geonode/catalogue/backends/pycsw_local.py#L46

Before I push this (for 2.5) we need to decide whether `is_published=True` is a./ the default state/behaviour or whether b./ this should be configurable (in `geonode.settings`).  I would vote for option A but would like to hear thoughts from others.

tomkralidis:
Note that I also found a pycsw [bug](https://github.com/geopython/pycsw/issues/388) in the process of addressing this issue which has since been fixed.  Once we have consensus on GeoNode behaviour, and pycsw 1.10.3 is released, I will submit a PR of the branch for merge.

jj0hns0n:
@tomkralidis any update on this? Moving to 2.7 but feel free to move back.

tomkralidis:
Fixed in master.

PR #2608, opened 2016-08-22
Use isPublished field to control whether a Resource is exposed in pycsw

tomkralidis:
As per #2332

### DIFF
--- a/geonode/catalogue/backends/pycsw_local.py
+++ b/geonode/catalogue/backends/pycsw_local.py
@@ -46,6 +46,7 @@ CONFIGURATION = {
     },
     'repository': {
         'source': 'geonode',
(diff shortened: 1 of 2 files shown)


jj0hns0n:
@tomkralidis can you check the failing test?

jj0hns0n:
@tomkralidis can you try to address the failing test before we merge?

tomkralidis:
The failing test is due to a resulting change to the workflow. I'll look deeper but we shouldn't merge this just yet b

jj0hns0n:
Simone fixed it. Feel free to advise if it is not a good solution

jj0hns0n:
@tomkralidis can you give an update on this ?

PR #2624, opened 2016-08-26
Set to 2.5 final version for release

BerryDaniel:


jj0hns0n:
@BerryDaniel ... @simod and I were trying to get the hierarchical keywords branch merged before the 2.5 as it has some schema changes. Can you give us a chance to fix the one failing test before you cut 2.5?

BerryDaniel:
2.5 tag and release have been completed. We might be able to delete it and
run it again or do a 2.5.1 release?

jj0hns0n:
I guess 2.5.1 is ok, we will probably tag a few more this way right? Did
you guys push packages yet?

simod:
don't think so. note: we never pushed the "final" version in master so to have the dev version appearing on it. Should we roll it back then? Paver publish will take care of all the tagging for us.

jj0hns0n:
yeah, lets do.

PR #2629, opened 2016-08-29
[WIP] Simpler migrations

ingenieroariel:
Hello all,

I spent an afternoon trying to make `manage.py migrate` work without having to specify an order.

The main problems were:
- user_messages and dialogos had a migration folder but it was not present in the packages via pypi. I added the migrations folder to the setup.py and released new versions for those.
- Migration dependencies were not properly ordered. This caused circular dependencies.
- layers depends on services and services depends on layers, I make service depend on layers and removed the specific field for service in the layer model - it can be queried using the django ORM (untested).

This removes the need to migrate one app before the other and makes tests run successfully without complaining about a missing profiles_profile.

What are your thoughts on getting rid of the interdependency between layers and services?

What are your thoughts on just adding a new migration file on each development version and locking it up in the production version (this one would be called geonode25 until geonode-2.6 is released and we freeze it, in the next release cycle it would be called geonode27).

CC @pjdufour @francbartoli @afabiani @simod @sarasafavi 

MISSING:
- [x] Agree with other devs on best way forward and timing.
- [x] Fix tests that relied on 'service'.
- [x] Verify functionality manually
- [x] Specify in the migration files that it 'replaces' the old ones.
- [ ] Decide what to do with people that are migrating from 2.4


simod:
@ingenieroariel I'm +1 on both. I like that the service doesn't depend on the layer anymore. Thanks for this. I see there's still some issue in the resolve_layer somewhere, could you take a look?

ingenieroariel:
Simone, you are right about the model freeze. I agree with keeping a model freeze during the 2.5 release until we get to 2.6 but would like to make a special case for this change for the reasons outlined below.

The main issue is that if we don't fix migrations in this cycle - we are going to have to wait a lot before being able to cleanly apply migrations. Since this (2.5/2.6) is the first release with Django migrations, compatibility should be first with future releases instead of past ones.

Having circular dependencies between apps is IMHO a showstopper. I brought up the issue on Gitter last week before the release but we did not know enough about why we needed to migrate apps one by one in order to solve it.

jj0hns0n:
+1 on this approach. Lets not make any more model changes after this?

simod:
Ok Ariel, +1!

ingenieroariel:
Thanks all - I'll fix the tests and take care of the outstanding issues.

jj0hns0n:
@ingenieroariel where are we at on this?

simod:
@all please give it a try ^^

ingenieroariel:
Tests pass and coverage goes down by small percentage. We encourage downstream projects to try this so we can finally get rid of migrating app by app.

jj0hns0n:
@all be sure you do a `git clean` in your source tree as I found the old migration files were hanging around unless forcibly deleted.

ingenieroariel:
@jj0hns0n are you sure it was okay for you to have only clean migrations instead of migrations to go from 2.4 to 2.6? That was the missing bit in the list of things in the PR description.

jj0hns0n:
Migrations from 2.4 to 2.6 should be handled separately IMO. This approach is much cleaner for master and new installs.

PR #2660, opened 2016-09-28
Respond correctly when user removes the last related file for a layer…

travislbrundage:
… upload

The motivation for this change is because the user can currently get "stuck" with an empty layer upload.

Steps to reproduce:
1) Go to the "layer upload" page.
2) Select some files to upload.
3) Click "remove" for all files related to a layer.

You are now in the "broken layer" state. Even if you have correct layers, having one layer missing files it needs will cause an error on upload. Currently it can only be fixed by clicking the "Clear" button to clear all staged files. This is not intuitive and may not be what the user wants.

With this change, clicking "remove" on the final file associated with the layer will automatically remove this layer from the staged layers for upload.

### DIFF
--- a/geonode/static/geonode/js/upload/LayerInfo.js
+++ b/geonode/static/geonode/js/upload/LayerInfo.js
@@ -699,6 +699,7 @@ define(function (require, exports) {
             }
             a.data('layer', self.name);
             a.data('file',  file.name);
+            a.attr('class', 'remove-file');
             a.appendTo(p);
             a.on('click', function (event) {
                 var target = $(event.target),
@@ -706,10 +707,14 @@ define(function (require, exports) {
                     layer_name = target.data('layer'),
                     file_name  = target.data('file');
                 self.removeFile(file_name);
+                if (self.files.length == 0) {
+                    delete layers[self.name];
+                }
                 if (file_ext === 'xml') {
                     $('#metadata_uploaded_preserve_check').hide();
                 }
-                self.displayRefresh();
+                self.errors = self.collectErrors();
+                self.displayErrors();
             });
         });
     };
--- a/geonode/static/geonode/js/upload/upload.js
+++ b/geonode/static/geonode/js/upload/upload.js
@@ -352,6 +352,10 @@ define(['underscore',
             buildFileInfo(_.groupBy(file_input.files, path.getName));
             displayFiles(file_queue);
         });
+        // Detect click on "Remove" link and update the file_queue
+        $(options.file_queue).on('click', '.remove-file', function () {
+            displayFiles(file_queue);
+        });
         $(options.clear_button).on('click', doClearState);
         $(options.upload_button).on('click', doUploads);
         $("[id^=delete]").on('click', doDelete);


geonode/static/geonode/js/upload/LayerInfo.js
```diff
@@ -706,10 +706,12 @@ define(function (require, exports) {
                     layer_name = target.data('layer'),
                     file_name  = target.data('file');
                 self.removeFile(file_name);
+                if (self.files.length == 0) {
```
travislbrundage: Comment on L709: Detects if the above call to `removeFile` has removed the final file associated with the LayerInfo (length == 0). If so, delete this LayerInfo object from the list of stored LayerInfo objects that have been staged for upload. (Previously this was not being done, trapping the user in an error state unless they clicked "Clear")

geonode/static/geonode/js/upload/LayerInfo.js
```diff
@@ -706,10 +706,12 @@ define(function (require, exports) {
...
+                    delete layers[self.name];
+                }
                 if (file_ext === 'xml') {
                     $('#metadata_uploaded_preserve_check').hide();
                 }
-                self.displayRefresh();
```
travislbrundage: Comment on L712: This process is being handled in upload.js now (L355).

geonode/static/geonode/js/upload/upload.js
```diff
@@ -352,6 +352,9 @@ define(['underscore',
             buildFileInfo(_.groupBy(file_input.files, path.getName));
             displayFiles(file_queue);
         });
+        $(options.file_queue).on('click', 'a', function () {
+            displayFiles(file_queue);
+        });
```
travislbrundage: Comment on L355: Refresh the display of all layers when a file is removed, this way we can handle when a layer has been deleted.

I'm open for suggestions on better ways to handle this case, but I'm not sure it's possible to achieve this without delegating responsibility to upload.js.

Perhaps there is a way to detect a deletion of an object in `layers` and respond to that?

Note to self: This section isn't intuitive, will add a comment in the code explaining above and push shortly.

geonode/static/geonode/js/upload/LayerInfo.js
```diff
@@ -706,10 +706,14 @@ define(function (require, exports) {
...
+                self.errors = self.collectErrors();
+                self.displayErrors();
```
travislbrundage: Displaying files is now being handled in upload.js (L355), so I've extracted only the error collection steps.

geonode/static/geonode/js/upload/upload.js
```diff
@@ -352,6 +352,10 @@ define(['underscore',
             buildFileInfo(_.groupBy(file_input.files, path.getName));
             displayFiles(file_queue);
         });
+        // Detect click on "Remove" link and update the file_queue
+        $(options.file_queue).on('click', 'a', function () {
+            displayFiles(file_queue);
+        });
```
travislbrundage: Refresh the display of all layers when a file is removed, this way we can handle when a layer has been deleted.

I'm open for suggestions on better ways to handle this case, but I'm not sure it's possible to achieve this without delegating responsibility to upload.js.

Perhaps there is a way to detect a deletion of an object in layers and respond to that?

travislbrundage: Refresh the display of all layers when a file is removed, this way we can handle when a layer has been deleted.

I'm open for suggestions on better ways to handle this case, but I'm not sure it's possible to achieve this without delegating responsibility to upload.js.

Perhaps there is a way to detect a deletion of an object in `layers` and respond to that?

pjdufour: Could the jquery event selector use a class too?  Just `a` doesn't provide any semantic meaning.  Hopefully just minor change.

travislbrundage: Yeah, that is a good point. I agree with your thinking. I believe I should just have to modify the generated element to have a class attribute, then I can do that. I'll get that update in tomorrow morning for me, or if it's not possible for some reason I'll post back here with the problem.

Issue #2646, opened 2016-09-06
shapefiles with name longer that 63 chars cannot be loaded


simod:
is a limitation on postgres table name which is max 63 chars

harts-boundless:
This may have been closed by #2671, would someone mind checking and closing if so?

PR #2671, opened 2016-10-17
truncate upload names over length to allow upload

harts-boundless:
One approach to the problem described in
https://github.com/GeoNode/geonode/issues/2646

This patch uses the Django field definition to get the length, so this can be controlled from that one place and won't get out of sync.

### DIFF
--- a/geonode/upload/upload.py
+++ b/geonode/upload/upload.py
@@ -290,6 +290,10 @@ def save_step(user, layer, spatial_files, overwrite=True,
         next_id = max(int(last_importer_session), int(upload_next_id)) + 1
         next_id = max(int(last_importer_session), int(upload_next_id)) + 1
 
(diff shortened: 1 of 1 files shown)


pjdufour:
Looks like good implementation.  Passes automated tests, but have you manually checked a few workflows by manually uploading files of different sizes through interface, etc?  Would want to make sure the GeoServer handshake works.

harts-boundless:
Sorry I lagged getting back on this comment! 

I did manually check with shapefiles of several sizes around the limit (e.g. -1, +0, +1) and it passed that manual inspection, so I can confirm for example that this should not contain an off-by-one error. 

The resulting length of names in many parts of the UI is far short of what is stored in the db, but that's a separate issue (and in many cases it's already so long that things are visibly garbled).

PR #2672, opened 2016-10-18
Fixing missing and broken links to actions on the user's activity page

travislbrundage:
Prior to these applied changes, the "My Activity" link would sometimes incorrectly display the user's actions. They would lack the link to the object sometimes, and sometimes not display at all.

The User activity feed prior to this change:
alt tag

### DIFF
--- a/geonode/social/views.py
+++ b/geonode/social/views.py
@@ -42,3 +42,26 @@ class RecentActivity(ListView):
             public=True,
             action_object_content_type__model='comment')[:15]
         return context
+
+
+class UserActivity(ListView):
+    """
+    Returns recent user activity.
+    """
+    context_object_name = 'action_list'
+    template_name = 'actstream/actor.html'
+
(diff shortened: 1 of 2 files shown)


geonode/social/views.py
```diff
@@ -42,3 +42,26 @@ def get_context_data(self, *args, **kwargs):
...
+
+    def get_queryset(self):
+        # There's no generic foreign key for 'actor', so can't filter directly
+        # Hence the code below is essentially applying the filter afterwards
+        user_actions = []
+        action_list = Action.objects.filter(public=True)[:15]
+        for action in action_list:
+            if action.actor.username == self.kwargs['actor']:
+                user_actions.append(action)
+        return user_actions
```
travislbrundage: Not sure if there is a better way to do this since `actor` can't be filtered by. I'm open to feedback and suggestions.

PR #2703, opened 2016-11-07
Added option to add a "Register" link to the GUI

jondoig:
As advised by Simone Dalmasso here: https://groups.google.com/forum/#!topic/geonode-users/oU9I8vI_Npo


simod:
Hi, we have already some doc on this http://docs.geonode.org/en/master/tutorials/advanced/geonode_settings/settings.html#user-account-settings. Maybe is good to just point to it.

PR #2709, opened 2016-11-08
Update the object_list limit 

Coop56:
Update the object_list limit to use the value assigned to CLIENT_RESULTS_LIMIT in settings rather than a hard coded value.

### DIFF
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -411,7 +411,7 @@ class CommonModelApi(ModelResource):
 
         object_list = {
            "meta": {
-                "limit": settings.CLIENT_RESULTS_LIMIT,
+                "limit": settings.API_LIMIT_PER_PAGE,
                 "next": next_page,
                 "offset": int(getattr(request.GET, 'offset', 0)),
                 "previous": previous_page,
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -410,14 +410,15 @@ class CommonModelApi(ModelResource):
             objects = []
 
         object_list = {
-           "meta": {"limit": settings.CLIENT_RESULTS_LIMIT,
-                    "next": next_page,
-                    "offset": int(getattr(request.GET, 'offset', 0)),
-                    "previous": previous_page,
-                    "total_count": total_count,
-                    "facets": facets,
-                    },
-            'objects': map(lambda x: self.get_haystack_api_fields(x), objects),
+           "meta": {
+                "limit": settings.CLIENT_RESULTS_LIMIT,
+                "next": next_page,
+                "offset": int(getattr(request.GET, 'offset', 0)),
+                "previous": previous_page,
+                "total_count": total_count,
+                "facets": facets,
+            },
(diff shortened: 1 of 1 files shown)


harts-boundless:
I could be wrong, but in the Travis CI log, it looks like this bailed right after flake8 ran. Could be because you removed the `#noqa`. 

When I run flake8 myself, I get this:
```
resourcebase_api.py:420:13: E131 continuation line unaligned for hanging indent
```

Although it's not your fault, and I am just another contributor and do not speak for GeoNode, but can you make this patch so that it keeps flake8 happy without reintroducing the noqa? Example approach

```
:100644 100644 40a04c4... 0000000... M	geonode/api/resourcebase_api.py

diff --git a/geonode/api/resourcebase_api.py b/geonode/api/resourcebase_api.py
index 40a04c4..2a895c1 100644
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -410,13 +410,14 @@ class CommonModelApi(ModelResource):
             objects = []
 
         object_list = {
-           "meta": {"limit": 100,  # noqa
-                    "next": next_page,
-                    "offset": int(getattr(request.GET, 'offset', 0)),
-                    "previous": previous_page,
-                    "total_count": total_count,
-                    "facets": facets,
-                    },
+            "meta": {
+                "limit": 100,
+                "next": next_page,
+                "offset": int(getattr(request.GET, 'offset', 0)),
+                "previous": previous_page,
+                "total_count": total_count,
+                "facets": facets,
+            },
             'objects': map(lambda x: self.get_haystack_api_fields(x), objects),
         }
         self.log_throttled_access(request)
```

Coop56:
I've updated the PR.  I made sure it passed PyCharm's PEP8 inspection.  Is there a specific way we deal with single quotes vs. double quotes in the project?  I see them both used in various places, both are valid Python code but I wasn't sure if Geonode preferred one over the other.

geonode/api/resourcebase_api.py
```diff
@@ -410,14 +410,15 @@ def get_search(self, request, **kwargs):
...
-           "meta": {"limit": settings.CLIENT_RESULTS_LIMIT,
-                    "next": next_page,
-                    "offset": int(getattr(request.GET, 'offset', 0)),
-                    "previous": previous_page,
-                    "total_count": total_count,
-                    "facets": facets,
-                    },
-            'objects': map(lambda x: self.get_haystack_api_fields(x), objects),
+           "meta": {
+                "limit": settings.CLIENT_RESULTS_LIMIT,
```
simod: @Coop56 we have also this one in settings https://github.com/GeoNode/geonode/blob/master/geonode/settings.py#L783 which is for that specific purpose. CLIENT_RESULTS_LIMIT is actually the one that the js will use to limit the number of results returned in the api query.

Coop56: I'm happy to change it to that variable instead - but either way it should be a variable and not a hard coded value.  Just let me know if you want me to change it and I'll update the PR.

Coop56:
@simod I went ahead and updated this PR based on our conversation.

PR #2711, opened 2016-11-10
The standalone TEMPLATE_* settings were deprecated in Django 1.8

BerryDaniel:
-  The standalone TEMPLATE_* settings were deprecated in Django 1.8 and the TEMPLATES dictionary takes precedence. You must put the values of the following settings into your default TEMPLATES dict: TEMPLATE_DIRS, TEMPLATE_CONTEXT_PROCESSORS, TEMPLATE_DEBUG. https://docs.djangoproject.com/en/1.8/topics/templates/#support-for-template-engines
- historic pep8 fixes


ingenieroariel:
Thanks @BerryDaniel!

BTW, In the future it's better to send the pep8 fixes as a separate PR or at least a separate commit.

PR #2719, opened 2016-11-22
Correctly handle xml sidecar files within a zip

travislbrundage:
This is a small change to fix a minor big with the uploader.

The problem exists when we try to upload a zipped shapefile that includes an xml sidecar file. When we ended up in the final_step of the upload process, the xml file would be attempted to be read. However, the code was operating correctly only in the case of the xml being uploaded alongside other individual shapefiles. If they are packaged together in a .zip, it would fail to open the xml file since it hasn't been extracted in this context.

This update will xml files packaged in a zip to be correctly read and handled by the uploader.

### DIFF
--- a/geonode/upload/tests/integration.py
+++ b/geonode/upload/tests/integration.py
@@ -530,6 +530,21 @@ class TestUpload(UploaderBase):
         self.upload_file(abspath, self.complete_upload,
                          check_name='san_andres_y_providencia_poi')
 
+    def test_zipped_upload_xml_sidecar(self):
+        """Test uploading a zipped shapefile with xml sidecar"""
+        fd, abspath = self.temp_file('.zip')
+        fp = os.fdopen(fd, 'wb')
+        zf = ZipFile(fp, 'w')
+        fpath = os.path.join(
+            GOOD_DATA,
+            'vector',
+            'Air_Runways.*')
+        for f in glob.glob(fpath):
+            zf.write(f, os.path.basename(f))
+        zf.close()
+        self.upload_file(abspath, self.complete_upload,
+                         check_name='Air_Runways')
+
     def test_invalid_layer_upload(self):
         """ Tests the layers that are invalid and should not be uploaded"""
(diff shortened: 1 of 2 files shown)


pjdufour:
This is for the xml Metadata right?  Just to double check.

travislbrundage:
Yeah. Metadata contained within the XML.

I'm not particularly well versed in GIS data, so if there's something I'm missing or haven't considered, please let me know. For instance, I don't know if the code should *purposefully* skip that section for any reason... from the looks of it, it seems it'll handle any format for XML metadata. I'm not sure if there's any other reason one would include XML in your upload, other than for metadata.

harts-boundless:
Don't know exactly what's up with the Travis CI build, but by my best guess it seems to be that flake8 is returning with a nonzero error code.

This branch rebased on geonode/master has a lot of flake8 errors (if certain ones are not ignored)

I audited geonode master (not this feature branch) for flake8 errors using python2.7 - for details see https://github.com/GeoNode/geonode/pull/2730

Once this branch was rebased on that, flake8 gave the following error:

`geonode/upload/upload.py:742:18: F821 undefined name 'zipfile'`

I can understand why flake8 doesn't see this, because `zipfile` has a deferred import (not at the top of the file) - even though to my understanding it is part of python stdlib (see e.g. https://docs.python.org/2/library/zipfile.html) and there is another branch which may be using this name when it wasn't imported

I suggest that this PR could be amended to include an additional commit which deletes the 'import zipfile' and moves it to the top of the file. I doubt this would cause anything to break, but things would have to be tested after making this change

jj0hns0n:
@travislbrundage can you check the flake8 issues and fix.

travislbrundage:
I'll take a look and see if I can get the flake8 errors fixed. I thought it was because the line below my edits is too long (which I did not write). I'll see if moving the zipfile import helps.

travislbrundage:
Travis CI was fixed by moving the zipfile import, but coverage checks are now complaining. :)

PR #2740, opened 2016-11-28
changing version number of the documentation

d3netxer:
changing version number of the documentation


sarasafavi:
@d3netxer looks like you were working against an outdated version of this file -- closing as dates have since been fixed, but thanks! Hit me up IRL today if you want more info :)

PR #2751, opened 2016-11-30
Authentication geonode in  geoserver

waybarrios:


francbartoli:
@waybarrios let's wait to merge if we have different solution on the docker orchestration topic. this is associated to [https://github.com/GeoNode/geoserver-docker](https://github.com/GeoNode/geoserver-docker) and [#8](https://github.com/GeoNode/geoserver-docker/pull/8)

PR #2753, opened 2016-11-30
Creating authentication for geonode into geoserver container

waybarrios:
PR contains the following:
-  Organizing env variables on geoserver docker container
-  Overwrite config.xml with django docker ip using env variable. Reminder: By default is `http://django:8000/`it means django (link on docker compose) and the port that is running. 
- Make overwriting while you do `make up` 

Thanks to @BerryDaniel for advising me to organize the PR.


waybarrios:
@ingenieroariel look at this!

PR #2754, opened 2016-12-01
fix #2573

### LINKED ISSUE #2573: Private layers show in auto-complete search
I have a geonode with several hundred layers, only a few dozen are public. All layers show in the auto-complete dropdown in the header, regardless of whether they are private or not.

simod:


harts-boundless:
for ease of reference, https://github.com/GeoNode/geonode/issues/2573 is about private layers showing in the autocomplete dropdown along with public layers.

PR #2757, opened 2016-12-01
add geonode.mp as contrib app

simod:


simod:
todo: add some doc...

PR #2763, opened 2016-12-02
Add typename to search API output, fix layer creation bug.

theduckylittle:
The Typename is required to reference the layer to the appropriate layer in GeoServer when the title has been edited.  This change puts the typename of the layer in the output.

Also found a bug in pycsw_local.py which caused PostgreSQL to try to compare a boolean to an integer. Postgres will not automatically cast 1 to "true" and this causing layer uploads to fail.

### DIFF
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -479,6 +479,7 @@ class CommonModelApi(ModelResource):
             'title',
             'date',
             'abstract',
+            'typename',
             'csw_wkt_geometry',
             'csw_type',
             'owner__username',
--- a/geonode/catalogue/backends/pycsw_local.py
+++ b/geonode/catalogue/backends/pycsw_local.py
@@ -46,7 +46,7 @@ CONFIGURATION = {
     },
     'repository': {
         'source': 'geonode',
-        'filter': 'is_published = 1',
+        'filter': 'is_published = true',
         'mappings': os.path.join(os.path.dirname(__file__), 'pycsw_local_mappings.py')
     }
 }


simod:
i'm afraid that this changes break geonode. the typename is not available in the resourcebase table, though the detail_url contains it and we already used it to get the typename. also che change to the pycsw configuration seems to break the tests.

jj0hns0n:
@tomkralidis can you look at the pycsw portion of this.

tomkralidis:
@theduckylittle thanks for the report.  I think the following would work (can test later today):

```diff
diff --git a/geonode/catalogue/backends/pycsw_local.py b/geonode/catalogue/backends/pycsw_local.py
index 2eb2a6b..157c56d 100644
--- a/geonode/catalogue/backends/pycsw_local.py
+++ b/geonode/catalogue/backends/pycsw_local.py
@@ -28,6 +28,11 @@ from geonode.catalogue.backends.generic import CatalogueBackend as GenericCatalo
 from geonode.catalogue.backends.generic import METADATA_FORMATS
 from shapely.geometry.base import ReadingError
 
+if 'sqlite' in settings.DATABASE_URL:
+    true_value = '1'
+else:
+    true_value = 'true'
+
 # pycsw settings that the user shouldn't have to worry about
 CONFIGURATION = {
     'server': {
@@ -46,7 +51,7 @@ CONFIGURATION = {
     },
     'repository': {
         'source': 'geonode',
-        'filter': 'is_published = 1',
+        'filter': 'is_published = %s' % true_val,
         'mappings': os.path.join(os.path.dirname(__file__), 'pycsw_local_mappings.py')
     }
 }
```

tomkralidis:
Having said this, the [PostgreSQL docs](https://www.postgresql.org/docs/9.1/static/datatype-boolean.html#DATATYPE-BOOLEAN-TABLE) say `1` and `0` are valid literal boolean values.

tomkralidis:
@theduckylittle we can go with the above patch snippet.

theduckylittle:
Thanks @tomkralidis , that should fix my problem.  I just wanted to put this in here for posterity:
```
gis=> select version();
                                                   version                                                    
--------------------------------------------------------------------------------------------------------------
 PostgreSQL 9.3.15 on x86_64-unknown-linux-gnu, compiled by gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4, 64-bit
(1 row)

gis=> create table test_boolean (bool boolean);
CREATE TABLE
gis=> insert into test_boolean values (true);
INSERT 0 1
gis=> insert into test_boolean values (false);
INSERT 0 1
gis=> select count(*) from test_boolean where bool = 1;
ERROR:  operator does not exist: boolean = integer
LINE 1: select count(*) from test_boolean where bool = 1;
                                                     ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
gis=> select count(*) from test_boolean where bool = 1::boolean;
 count 
-------
     1
(1 row)
```

Postgres will accept 0/1 as boolean values but only if they are CAST to the boolean type first.

PR #2767, opened 2016-12-05
Fix for PostgreSQL/SQLite boolean type differences

theduckylittle:
Inspired by the Tom Kradalis code seen here:
 - https://github.com/GeoNode/geonode/pull/2763

### DIFF
--- a/geonode/catalogue/backends/pycsw_local.py
+++ b/geonode/catalogue/backends/pycsw_local.py
@@ -28,6 +28,10 @@ from geonode.catalogue.backends.generic import CatalogueBackend as GenericCatalo
 from geonode.catalogue.backends.generic import METADATA_FORMATS
 from shapely.geometry.base import ReadingError
 
+true_value = 'true'
+if 'sqlite' in settings.DATABASE_URL:
+    true_value = '1'
+
 # pycsw settings that the user shouldn't have to worry about
 CONFIGURATION = {
     'server': {
@@ -46,7 +50,7 @@ CONFIGURATION = {
     },
     'repository': {
         'source': 'geonode',
-        'filter': 'is_published = 1',
+        'filter': 'is_published = %s' % true_value,
         'mappings': os.path.join(os.path.dirname(__file__), 'pycsw_local_mappings.py')
     }
 }


code-geek:
This still breaks when the DATABASE_URL environment variable is not set.

`DATABASE_URL = os.getenv(
    'DATABASE_URL',
    'sqlite:///{path}'.format(
        path=os.path.join(PROJECT_ROOT, 'development.db')
    )
)`

simod:
Can't replicate it, what is breaking exactly?

code-geek:
since the patch is:

`true_value = 'true'
if 'sqlite' in settings.DATABASE_URL:
    true_value = '1'`

and the default for settings.DATABASE_URL is:

`DATABASE_URL = os.getenv( 'DATABASE_URL', 'sqlite:///{path}'.format( path=os.path.join(PROJECT_ROOT, 'development.db') ) )`

which of course contains 'sqlite',

it still sets true_value to '1' and the same error occurs when trying to upload layers or documents. 

image

image

So, it's become mandatory to set the DATABASE_URL environment variable or reset it in local_settings.py for it to work correctly.

simod:
Ah! good catch! @tomkralidis should we check for DATABASE['default']['engine'] or something like this instead?

tomkralidis:
@code-geek can you provide an example so I can reproduce?  i.e. in which scenario is `geonode.settings.DATABASE_URL` not set?

code-geek:
@tomkralidis Here's a snippet from my local_settings.py

```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.postgresql_psycopg2',
        'NAME': 'aaaaaaaaaa',
        'USER': 'aaaaaaaaa',
        'PASSWORD': 'aaaaaaaaaaa',
    }
}
```

In this case I'm not overriding DATABASE_URL, which means the default value for it is used. The default value contains the word 'sqlite', although my implementation uses PostGreSQL. So, true value is set to '1' and the same error occurs.

tomkralidis:
thanks @code-geek can you test the below (I'll commit/push thereafter):

```diff
diff --git a/geonode/catalogue/backends/pycsw_local.py b/geonode/catalogue/backends/pycsw_local.py
index 6f92cc1..c4b2b80 100644
--- a/geonode/catalogue/backends/pycsw_local.py
+++ b/geonode/catalogue/backends/pycsw_local.py
@@ -29,7 +29,7 @@ from geonode.catalogue.backends.generic import METADATA_FORMATS
 from shapely.geometry.base import ReadingError

 true_value = 'true'
-if 'sqlite' in settings.DATABASE_URL:
+if 'sqlite' in settings.DATABASES['default']['ENGINE']:
     true_value = '1'

 # pycsw settings that the user shouldn't have to worry about
```

tomkralidis:
Fixed in master: https://github.com/GeoNode/geonode/commit/0bef27795d637fd75cf24a436eb1bbdeb77ace83 (issue #2771)

PR #2784, opened 2016-12-20
Fix for index.html for dataset name length <h4>

BerryDaniel:
L169 limited featured dataset title to 20 characters so the the name doesnt overlap the next dataset.

### DIFF
--- a/geonode/templates/index.html
+++ b/geonode/templates/index.html
@@ -45,7 +45,7 @@
   			<div class="container text-center">
   				<h1>Discover the available datasets.</h1>
   				<div class="col-xs-4 col-sm-2">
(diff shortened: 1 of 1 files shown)


capooti:
two things:
1. I would use django truncatechars filter in place of the angular limitto filter
2. I would not hardcode the categories (this is not related to your PR, but just an opinion)
thanks

capooti:
forget my first comment, as that is an angular variable not a django one

ppasq:
@capooti about your second comment: the idea was to show a selection of categories or metacategories since the ISOs are 19 (a lot) and some of them have a kind of fuzzy names. Btw I agree that hard coding the categories is bad and naming could be misleading (metacategories vs ISOs).

PR #2785, opened 2016-12-20
removal of corsheader app, since it was not working and still require…

BerryDaniel:
…d a proxy server such as nginx or apache. The reason for removal was to reduce the nonessential app dependencies.


BerryDaniel:
It appears that removing a few lines decreased our coverage. just fyi

PR #2789, opened 2016-12-23
Add a link to the Geogig Repository in the Layers API

theduckylittle:
- create_response now receives bundles and not dicts, fixing
   serialization issues
- corrected id-filtering based on the conversion to bundles.
- added geogig_enabled, geogig_link attributes to layers API.
- moved field filtering to specific APIs as there was layers-specific
   field names in the Base class.
- geogig_link is now published in the layer API results.

### DIFF
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -536,7 +536,7 @@ class LayerResource(CommonModelApi):
 
     """Layer API"""
 
-    geogig_enabled = fields.BooleanField(attribute='geogig_enabled', readonly=True)
+    geogig_link = fields.CharField(attribute='geogig_link', readonly=True, null=True)
 
     class Meta(CommonMetaApi):
         queryset = Layer.objects.distinct().order_by('-date')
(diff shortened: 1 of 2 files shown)


geonode/api/resourcebase_api.py
```diff
@@ -454,6 +454,12 @@ def get_list(self, request, **kwargs):
             collection_name=self._meta.collection_name)
         to_be_serialized = paginator.page()
 
+        bundles = [
```
davisc: @simod thanks for the background information. Were there any metrics captured for a performance baseline?

simod:
Yes there were.... don't remember where maybe @ingenieroariel remembers.

geonode/api/resourcebase_api.py
sarasafavi: @davisc @theduckylittle not the answer you are looking for, but for reference see https://github.com/GeoNode/geonode/pull/1537

theduckylittle:
Closing out in favor of: https://github.com/GeoNode/geonode/pull/2792

PR #2808, opened 2017-01-07
Fix bug with ajax POST and page refresh on inbox page when deleting m…

travislbrundage:
…essage

Previously, the ajax call would "work", but only once, after which the listener got removed and it would POST the delete the old way.

This should fix it to always delete the message using ajax correctly.

### DIFF
--- a/geonode/templates/user_messages/inbox.html
+++ b/geonode/templates/user_messages/inbox.html
@@ -40,17 +40,23 @@
 
 {% block extra_script %}
 <script type="text/javascript">
-$('.message_delete_btn').click(function(event) {
-  form = $(this).parent()[0];
-  $.ajax({
-    type: "POST",
-    url: $(form).attr('action'),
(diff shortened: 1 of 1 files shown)


geonode/templates/user_messages/inbox.html
```diff
@@ -40,17 +40,23 @@ <h2 class="page-title">{%  trans "Messages" %}</h2>
...
-  $.ajax({
-    type: "POST",
-    url: $(form).attr('action'),
-    data: $(form).serialize(),
-    success: function() {
-      $('#inbox').load(window.location.pathname + ' #inbox');
-      $('#all').load(window.location.pathname + ' #all');
-    }
-  });
+$('.tab-content').click(function(event) {
```
travislbrundage: Attaching the listener to the containing element - the bug before occurred because we reloaded the elements that had the listeners attached to them.

PR #2813, opened 2017-01-11
Fixes an upload error generated by not checking that geogig is not enabled

capooti:
Fixes an upload error generated by not checking that geogig is not enabled


travislbrundage:
Hey @capooti , I picked your commit into my PR here: https://github.com/GeoNode/geonode/pull/2812

It also fixes a bug with it when geogig_enabled is true. This should cover all cases - can you check to see if it works for you?

Issue #2815, opened 2017-01-12
OAuth2 "verify_token" api should not throw an error if user is not authenticated


afabiani:
Currently the "verify_token" api throws an exception if there is no authenticated user on the session. This causes an error when trying to acquire token info from an external URI. Moreover this check is not necessary at all, since the authentication process to obtain a valid "access_token" has been already done before.

afabiani:
PR: https://github.com/GeoNode/geonode/pull/2816

PR #2816, opened 2017-01-12
Fix for Issue 2815: OAuth2 verify_token api should not throw an error if user is not authenticated

afabiani:


afabiani:
Related to https://github.com/GeoNode/geonode/issues/2815

PR #2855, opened 2017-01-25
Oauth2 GeoServer/GeoNode implementation on Docker

waybarrios:
Hi All, 

This PR has new authentication on docker ecosystem. This authentication works taking into account the following aspects:
* Overwriting config xml files using [env variables](https://github.com/waybarrios/geonode/tree/docker-oauth/scripts/docker/env/production) with GeoServer and Django Location 
* Loading  [default authenticate configuration](https://github.com/waybarrios/geonode/blob/docker-oauth/geonode/base/fixtures/default_oauth_apps_docker.json) on [Makefile](https://github.com/waybarrios/geonode/blob/docker-oauth/Makefile#L13)
* New geoserver docker image on [alias branch.](https://github.com/GeoNode/geoserver-docker/tree/alias)

**Note:**  You need to edit hosts file on your system using `sudo vim /etc/hosts` and adding `geonode` hostname as localhost:

```
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1       localhost geonode
255.255.255.255 broadcasthost
::1             localhost
```


docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
       - ./scripts/docker/env/production/django.env
 
   geoserver:
-    image: geonode/geoserver
+    image: waybarrios/geoserver:piensa
```
francbartoli: @waybarrios I'd rather if you can provide a PR with an alias branch for the geoserver repo as well in order to have an official alias image from the geonode docker hub. What do you think?

waybarrios: Good idea @francbartoli! I just did it [here](https://github.com/GeoNode/geonode/pull/2855/commits/950af68c985b2b33fc7d3eaae1c8fc52913fda62)

PR #2856, opened 2017-01-25
Format people cards to be universal height so sorting works properly

travislbrundage:
This change solves an issue where sorting looks incorrect in the UI due to formatting errors.

Steps to reproduce visual bug:
1. Create more than 4 users.
2. Give each user an Organization Name, varying their length to emphasize the problem.
3. Navigate to the People search page.
4. Select the filter dropdown, and pick any of them which will cause the tiles to become reorganized.

You should see the profile cards look out of order due to being of varying sizes.

Additionally, another small bug existed where the tiles were supposed to display "No Organization Info" if the organization field had no data. This was also corrected.


jj0hns0n:
Here is a screenshot of what I have been seeing. Not sure if this will take care of that too.

screen shot 2017-01-25 at 18 58 01

PR #2868, opened 2017-01-30
paver start shouldn’t call sync

simod:


francbartoli:
Maybe tests have to be upgraded

PR #2880, opened 2017-02-01
move metadata_uploaded_preserve to the right migration

simod:
thanks @kalxas


kalxas:
Thanks for the fix @simod

Issue #2728, opened 2016-11-28
if "LAYER_PREVIEW_LIBRARY" is set to "leaflet", map detail does not render correctly


martagi:
When `localhost:8000/maps/<mapid>` is loaded, the leaflet panel is empty.
![mapdetailleaflet_console](https://cloud.githubusercontent.com/assets/19388911/20665077/67305134-b55e-11e6-9169-78a905cf9c77.jpg)

If `localhost:8000/layers/<layername>`, the leaflet panel is ok.
![layerdetail_console](https://cloud.githubusercontent.com/assets/19388911/20665184/d5356e12-b55e-11e6-8d0f-ac6aeda23442.jpg)

simod:
seems that the initial map zoom and center are invalid.

fargalaxy1:
I found that in `maps/map_leaflet.html` the map center `({{ resource.center_x }}, {{ resource.center_y }})` is in EPSG:900913 projection while leaflet setview() method uses (lat,lon) coordinates.
I used [proj4js](http://proj4js.org/) to convert the map center in (lat, lon). If I set a "static" tilelayer and add it to the map, the leaflet panel correctly shows up the expected area. If I use tilelayers created from geonode layers, the panel is empty. I'll go on on this, if you have suggestions..

simod:
not really suggestions apart of some debug that need to be done in the js generated by django. Thanks @fargalaxy1, if you can come up with a PR then, we'll appreciate :)

PR #2898, opened 2017-02-09
Fix issue with Leaflet library

fargalaxy1:
Fix to issue #2728 I opened past November.
When "LAYER_PREVIEW_LIBRARY" was set to "leaflet", the map detail was not rendering in the correct way. There where a number of problems with the previous implementation:

1. Base layers where not correctly created. I solved by using the base layers provided [here](https://github.com/GeoNode/geonode/blob/master/geonode/layers/templates/layers/layer_leaflet_map.html#L13);
2. the coordinates of the map center (`{{ resource.center_x }}`,` {{ resource.center_y }}`) where in google mercator, now they're converted in WSG84 by using [proj4js](http://proj4js.org/) library;
3. in the loop over the user's layers, the wrong Leaflet function was used to instantiate a WMS tile layer object given the URL to the WMS;
4. add a LayerControl object correctly addressing the different base layers and switching overlays on/off.

### DIFF
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -932,7 +932,8 @@ CACHES = {
     #     }
 }
 
-LAYER_PREVIEW_LIBRARY = 'geoext'
+# LAYER_PREVIEW_LIBRARY = 'geoext'
+LAYER_PREVIEW_LIBRARY = 'leaflet'
 
 SERVICE_UPDATE_INTERVAL = 0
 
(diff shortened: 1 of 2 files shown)


geonode/settings.py
```diff
@@ -932,7 +932,8 @@
     #     }
 }
 
-LAYER_PREVIEW_LIBRARY = 'geoext'
+# LAYER_PREVIEW_LIBRARY = 'geoext'
+LAYER_PREVIEW_LIBRARY = 'leaflet'
```
jj0hns0n: Please leave this with `geoext` as the default and we can merge this.

fargalaxy1:
ok, done

Issue #2896, opened 2017-02-09
Private data is publicly visible at /uploaded/layers


jondoig:
In Geonode 2.4, all uploaded data can be listed and downloaded from `http://<host>/uploaded/layers` regardless of security permissions.

This seems to be by design. The [installation doco](http://docs.geonode.org/en/master/tutorials/install_and_admin/geonode_install/setup_configure_httpd.html) says to make it all wide open:

```
sudo chmod -Rf 777 /home/geonode/geonode/geonode/uploaded/thumbs
sudo chmod -Rf 777 /home/geonode/geonode/geonode/uploaded/layers
```

Removing ‘other’ permission (chmod 770) breaks the upload function.

jondoig:
At least we can make it a bit harder by denying directory listing in the Apache config. I.e. change the [installation doco](http://docs.geonode.org/en/master/tutorials/install_and_admin/geonode_install/setup_configure_httpd.html#apache-configuration) from:

```
    <Directory "/home/geonode/geonode/geonode/uploaded/layers/">
        Order allow,deny
        Options Indexes FollowSymLinks
        Allow from all
        Require all granted
        IndexOptions FancyIndexing
    </Directory>
```
to
```
    <Directory "/home/geonode/geonode/geonode/uploaded/layers/">
        Order allow,deny
        Options -Indexes +FollowSymLinks
        Allow from all
        Require all granted
    </Directory>
```

(...and same for the other directories configured there.)

But it's still not hard to infer the actual filenames and download them without listing the directory.

simod:
Hi, the uploaded/layers directory doesn't have to be served by apache, it only needs write permissions. Thumbs on the other hands need to be publicly available. If the doc says that uploaded has to be served then we need to fix it.

Thanks

jondoig:
OK so I've made a [pull request](https://github.com/GeoNode/geonode/issues/2899) to fix the [manual install doco](http://docs.geonode.org/en/master/tutorials/install_and_admin/geonode_install/setup_configure_httpd.html#apache-configuration), but the same change may be needed to other install methods ...Quick install? Ansible? Other?

simod:
Thanks I've also  fixed it at package level see https://github.com/GeoNode/geonode/commit/dedd939909104b88af4af97b744b334896c819a1

PR #2899, opened 2017-02-09
Removed uploaded/layers from Apache config

jondoig:
"The uploaded/layers directory doesn't have to be served by apache, it only needs write permissions"
- @simod on #2896

https://github.com/GeoNode/geonode/issues/2896#issuecomment-278587461


jj0hns0n:
@jondoig thanks for this and thanks for letting everyone know about this vulnerability.

PR #2904, opened 2017-02-13
Specifically deny access to uploaded/layers

jondoig:
To match change made to package at https://github.com/GeoNode/geonode/commit/dedd939

### DIFF
--- a/docs/tutorials/install_and_admin/geonode_install/setup_configure_httpd.txt
+++ b/docs/tutorials/install_and_admin/geonode_install/setup_configure_httpd.txt
@@ -66,6 +66,11 @@ Place the following content inside the file:::
             IndexOptions FancyIndexing
         </Directory>
 
+        <Directory "/var/www/geonode/uploaded/layers/">
+            Order allow,deny
+            Deny from all
+        </Directory>
+
         <Proxy *>
             Order allow,deny
             Allow from all


simod:
Hi, I'm not sure that this is correct for this case. This setup does not use the /var/www directory but the /home /geonode, moreover the allow is specified for /home/geonode/geonode/geonode/uploaded/thumbs and not for /home/geonode/geonode/geonode/uploaded/layers so this should be already ok. Are you experiencing anything different with this setup?

jondoig:
Sorry about the path error (copy/pasted). Thanks for picking that up.

My previous doco pull request just dropped uploaded/layers from the config,
which worked for me.

This one specifically denies
uploaded/layers. I'm not sure why that's needed. I just added it to make it
consistent with your change to the package.


On 14 Feb 2017 5:25 PM, "Simone Dalmasso" <notifications@github.com> wrote:

Hi, I'm not sure that this is correct for this case. This setup does not
use the /var/www directory but the /home /geonode, moreover the allow is
specified for /home/geonode/geonode/geonode/uploaded/thumbs and not for
/home/geonode/geonode/geonode/uploaded/layers so this should be already ok.
Are you experiencing anything different with this setup?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/GeoNode/geonode/pull/2904#issuecomment-279619228>, or mute
the thread

simod:
Yeah, the setup is a little bit different though. I think that now the doc is correct without the need of this. Or you tested that is really necessary?

jondoig:
Nope my test showed it *wasn't* necessary in my setup (which is based on
that doco).

So feel free to bin this PR.

simod:
ok, thanks, so we can close this?

jondoig:
Yes, close.

Issue #2910, opened 2017-02-15
Allow contrib apps to extend "site base"


afabiani:
As a developer I need GeoNode "contrib" apps to be able to extend and override (if needed) the GeoNode "site_base" template.

Currently this is not possible since "site_base.html" takes precedence on every other template of GeoNode apps.

The solution I propose to solve this issue is to create a new "base/geonode_base.html" basically empty and make all other apps extend this one instead of "site_base.html".

Other than this the order of "INSTALLED_APPS" must be revised in order to allow contribs to be scanned before the other apps.

In this way and the automatic improved templating system of Django 1.8+, it is possible to override / extend the site_base.html template from a contrib app.

PR #2912, opened 2017-02-15
PR for ISSUE #2910 - Allow contrib apps to extend site base

afabiani:
Please refer to ISSUE #2910 

> As a developer I need GeoNode "contrib" apps to be able to extend and override (if needed) the GeoNode "site_base" template.
> 
> Currently this is not possible since "site_base.html" takes precedence on every other template of GeoNode apps.
> 
> The solution I propose to solve this issue is to create a new "base/geonode_base.html" basically empty and make all other apps extend this one instead of "site_base.html".
> 
> Other than this the order of "INSTALLED_APPS" must be revised in order to allow contribs to be scanned before the other apps.
> 
> In this way and the automatic improved templating system of Django 1.8+, it is possible to override / extend the site_base.html template from a contrib app.


simod:
@afabiani would it be possible to complete wipe out site_base.html and just user geonode_base.html which extends from base this to not have three level inheritance? geonode-project will need to be updated accordingly

francbartoli:
@afabiani in the geonode-project I added in the past the [overextends](https://github.com/stephenmcd/django-overextends) app for managing circular templating inheritance. Have you already investigated if this can solve partially your cases?

afabiani:
@simod it is possible to avoid inheritance from site_base.html but unfortunately is not possible to get rid of it; DJango complains if no "site_base.html" file is available on fileSystem.

Good catch with overextends app, I'm going to update the PR with that one. It looks very helpful for this use case.

I will take a look at the geonode-project too if needed. Updating it is necessary only if we wipe out the "site-base.html" though, since this is a different use case and "site-base" and "geonode-base" are basically the same thing.

afabiani:
@simod altough the overextends extension is very useful for our use case to merge more contribs and better manage templates extensions, unfortunately it doesn't solve the issue. Directly extending (or overextending) the site_base.html just ignores others app templates.

afabiani:
@simod fixes committed. Going to modify the geonode-project.

Issue #2921, opened 2017-02-23
Error: Module: 'earth' not found!


gonrial:
Visiting the demo site: http://demo.geonode.org/maps/272
Click on _3d View_
See the error
```
https://www.google.com/jsapi?autoload=%7B%22modules%22%3A%5B%7B%22name%22%3A%22earth%22%2C%22version%22%3A%221%22%2C%22callback%22%3A%22gxp.plugins.GoogleEarth.loader.onScriptLoad%22%7D%5D%7D
```
I found related information from http://stackoverflow.com/questions/42007768/google-maps-api-uncaught-error-module-earth-not-found

Is it possible to disable _3d view_?

gonrial:
Possibly adding the line `config.viewerTools.splice(10,1);` to all these branches?

https://github.com/GeoNode/geonode/blob/2.0.x/geonode/templates/geonode/geo_header.html#L24
https://github.com/GeoNode/geonode/blob/2.4.x/geonode/templates/geonode/geo_header.html#L24
https://github.com/GeoNode/geonode/blob/2.5.x/geonode/templates/geonode/geo_header.html#L24
https://github.com/GeoNode/geonode/blob/2.6.x/geonode/templates/geonode/geo_header.html#L24
https://github.com/GeoNode/geonode/blob/master/geonode/templates/geonode/geo_header.html#L24

simod:
@bartvde any comments on this one?

bartvde:
yeah I would take out the tool

Issue #2892, opened 2017-02-06
404 for geoexplorer link in ext-map dialog


t-book:
Hi,

unsure if this is already documented. When you click the about icon on map details a dialog appears.  Try clicking the geoExplorer tab and you´ll get a 404.

Link: http://demo.geonode.org/maps/169 (happens in all demo maps)
![bildschirmfoto 2017-02-06 um 21 02 57](https://cloud.githubusercontent.com/assets/20478652/22664549/79ca457c-ecb1-11e6-8e57-b2e52af98331.png)

The problem is a wrong property in: http://demo.geonode.org/static/geoexplorer/script/GeoExplorer.js?v=2.4.1
here:

```
    contactText: "Contact",
    aboutThisMapText: "About this Map",
    aboutUrl: "../about.html", /* wrong */
    mapPanel: null,
    constructor: function(a) {
```

should read:

```
    contactText: "Contact",
    aboutThisMapText: "About this Map",
    aboutUrl: "../about", /* fixed */
    mapPanel: null,
    constructor: function(a) {
```
I would create a pull request but as this seems to come from geoExplorer repository I´m unsure where to fix it.

PR #2929, opened 2017-02-24
Multiple changes in Geoexplorer

gonrial:
Disable 3D viewer #2921
Change Geoexplorer aboutUrl #2892


simod:
thanks for this, I'm merging it, though we will not do any 2.0 release, for future commits is enough to do on 2.6.x and master.

PR #2931, opened 2017-02-24
Add style to comment section and Page Detail About section

ppasq:
screen shot 2017-02-24 at 12 54 26


simod:
thanks @ppasq , is this worth to have in master to?

ppasq:
Yes it is. Layout fixes can be added in master as well.

PR #2936, opened 2017-02-28
Fix setup

glennvorhes:
update to Paver version in requirements.txt, Force add project path in pavement.py so it can find geonode.settings


simod:
Hi, what is the issue you find?

glennvorhes:
Dev environment is Ubuntu 16.04.  The issues are given in the commit descriptions.  

On running paver start, geonode.settings was not found as the project root is not in the python path.  Fixed by extending sys.path

pip install -r requirements.txt   
paver installation fails with an import error with version 1.2.1 (given in requirements.txt).  when changed to 1.2.4 (current) the problem was resolved

simod:
I tested this morning on 16.04, the paver version is definitely relevant, could you please add it to setup.py too? For the path in pavement.py, I cannot reproduce, it works fine, can you check your setup?

glennvorhes:
Paver version update has been done in setup.py

Regarding the sys.path issue.  I did some checking and would really like to see the 
sys.path.extend([os.path.dirname(__file__)])
line added to pavement.py

I followed the instructions in the readme and created the virtualenv and activated it.  When running 
paver start
from the project root, I see that the first item in sys.path is the bin directory of the virtualenv and the project directory is not listed unless it is explicitly included.  

As a check I copied paver from the bin directory into the project and ran.  
./paver start
and this worked without the sys.path.extend line in pavement.py as the first entry in sys,path was the project root.

In short the sys.path.extend line to explicitly add the project root seems to be an easy way to ensure that geonode.settings can be imported.

francbartoli:
@glennvorhes I used paver commands regularly, what is this issue and the command which isn't working well?

glennvorhes:
paver setup and sync worked fine.  It wasn't until running paver start that I had the problem described.  

looking into it a little further, it's only the start_geoserver task that has the line
`from geonode.settings import OGC_SERVER`
which is where the import error is raised
Maybe it's before this line that would be the more appropriate place to extend the path

I thought it might be a virtualenv problem and tried deactivating and running
/path/to/myenv/bin/paver start 
but still had the same issue without explicitly adding the path in pavement.py

Also tried installing paver to the default python2.7 interpreter with paver available in the $PATH but still didn't have any luck.  

It always seems to be running from the directory of the pavement executable.  

Thanks for working with me on this.  I think I've tried it a lot of different ways but haven't succeeded in any way other than the workaround in this PR.

simod:
It seems to me likely dependent to your setup, because I cannot replicate it.
And since this is only for development I think for you is just ok to add that line locally.
Could you please remove it from this PR so that we can merge the paver version contribution?

Thanks

pjdufour:
Might not be it, but be sure to install geonode into the python venv as `editable` during development.

glennvorhes:
The last commit removes the line in question.  

Still not sure why I was having the issue and the goal was to prevent a similar problem for others. I'm getting geoserver started as a service and won't be invoking the paver tasks anymore so for me it's no longer a problem.  

In short, this PR is now just for updating the paver version in requirement.txt and setup.py

simod:
Thanks!

Issue #2937, opened 2017-03-01
extra_script block is overwritten when accessing maps or layers


gonrial:
**2.4.x**: Found several `{% block extra_script %}` without `{% block.super %}` so every template overwrites the previous values. 
For example if analytics code is put in `site_base.html` it will not be called when the following templates are used:

* geonode/contrib/favorite/templates/favorite/favorite_list.html
* geonode/documents/templates/documents/document_detail.html
* geonode/documents/templates/documents/document_list.html
* geonode/documents/templates/documents/document_metadata.html
* geonode/documents/templates/documents/document_upload.html
* geonode/groups/templates/groups/group_detail.html
* geonode/groups/templates/groups/group_list.html
* geonode/groups/templates/groups/group_members.html
* geonode/layers/templates/layers/layer_detail.html
* geonode/layers/templates/layers/layer_list.html
* geonode/layers/templates/layers/layer_metadata.html
* geonode/layers/templates/layers/layer_replace.html
* geonode/layers/templates/layers/layer_style_manage.html
* geonode/layers/templates/upload/layer_upload.html
* geonode/layers/templates/upload/layer_upload_csv.html
* geonode/layers/templates/upload/layer_upload_time.html
* geonode/maps/templates/maps/map_detail.html
* geonode/maps/templates/maps/map_download.html
* geonode/maps/templates/maps/map_list.html
* geonode/maps/templates/maps/map_metadata.html
* geonode/people/templates/people/profile_detail.html
* geonode/people/templates/people/profile_list.html
* geonode/services/templates/services/service_register.html
* geonode/templates/account/signup.html
* geonode/templates/index.html

PR #2938, opened 2017-03-01
Added `{{ block.super}}` to all `{% block extra_script %}`

gonrial:
As these templates extended from base if you want to add anaylics the script was not
included in all the site

Solves: https://github.com/GeoNode/geonode/issues/2937

### DIFF
--- a/geonode/contrib/favorite/templates/favorite/favorite_list.html
+++ b/geonode/contrib/favorite/templates/favorite/favorite_list.html
@@ -32,6 +32,7 @@
 {% endblock %}
 
 {% block extra_script %}
+{{ block.super }}
 <script type="text/javascript">
     // add click behavior to the Delete Favorite button.
     // ajax call to delete favorite and remove row on success.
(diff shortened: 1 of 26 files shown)


simod:
Hi, I see the point but what about if in the site_base.html you put the analytics just outside the extra_script block?

gonrial:
I've checked and it's not showing. As `site_base.html` extends `base.html` all the code outside the block sections is not rendered.

simod:
Sorry, I meant in base.html. site_base is just a mute page

gonrial:
Yes I could do it, but the idea is not to copy a hole file just to add a simple script. Next release and I need to rewrite the hole theme.

simod:
I can't see the difference though, even with "super" you will have to add your analytic script in the base.html inside the extra_script block no?
Sorry I'm asking this clarification because if the use case is correct then it is worth having this also in master. Thanks

simod:
makes sense. Could you do a similar one on 2.6.x and master?

gonrial:
2.5.x??

PR #2939, opened 2017-03-02
Added {{block.super}} to all {% block extra_script %}

gonrial:
As these templates extended from base if you want to add anaylics the script was not
included in all the site

Solves: #2937

Also replaced repeated code for facebook buttons with an include to facebook_sdk

### DIFF
--- a/geonode/contrib/favorite/templates/favorite/favorite_list.html
+++ b/geonode/contrib/favorite/templates/favorite/favorite_list.html
@@ -32,6 +32,7 @@
 {% endblock %}
 
 {% block extra_script %}
+{{ block.super }}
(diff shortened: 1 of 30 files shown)


pjdufour:
I think by convention `*_extra blocks` shouldn't need to `super` since they are simply empty placeholders.  Please don't be afraid to add new placeholder blocks to the community templates as needed (as long as mostly backward compatible and proportionate).

gonrial:
@pjdufour I've just applied the same solution previously used for `extra_head`, which is used to add and extend styles. If we need to revert this, we will need to find a broader solution in both fronts (css/js)

PR #2943, opened 2017-03-09
Remove 3d viewer tool from GeoExplorer 

Gpetrak:
Merging the changes from the pull request: "Remove the 3D viewer tool from maps #2925" in my GeoNode instance, I was thinking if 3D viewer button should be removed from GeoExplorer (maps/new) too.


Gpetrak:
I'm sorry for all these changes in my commits. I didn't know that git tracked all my updates in the branch. The changes are just four lines of code in each of these two files: geo_header.html, geo_header_debug.html.

ingenieroariel:
This makes sense to me, is anyone against getting rid of that button?

capooti:
Sounds good to me. Though better way would be to remove it from GeoExplorer itself rather than adding (obscure) javascript code to the GeoNode page for this purpose

PR #2955, opened 2017-03-16
Upgrading FontAwesome to 4.7 and packaging to be served through GeoNode

travislbrundage:
This allows FontAwesome font to be served statically through GeoNode, and no longer requires reaching outward to the Internet to load the font.


pjdufour:
I'm +1.  Any significant changes to the fonts, or is it mostly backwards compatible?

travislbrundage:
I'll look into it to see if there's any significant changes, but just from quick smoke testing everything looked the same. I'm 99% sure FontAwesome only adds to, and does not remove functionality or icons.

ppasq:
Ok for the version upgrade but FontAwesome is already served locally. Please see https://github.com/GeoNode/geonode/blob/master/geonode/static/geonode/css/base.css#L7
And why dowloading also the scss and less files? Is there a need to edit a font?

afabiani:
@travislbrundage @ppasq @pjdufour please check the reviewed version of this PR here

https://github.com/GeoNode/geonode/pull/3052

PR #2959, opened 2017-03-20
Handle OS X zipped files in upload

travislbrundage:
OS X adds a "__MACOSX" directory when creating a zipped file. We allow for uploading of zip files in importer, but it will trip up in the `_contains_bad_names` function, because it recognizes this directory as an xml unsafe name. If we skip this, it'll give an index out of bounds error later, presumably because there's extra copies of the files (I'm not actually 100% sure why).

The goal of this commit is to just clean the directory, so I added a function which just strips anything in a "__MACOSX" directory.

### DIFF
--- a/geonode/upload/files.py
+++ b/geonode/upload/files.py
@@ -207,6 +207,13 @@ def _find_file_type(file_names, extension):
     return filter(lambda f: f.lower().endswith(extension), file_names)
 
 
+def clean_macosx_dir(file_names):
+    """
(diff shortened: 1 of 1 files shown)


capooti:
How about checking the extracted files vs a list containing all of the acceptable files extensions? Something like ['.shp', '.shx', '.prj', '.xml'.....]
This way we can also prevent cases when users upload a zipped file containing invalid stuff.

PR #2966, opened 2017-03-23
Database shards contrib module initial implementation

capooti:
A contrib application for using shard PostgreSQL databases and GeoServer stores in the GeoNode instance.
Documentation attached.
Tests are not available as it is entirely based on PostgreSQL.


capooti:
If accepted I will send this PR also vs master

simod:
Thanks Paolo, this has some changes to the core code too, I would merge it on master an not 2.6. Others opinion?

capooti:
I understand the point. Unfortunately I need to integrate several contrib modules vs 2.6.x. I cannot use master because my deadline is in 3/4 months and I think it is safer to go with 2.6.x.
For now I will keep a fork of 2.6.x, and after my deadline I will start merging the contrib modules to master.

Issue #2958, opened 2017-03-17
importlayers / additional options


pjdufour:
- [x] name
- [x] abstract
- [x] license

PR #2976, opened 2017-03-27
importlayers improvements 

pjdufour:
- added name, title, and abstract to importlayers
-  fixes to importlayers --overwrite
-  catch errors in social signals

#2958


jj0hns0n:
Might be good to add some tests for these things, if only to keep coveralls on our good side.

afabiani:
Review and fixed here
https://github.com/GeoNode/geonode/pull/3044

PR #2982, opened 2017-03-29
keywords and regions default should be list and not tuple

simod:


simod:
I did in master already some time ago. But forgot 2.6 :)

PR #3001, opened 2017-04-12
parse separated allowed hosts

glennvorhes:
The ALLOWED_HOSTS setting needs to be a list but
there is no way to make it so just by using os.getenv

Added logic to check if the environment variable exits
and if so, split on comma. colon, or semicolon excluding
leading or trailing spaces

### DIFF
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -20,6 +20,7 @@
 
 # Django settings for the GeoNode project.
 import os
+import re
 
 from kombu import Queue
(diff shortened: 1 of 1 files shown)


lucernae:
I don't if this will help. I encountered the same problem and I solved it using ast package:

```
import ast
allowed_host_string = "['localhost', 'django']"
ALLOWED_HOSTS = ast.literal_eval(allowed_host_string)
```

PR #3002, opened 2017-04-12
catch and fix import error

glennvorhes:
Still having a problem with this and if I am, hopefully this quick
check and fix will prevent it for other users.

Begin commit message

revisiting a previous issue.  The problem has been isolated and
in contrast with a previous PR, the problem statement and
fix in a try catch block has been isolated to where the problem
occurs as opposed to a top level, beginning of file  sys.path.append

The ALLOWED_HOSTS setting needs to be a list but
there is no way to make it so just by using os.getenv

Added logic to check if the environment variable exits
and if so, split on comma. colon, or semicolon excluding
leading or trailing spaces

### DIFF
--- a/pavement.py
+++ b/pavement.py
@@ -360,7 +360,12 @@ def start_geoserver(options):
     Start GeoServer with GeoNode extensions
     """
 
-    from geonode.settings import OGC_SERVER
+    try:
+        from geonode.settings import OGC_SERVER
+    except ImportError:
+        sys.path.extend([os.path.dirname(__file__)])
+        from geonode.settings import OGC_SERVER
+
     GEOSERVER_BASE_URL = OGC_SERVER['default']['LOCATION']
     url = GEOSERVER_BASE_URL
 


pjdufour:
It's not clear how your commit message and actual changes are related.  If you are running into issues with importing `geonode.*` files, be sure to run `pip install -e .` on the root geonode directory.  Could you clarify your issue and solution?  Thank you.

simod:
I agree with @pjdufour this seems to be a setup issue.

glennvorhes:
Thanks to @pjdufour for the 
`pip install -e`
hint.  I wasn't aware of that option before.

I had thought I followed the instructions at the [Install GeoNode for Development](http://docs.geonode.org/en/master/tutorials/devel/devel_env/index.html) and now see the mention
of `pip install -e geonode`   The documentation has either changed in the past month or so or I missed it the first time.

In any case, I thank you and am closing this PR.

simod:
Thanks. No the doc hasn't changed :)

PR #3006, opened 2017-04-20
Fix too many geoserver warning on travis

lucernae:
Can someone review this?
I had problem with too many geoserver error/warn message in Travis in the current 2.6.x branch.

* Add rabbitmq service by switching to VM based travis

There were too many connection errors related with notification
services in GeoServer, because it can't found rabbitmq. Spin up
rabbitmq instances to handle this.

* Avoid posting duplicate geofence rules for anonymous

There were too many warning on duplicate geofence rule for anonymous
layer. Do some checks if a certain layer already has rule, do not
attempt to post the same rule.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+
 language: python
 
 python:
@@ -5,6 +7,11 @@ python:
 
 virtualenv:
   system_site_packages: true
+
+services:
+  - docker
+  # GeoServer used Rabbitmq for notifier services
+  - rabbitmq
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


simod:
Hi, this PR should be against master branch. For the geofence rules I'm wondering if it should be solved on the geoserver side

lucernae:
I see, thank you Simone. I think we can close this.
One other question @simod, I also want to make another PR for moving qgis-server backend django app to geonode. Should I make PR to 2.6.x or master? I thought 2.6.x is the main develop branch.

simod:
actually master is the dev branch. 2.6.x is the "frozen" branch for the 2.6 release and is the default because master is not easily usable now so people pulling from github get a fully working branch.

Issue #2975, opened 2017-03-27
Upload Layer Issue


pjdufour:
I'm getting the following issue when uploading a shapefile using webui and importlayers.  Django will create a layer object, but a layer won't be created in GeoServer.  Looks like a geofence sync issue.

```
2017-03-27 20:34:45.050:WARN:oejs.ServletHandler:/geoserver/geofence/rest/rules
org.geoserver.geofence.core.dao.DuplicateKeyException: Duplicate Rule Rule[id:null pri:0 ws:geonode l:nepal_tracks acc:ALLOW]
	at org.geoserver.geofence.core.dao.impl.RuleDAOImpl.persist(RuleDAOImpl.java:55)
	at org.geoserver.geofence.core.dao.impl.RuleDAOImpl.persist(RuleDAOImpl.java:31)
	at sun.reflect.GeneratedMethodAccessor292.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:302)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:190)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:157)
	at org.springframework.transaction.interceptor.TransactionInterceptor$1.proceedWithInvocation(TransactionInterceptor.java:99)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:281)
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:96)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:208)
	at com.sun.proxy.$Proxy58.persist(Unknown Source)
	at org.geoserver.geofence.services.RuleAdminServiceImpl.insert(RuleAdminServiceImpl.java:63)
	at org.geoserver.geofence.rest.RulesRestController.insert(RulesRestController.java:134)
[...]

PR #3032, opened 2017-05-05
Fix for ISSUE #2975 - Upload Layer Issue #2975

afabiani:


geonode/settings.py
```diff
@@ -316,7 +316,9 @@
     'autocomplete_light',
     'mptt',
     # 'modeltranslation',
+    # 'djkombu',
     'djcelery',
+    'kombu.transport.django',
```
simod: @afabiani do we need kombu enabled always, also when no celery is used?

afabiani:
Yep, absolutely.

geonode/settings.py
afabiani: Hi @simod, no, I don't think so. Let me double check.

afabiani: @simod dependencies removed. Thanks for the review.

afabiani:
@simod kombu deps removed. Can you maybe review and merge the PR?

geonode/security/models.py
```diff
@@ -17,6 +17,10 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 #########################################################################
+try:
```
simod: very small comment, I think that with the django version we have and 2.7 as required python version, this could be opnly import json.

PR #3036, opened 2017-05-10
 - Fix for Issue #3035: Upload Layer: the check on default style is w…

afabiani:
…eak. Use geometry type instead


kalxas:
Thank you @afabiani 
This looks good.

PR #3041, opened 2017-05-11
 - Migrate Base URL Mgmt Command: Allows to easily update Maps, Style…

afabiani:
…s, Layers and Links


kalxas:
Very useful command, thanks @afabiani

kalxas:
I am thinking if we should backport to 2.6.x, it would be nice to have for the stable release... :)

simod:
+1 this is good and harmless.

PR #3060, opened 2017-05-15
 - Management Command : fixgeofencerules

afabiani:


capooti:
@afabiani, I am still not familiar with the authorization with GeoFence but: would make sense to have the code that sync django-guardian with GeoFence usable also from a celery task? Because I see this as a periodic task

PR #3064, opened 2017-05-16
Fix for #3062: Header becomes to high if username is too long

mekanix:


mekanix:
The value of "20" for truncating is what gives best results on small laptops (I use 12" laptop), just so it doesn't look like a magic number.

PR #3076, opened 2017-05-19
improve dockerization of geonode

Vampouille:
This PR add database bootstrap process in django image. At startup, django image check if a 'geonode' database exists, if it does not exists:
* it will create a new database,
* populate it with a new geonode database and
* load admin account and sample data (if IGNORE_DATA var is not set to TRUE)

Work is in progress on this PR, so please do not merge as is. I create this PR to start discuss on this modification. (At least commit message should be changed !)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       - rabbitmq_data:/var/lib/rabbitmq
 
   django:
-    image: geonode/django
+    image: geonode/geonode
     links:
       - database
       - elasticsearch
@@ -50,7 +50,7 @@ services:
       - geonode_config:/mnt/geonode_config
 
   celery:
-    image: geonode/django
+    image: geonode/geonode
     links:
       - rabbitmq
       - database


Dockerfile
```diff
@@ -1,8 +1,57 @@
...
+WORKDIR /usr/src/app
+COPY . /usr/src/app/
+RUN pip install --no-cache-dir -r requirements.txt
+RUN pip install --no-cache-dir --no-deps -e /usr/src/app/
+
+# Install geonode configuration in named volume
+VOLUME ["/mnt/geonode_config", "/mnt/geonode_data"]
+RUN mkdir -p /mnt/geonode_data/uploaded /mnt/geonode_data/static /mnt/geonode_config \
+  && cd /usr/src/app/geonode/ \
+  && mv local_settings.py.docker /mnt/geonode_config/ \
```
fvanderbiest: `mv local_settings.py.docker /mnt/geonode_config/local_settings.py` ?

docker-compose.yml
```diff
@@ -1,53 +1,83 @@
 version: '2'
 services:
 
-  postgres:
-    image: postgres
+  database:
+    image: camptocamp/postgis:9.5
```
fvanderbiest: I'd rather stick to official image, since postgis is not required here to my knowledge.

package/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+DIR=/docker-entrypoint.d
+
+if [[ -d "$DIR" ]]
+then
+  /bin/run-parts -v --regex '\.(sh|rb)$' "$DIR"
```
fvanderbiest: `.rb` not required here

docker-compose.yml
gamesbook: The official install guide has postgis as a dependency (http://docs.geonode.org/en/master/tutorials/admin/install/custom_install.html) - is this different?

docker-compose.yml
```diff
@@ -1,53 +1,83 @@
...
 
   django:
     image: geonode/django
     links:
-      - postgres
+      - database
       - elasticsearch
       - rabbitmq
     command: django-admin.py runserver 0.0.0.0:8000 --settings=geonode.settings
-    env_file:
```
francbartoli: I'd rather leave env file and keep configuration for production and development separated

docker-compose.yml
```diff
@@ -1,53 +1,83 @@
...
+      - DB_PORT=5432
+      - DB_ADMIN_USER=postgres
+      - DB_ADMIN_PASSWORD=secret
+      - DB_GEONODE_NAME=geonode
+      - DB_GEONODE_USER=geonode
+      - DB_GEONODE_PASSWORD=secret
+      - DB_GEOSERVER_NAME=geoserver
+      - DB_GEOSERVER_USER=geoserver
+      - DB_GEOSERVER_PASSWORD=secret
+    volumes:
```
francbartoli: I'd like if you follow the pattern of named volumes and volumes_from directive

Dockerfile
```diff
@@ -1,8 +1,57 @@
-FROM geonode/django
-MAINTAINER GeoNode development team
+FROM python:2.7.9
```
francbartoli: We should maintain it here, so please update the dockerfile from the geonode/django repo

fvanderbiest: > We should maintain it here

What are the advantages of doing so ?
In our opinion, maintaining this override complicates image lisibility.

docker-compose.yml
fvanderbiest: You mean, having a dedicated container for named volume access ?

docker-compose.yml
francbartoli: GeoNode has two different databases one for configuration data and another for spatial data from postgis tables as datasource for geoserver

docker-compose.yml
francbartoli: Yep as well as we did for geoserver data directory

Dockerfile
francbartoli: The main reason here is that geonode can be easily customizable with a geonode-project which is a separated django application on top of that.

Can you describe in detail what the complications would be?

docker-compose.yml
fvanderbiest: If I understand correctly, you mean it's OK to keep one postgis container for both databases ?

docker-compose.yml
fvanderbiest: `volumes_from` has been deprecated in docker-compose v3. We'd rather use a dedicated container with openssh-server to access volume contents.
We'll come up with a proposal.

docker-compose.yml
francbartoli: as long as we are going to create two different databases as above which are then configured in the local_settings.py. This should be valid only for production environment while in dev we use the django default db.

In theory if we'd like better segregation of duty and microservices' concept then we should run two distinct containers one from a plain postgres image and another from a postgis one with attached a data container volume for each in case of easier backup and data migration policies. Maybe in future this makes sense.

Dockerfile
fvanderbiest: You mean `geonode/django` is a base image, also used in https://github.com/GeoNode/geonode-project/blob/docker/Dockerfile#L1. We were not aware of it.

The `terranodo/django` image is indeed built from https://github.com/GeoNode/django-docker but this lacks readability, not to mention the `geonode` image tag.

The `geonode-project` duplicates code... where should we contribute ?
We're a bit lost ;-)

docker-compose.yml
francbartoli: good point! I'd love if you can have a safe sanitising from v2 to v3 docker-compose syntax

Dockerfile
francbartoli: Probably https://github.com/GeoNode/geonode-project/blob/docker/Dockerfile#L1 it missed `geonode/django` rather than `terranodo/django`.

However we maintain such images in the official [docker hub](https://hub.docker.com/u/geonode/) where `geonode/django` is built from our django-docker github repo and definitively inherits from your proposed base `python:2.7.9`

fvanderbiest: OK, we're rebasing our work on top of `geonode/django`. Thanks for the pointers.

docker-compose.yml
fvanderbiest: What about keeping common env vars in the compose file, and using the env files for the differences between dev and prod ?

Dockerfile
Vampouille: Done with 0358de7b16cb8251adc61ca9ce1c849ebd6fac9a

package/docker/docker-entrypoint.sh
Vampouille: Done with 0358de7b16cb8251adc61ca9ce1c849ebd6fac9a

docker-compose.yml
francbartoli: Makes sense to have a common env file and inherits from that for those environment differences. In general going with an env file has advantages for better pointing from the documentation and keeping the compose file as much as possible more readable IMHO

docker-compose.yml
fvanderbiest: We would first remove the `data_dir_conf` container.
Then create 2 tags for the `geonode/geoserver` image, one `2.9.x-oauth2` and another one named `2.9.x`. These images contain a geoserver datadir coming from http://build.geonode.org/geoserver/latest/data-${GEOSERVER_VERSION}.zip, unzipped into `/mnt/geoserver_datadir`. This directory will be runtime copied in the `geoserver_datadir` named volume if empty (this is standard docker feature).
We'd also add an entrypoint script which updates geoserver configuration according to env vars.

francbartoli: Does this change the behavior from the current data container volume?

I mean for example now there are advantages for keeping data safe although you stop the containers as we stated in the [repo](https://github.com/GeoNode/data-docker) readme

fvanderbiest: Data is safe in named volumes, even when containers using them are stopped.

docker-compose.yml
fvanderbiest: OK.

docker-compose.yml
francbartoli: So geonode/data image would become deprecated...right? In this case we'd like to add a bit of documentation on how to migrate data already existing

francbartoli: Maybe this [issue](https://github.com/docker/compose/issues/4379) can be used as a reference for documenting a way to be backward compatible with those which are using v2 and data container volume

fvanderbiest:
@Vampouille there are several issues running the composition locally:
```
django_1         | Loading sample data and admin:admin account
django_1         | System check identified some issues:
django_1         | 
django_1         | WARNINGS:
django_1         | base.ResourceBase.tkeywords: (fields.W340) null has no effect on ManyToManyField.
django_1         | Traceback (most recent call last):
django_1         |   File "manage.py", line 29, in <module>
django_1         |     execute_from_command_line(sys.argv)
[...]
django_1         | django.db.utils.OperationalError: Problem installing fixture '/usr/src/app/geonode/people/fixtures/sample_admin.json': Could not load people.Profile(pk=1000): no such table: people_profile
django_1         | System check identified some issues:
django_1         | 
django_1         | WARNINGS:
django_1         | base.ResourceBase.tkeywords: (fields.W340) null has no effect on ManyToManyField.
django_1         | Traceback (most recent call last):
django_1         |   File "manage.py", line 29, in <module>
django_1         |     execute_from_command_line(sys.argv)
django_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
django_1         |     utility.execute()
[...]
django_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/sqlite3/base.py", line 318, in execute
django_1         |     return Database.Cursor.execute(self, query, params)
django_1         | django.core.serializers.base.DeserializationError: Problem installing fixture '/usr/src/app/geonode/base/fixtures/default_oauth_apps_docker.json': no such table: people_profile
django_1         | System check identified some issues:
django_1         | 
django_1         | WARNINGS:
django_1         | base.ResourceBase.tkeywords: (fields.W340) null has no effect on ManyToManyField.
django_1         | Traceback (most recent call last):
django_1         |   File "manage.py", line 29, in <module>
django_1         |     execute_from_command_line(sys.argv)
[...]
django_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/sqlite3/base.py", line 318, in execute
django_1         |     return Database.Cursor.execute(self, query, params)
django_1         | django.db.utils.OperationalError: Problem installing fixture '/usr/src/app/geonode/base/fixtures/initial_data.json': Could not load base.TopicCategory(pk=13): no such table: base_topiccategory
django_1         | Bootstraping Geoserver database...
django_1         | ----------------------------------
django_1         | Creating new role geoserver...OK
django_1         | Creating new database ...OK
django_1         | Adding postgis extension...OK
```

docker-compose.yml
```diff
@@ -1,53 +1,83 @@
...
 
   celery:
     image: geonode/django
     links:
       - rabbitmq
-      - postgres
+      - database
       - elasticsearch
+    entrypoint: sh -c
     command: celery worker --app=geonode.celery_app:app -B -l INFO
```
fvanderbiest: The celery container stops with return code 0 when using the composition from the master branch, and also on this one. Do we miss something here ?

docker-compose.yml
fvanderbiest: > So geonode/data image would become deprecated...right?

This is correct.

docker-compose.yml
```diff
@@ -1,53 +1,83 @@
...
+      - elasticsearch_data:/usr/share/elasticsearch/data
 
   rabbitmq:
-     image: rabbitmq
+     image: rabbitmq:3.6
+     volumes:
+      - rabbitmq_data:/var/lib/rabbitmq
 
   django:
     image: geonode/django
```
fvanderbiest: Image cannot be named as the base image is derives, or strange problems might occur ...
Proposing renaming to `geonode/geonode`

docker-compose.yml
francbartoli: @fvanderbiest Can you test it with the branch 2.6.x?

docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       - rabbitmq_data:/var/lib/rabbitmq
 
   django:
-    image: geonode/django
+    image: geonode/geonode
```
francbartoli: This change requires probably that we modify the name into our docker hub. What is the mean of such commit?

Vampouille: It's "informative", for newcomers it indicates that this is the main applicative container.

francbartoli:
@fvanderbiest @Vampouille what is the status here?

afabiani:
@Vampouille @francbartoli I see some code conflicts here preventing the PR to be merged.

francbartoli:
@afabiani I think this became a little bit old now and has to be tested after the merge of master imho. I can have a look and hugely review asap possibly before the end of this month

Vampouille:
@francbartoli I will provide a new version of this PR with no conflicts.

francbartoli:
@Vampouille thanks! Can you retain also the current name for the django image so we don't have to change it in dockerhub?

Vampouille:
> Can you retain also the current name for the django image so we don't have to change it in dockerhub?

yes of course

docker-compose.yml
Vampouille: I revert image name to geonode/django

Vampouille:
I create a new PR for new version based on master branch: https://github.com/GeoNode/geonode/pull/3451

Issue #3091, opened 2017-05-25
REGISTRATION_OPEN setting


cezio:
There's ``REGISTRATION_OPEN`` setting, that is set to ``False`` by default. It's used in signup template only:
https://github.com/GeoNode/geonode/blob/ea645f27894ad73deeeefa53e5d568f2d5ec74a2/geonode/templates/account/signup.html#L15

It doesn't prevent creating new user however - proper POST request will create new user anyway. Also, this setting overlaps in functionality with ``ACCOUNT_OPEN_SIGNUP`` setting from django-user-account app. This faulty setting can be replaced with ``ACCOUNT_OPEN_SIGNUP`` in template. Also, this flag can be set to ``False`` by default.

PR #3092, opened 2017-05-25
use ACCOUNT_OPEN_SIGNUP instead of REGISTRATION_OPEN

cezio:
#3091 fix


capooti:
Hi, thanks a lot. Could you please update the documentation as well?
http://docs.geonode.org/en/master/tutorials/advanced/geonode_settings/settings.html?highlight=registration_open

PR #3105, opened 2017-06-02
ASCII support

ismailsunni:
Make GeoNode supports asc file uploader.

At first I just want to make the support available with QGIS Server backend, but it seems it will work also on geoserver. This PR is only addressing the GeoServer backend one. The QGIS Server backend will be added in the another PR (since we move qgis_server to geonode).

cc @gubuntu @lucernae

### DIFF
--- a/geonode/layers/forms.py
+++ b/geonode/layers/forms.py
@@ -137,8 +137,8 @@ class LayerUploadForm(forms.Form):
         if not cleaned["metadata_upload_form"] and base_ext.lower() not in (
                 ".shp", ".tif", ".tiff", ".geotif", ".geotiff", ".asc"):
             raise forms.ValidationError(
-                "Only Shapefiles and GeoTiffs are supported. You uploaded a %s file" %
-                base_ext)
+                "Only Shapefiles, GeoTiffs, and ASCIIs are supported. You "
+                "uploaded a %s file" % base_ext)
         elif cleaned["metadata_upload_form"] and base_ext.lower() not in (".xml"):
             raise forms.ValidationError(
                 "Only XML files are supported. You uploaded a %s file" %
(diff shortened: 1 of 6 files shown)


ismailsunni:
Btw, I want to add unit test, but I can't find asc file in the gisdata package. I made a PR to add an asc file to the gisdata package: https://github.com/GeoNode/gisdata/pull/6

afabiani:
@ismailsunni is this PR ready to merge? Or you still have some commits to push? Also, can you update the branch (or I can do this for you)?

ismailsunni:
Hi @afabiani 
I have updated the branch.

Actually, I want to add a unit test for uploading the asc file, but there is no asc file in the gisdata package. If it's ok to not add that test, the PR is ready to merge.

afabiani:
@ismailsunni of course if does not require too much effort to you, having test cases would be good. This would allow us to be sure the code still works in the future. Let me know if you have time to add tests or you prefer to close the PR as is.

In any case, before closing it, I will try to test it also on my side.

Thanks again for you work.

pjdufour:
I have a simple question.  What is the purpose of an `.asc` file in this context?  Thanks.

afabiani:
@pjdufour @ismailsunni .asc files are types of Rasters handled by geospatial server. However, as far as I know, it is not sufficient to upload the .asc file, also a correct .prj file must be uploaded along with the file data. The code should also check that both the files are present.

gubuntu:
ultimately asc will be one of many raster formats supported but for now we're implementing GeoTiff and asc download support because they are the formats currently used by the InaSAFE backend.  

The QGIS backend already turned those into TMS caches but since TH and users also need to download the original raster, we fast-tracked that implementation.

PR #3112, opened 2017-06-07
Peg django-oauth-toolkit to 0.12

ppasq:
Django OAuth Toolkit 1.0.0 has dropped support for Django 1.8 so a constraint to version <1 is added in the setup.py file.


simod:
Thanks paolo. By alessio fixed it already. Btw is see some extra commits here. Closing

PR #3151, opened 2017-07-01
set thumbnail function  for geonode-client

hishamkaram:
- Issues
     - set thumbnail button not working with geonode-client 
- solution: 
     - geonode-client use [boundless sdk](https://github.com/boundlessgeo/sdk) which contains a component to create thumbnail or create image for current view as blob so I created a function to post this image data (blob) to our endpoint [here](https://github.com/GeoNode/geonode-client/pull/173). 
     -  but our end point generate thumbnail server-side so we need to skip this step if `preview=='react'`(a flag in the request to identify sender). 

this PR skip generating image (server-side) and save blob data as thumbnail image

### DIFF
--- a/geonode/base/admin.py
+++ b/geonode/base/admin.py
@@ -33,15 +33,25 @@ from treebeard.forms import movenodeform_factory
 
 from modeltranslation.admin import TranslationAdmin
 
-from geonode.base.models import (TopicCategory, SpatialRepresentationType, Region, RestrictionCodeType,
-                                 ContactRole, Link, Backup, License, HierarchicalKeyword)
+from geonode.base.models import (
+    TopicCategory,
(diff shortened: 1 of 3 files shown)


ehl120:
I'd like remove the thumbnails place holder on bother layers/ and maps/ templates because the thumbnails don't work for me either. However the layouts (with panels.html) folder doesn't exist in this path:

/usr/local/lib/python2.7/dist-packages/geonode/layers/templates

Where to customize the page layout? Thanks a bunch.

afabiani:
@hishamkaram there are failures with Travis, can you check?

hishamkaram:
@afabiani Done

afabiani:
thanks @hishamkaram

Issue #3192, opened 2017-08-02
geonode will fail in tests when USE_TZ is set to True


cezio:
When GN is used with USE_TZ, tests will fail during setup:
```
in setUp
   create_models(type='layer')
 File "geonode/geonode/base/populate_test_data.py", line 206, in create_models
   l.save()
...
 File ".virtualenvs/geonode/lib/python2.7/site-packages/django/db/models/
fields/__init__.py", line 1476, in get_prep_value
   value = timezone.make_aware(value, default_timezone)
 File ".virtualenvs/geonode/lib/python2.7/site-packages/django/utils/
timezone.py", line 358, in make_aware
   return timezone.localize(value, is_dst=None)
 File ".virtualenvs/geonode/lib/python2.7/site-packages/pytz/tzinfo.py", line 
309, in localize
   loc_dt = dt + delta
OverflowError: date value out of range
```

This is caused by ``geonode.base.populate_test_data.create_models()`` which uses fixed date values like: 1900-06-03, 5000-11-01, 0001-01-01

This can be fixed by using dates generated during population.

cezio:
closing

PR #3193, opened 2017-08-02
geonode test data use generated dates

cezio:
#3192


afabiani:
@cezio Travis does not pass

PR #3221, opened 2017-08-19
Fixes #3191

### LINKED ISSUE #3191: Import layer's keywords from geoserver not works
If i insert manually or programmatically the keywords in Geoserver's layer and i launch "python manage.py updatelayers" in Geonode 2.6 the keywords are not visualized.

capooti:
Synchronize keywords both directions now (geoserver <--> django)

### DIFF
--- a/geonode/geoserver/signals.py
+++ b/geonode/geoserver/signals.py
@@ -139,6 +139,11 @@ def geoserver_pre_save(instance, sender, **kwargs):
         if getattr(ogc_server_settings, "BACKEND_WRITE_ENABLED", True):
             gs_catalog.save(gs_layer)
 
+    if gs_resource:
+        if gs_resource.keywords:
+            for keyword in gs_resource.keywords:
+                instance.keywords.add(keyword)
+
     """Get information from geoserver.
 
        The attributes retrieved include:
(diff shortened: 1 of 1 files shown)


capooti:
Actually I did not run tests, but this was working when testing it manually.
I will have a look

geonode/geoserver/signals.py
```diff
@@ -139,6 +139,11 @@ def geoserver_pre_save(instance, sender, **kwargs):
         if getattr(ogc_server_settings, "BACKEND_WRITE_ENABLED", True):
             gs_catalog.save(gs_layer)
 
+    if gs_resource:
+        if gs_resource.keywords:
+            for keyword in gs_resource.keywords:
+                instance.keywords.add(keyword)
```
capooti: @afabiani I think I can add a related keyword it the layer is already existing.
I agree with you that this can be problematic when creating a new layer. Need to figure out what is the best way to handle it

capooti:
@afabiani it looks to be a django-taggit bug: https://github.com/alex/django-taggit/issues/205
As far as I remember in the past I have successfully used the post_save signal to create related items.
It seems complicated to find a workaround for this, so I am closing this PR.

afabiani:
@capooti see my proposed implementation on master here https://github.com/GeoNode/geonode/pull/3223

we can backport this on 2.6 also.

PR #3222, opened 2017-08-22
Added settings variable to inform geonode-client to request map tiles…

travislbrundage:
… cross origin

Coincides with this PR: https://github.com/GeoNode/geonode-client/pull/179 (Not needed, but won't accomplish anything without this)

### DIFF
--- a/geonode/context_processors.py
+++ b/geonode/context_processors.py
@@ -133,6 +133,9 @@ def resource_urls(request):
             False
         ),
         THESAURI_FILTERS=[t['name'] for t in settings.THESAURI if t.get('filter')],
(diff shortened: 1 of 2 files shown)


pjdufour:
If possible, could you please rename to something more verbose, e.g., `MAP_CLIENT_CROSS_ORIGIN`?  That name is too general, since could apply to any number of things.  Also, just curious, why do you even need a setting?  Why wouldn't you want "cross origin" always enabled for the client?  I assume it has something to do with allowing CORS requests?  Thank you.

afabiani:
merging since the requested changes have been done

Issue #3269, opened 2017-09-06
GNIP: Create empty layer


capooti:
# Overview

In many GeoNode instances there is the need to create layers from scratch. A contrib application is needed which will let the user to create an empty layer. The user will be able for that given layer to assign:

* layer name
* layer title
* geometry type (Points, Lines, Polygons)
* a number of fields

For each field the user will be able to define the name and its type (String, Integer, Float, Date)

The application will be part of geonode.contrib applications and requires PostgreSQL/PostGIS, so it will not be part of default INSTALLED_APPS

A django management command will be include, to create empty layers from the Django shell.

# Proposed by

[Paolo Corti and Ben Lewis, Harvard CGA](http://gis.harvard.edu/)
[Francesco Stompanato and Dimitris Karakostis, WFP](https://geonode.wfp.org/)

# Assigned to release

2.8

# Mockup

<img width="1204" alt="screenshot 2017-09-06 17 13 53" src="https://user-images.githubusercontent.com/148380/30135074-79b27286-9327-11e7-991c-497802027d91.png">

capooti:
I am pretty close to submit a PR for this, here is the branch:
https://github.com/capooti/geonode/tree/create-layer-app

afabiani:
+1
it looks interesting. Just a question, this empty layer will be connected somehow to a geospatial dataset or it will remain just an empty descriptor?

capooti:
Yes, this application will create a PostGIS layer and add it to GeoServer, then register as a layer in GeoNode. It is useful in many scenarios when there is a need to digitize information from basemaps or create map notes using WFS/T.

capooti:
Here is the application in action:

http://recordit.co/anFE0BQRYr

PR #3275, opened 2017-09-07
Adding the createlayer contrib application, as per GNIP #3269

capooti:
This pull request implement GNIP #3269

The PR provides the createlayer contrib application with its related createvectorlayer django administrative commands.

Tests and documentation are included. To run tests you need to follow the instructions provided at the top of the createlayer/tests.py file

Here is the createlayer application in action: http://recordit.co/anFE0BQRYr


francbartoli:
+1 like a lot, thanks @capooti

afabiani:
Thanks @capooti very cool features. @simod in general I would agree with you, my only concern is if this works assuming GeoServer as backend or can be extended to other servers too. However +1 for the PR

kalxas:
Big +1 from me!
Thanks @capooti

davekennewell:
+1 This is great - I would also love to see it in the core, I've often needed this and used workarounds.

capooti:
@simod what do you mean with adding to the core? Moving the code out from contrib and enabling the application in INSTALLED_APPS by default?

This could be done but:
* we need to disable the application when datastore is not based on PostGIS
* we need to find out a way to run tests, they require the PostGIS datastore
* we need to disable the application when OGC server is not GeoServer, as it creates the layer using the GeoServer REST API (though it should be pretty trivial to change that code and run SQL to create the layer)

I have a question for you guys in charge of the release process: now this application is in master: do I need to send PR vs 2.7.x? It should be relatively easy also to merge vs 2.6.x but at this point I am not sure if for 2.6.x new features are accepted?

simod:
I see your points Paolo, maybe we will add to the core when all the tests will be running on postgres and postgres will be the default db, still it would be good to have a  check for geoserver anyway.

I think you could reissue the PR also for 2.7 while for 2.6 I wouldn't, this is not a minor feature. :)

PR #3290, opened 2017-09-13
 - Update Pinax Notifications to v.4.0.0, fix templates and enable them by default

afabiani:


geonode/settings.py
```diff
@@ -1028,7 +1028,7 @@
 FREETEXT_KEYWORDS_READONLY = False
 
 # notification settings
-NOTIFICATION_ENABLED = False or TEST
+NOTIFICATION_ENABLED = True or TEST
```
simod: @afabiani we enable them permanently now?

afabiani: @simod this is my idea, they works without no further setup and this option allows setup to create migrations at the beginning. If it's ok with you also I would merge the PR.

kalxas: +1 to merge if notifications now work

PR #3316, opened 2017-09-22
Add the "Add the layer to an existing map" feature in layer detail page

capooti:
As in the following screenshot

screenshot 2017-09-22 17 38 55


afabiani:
I would say ok, nice feature... but the combo box is not really a good choice if you have a big number of maps. At least it need autocompletion. Also, the list of maps takes into account user permissions?

PR #3324, opened 2017-09-26
Fixes messages colors.

zunware:
Messages now display with the correct color for the message type.


zunware:
This is for branch version 2.6.x

PR #3353, opened 2017-10-18
 - Group Managers and Staff Members can now Moderate resources before publishing them

afabiani:


capooti:
@afabiani do you still need to work on this before I check it?

afabiani:
@capooti nope the PR is ready

capooti:
@afabiani this is a great new feature. Awesome work. I have some comments to do:

* do we have a setting to enable this (resource moderation) or this is now the default? Because I think this is adding complexity (and slowness) to the API requests, and not every GeoNode instance needs this workflow
* Is it possible to include some documentation and some tests?
* did you consider existing django applications which let integrate a workflow model? (like for example django-viewflow). Maybe that would add unrequired complexity?

PR #3356, opened 2017-10-24
 - Fix an issue on catalogue APIs when looking for Public Groups IDs

afabiani:


geonode/catalogue/views.py
```diff
@@ -108,9 +108,13 @@ def csw_global_dispatch(request):
                 for group in request.user.groups.all():
                     groups_ids.append(group.id)
 
-            public_groups = GroupProfile.objects.exclude(access="private").values('group')
+            public_groups = GroupProfile.objects.exclude(access="private").values('group').distinct()
             for group in public_groups:
```
simod: Which is the case when we have more than one group per GroupProfile?

afabiani: no case, you are right. It is a refuse. Removing. Thanks @simod

PR #3360, opened 2017-10-24
Adding charset as option in importlayers command

Gpetrak:
I needed to import a bunch of datasets in a specific charset (not in UTF-8 which is the default) with the command importlayers. I realized that there is not an option of specifying a charset. So I extended the importlayers command in order to include that option. 
What do you think about it ?


simod:
thanks @Gpetrak. There a conflict in utils.py, can you fix it?

PR #3372, opened 2017-10-27
 - Admin and Staff members should be able to see all resources anyway

afabiani:


simod:
@afabiani this applies only to the groups right?

afabiani:
@simod yep. Can you approve again? There were some tests failing!

PR #3374, opened 2017-10-27
 - Display Group Profile Titles on Info page instead of Group Slugs

afabiani:


geonode/documents/views.py
```diff
@@ -113,12 +113,19 @@ def document_detail(request, docid):
...
+                group = GroupProfile.objects.get(slug=document.group.name)
+            except GroupProfile.DoesNotExist:
+                group = None
         context_dict = {
             'perms_list': get_perms(
                 request.user,
                 document.get_self_resource()),
             'permissions_json': _perms_info_json(document),
             'resource': document,
+            'group': group,
```
capooti: I don't get why you want to add group in the context dict. It should be always possible to get it from resource.group in the template

afabiani: It is not the group but the group profile associated to it

afabiani: It is not the group but the group profile associated to it

afabiani: @capooti unless further clarifications, if you are ok with the solution proposed, can you approve the PR again? Thanks.

capooti: oh I see, I thought geonode.Group had a FK with django.auth.Group (as for geonode.Profile)

PR #3377, opened 2017-10-27
Fixes #3376

### LINKED ISSUE #3376: updatelayers does not syncronize permissions with GeoFence

capooti:
updatelayers will now synchronize django-guardian permissions with GeoFence


afabiani:
@capooti can you please update with master? There is a test using an external service which randomly fails. A patch on master should fix the problem. Thanks.

capooti:
done. thanks @afabiani

PR #3364, opened 2017-10-25
QGIS Server Related PR to upstream.

lucernae:
Related with stage 1 for Issue #3362 : Merging QGIS Server Backend.

Include changes to improve unittests and CI:
- Parallel job for GeoServer and QGIS Server backend
- Switch from coveralls to codecov for code coverage
- Add methods and decorators to check active backend

PR #3378, opened 2017-10-28
Add codecov badge and fix syntax errors

francbartoli:


README
```diff
@@ -2,6 +2,9 @@
     :alt: Coverage Badge
     :target: https://coveralls.io/github/GeoNode/geonode?branch=master
```
Gustry: Why not remove the badge from coveralls.io?

afabiani:
@francbartoli can you please update with master? There is a test using an external service which randomly fails. A patch on master should fix the problem. Thanks.

README
francbartoli: Thanks @gustry for the suggestion. I left it because I did not find any discussion in the dev list what are the benefits to switch to codecov. Meanwhile we can have both

Gustry: AFAIK, coveralls is not used anymore, no?

Gustry: For reference: https://github.com/GeoNode/geonode/pull/3364#discussion_r146863773

francbartoli:
Yes sure @afabiani

README
francbartoli: Thanks for the reference. I’m personally inclined to having it approved from the dev community rather than the preferences of one single developer

lucernae: Thank's @francbartoli @Gustry . I'm fine with coveralls if it can support travis's matrix test, but from what I know coveralls doesn't support this (https://github.com/lemurheavy/coveralls-public/issues/156). When QGIS Server PR comes, it will use parallelized job on Travis to test each GeoServer and QGIS Server backend independently. Coveralls will report significantly lower coverage, because when it test GeoServer Backend, it will not execute QGIS Server Backend code and vice versa. That's the main reason I switched to codecov. Codecov will aggregate coverage from each parallel tests, so we have a total coverage. If this is a problem, then perhaps I can try reaching out through the dev community (through the dev mailing lists, I suppose?) to explain the reason why I proposed codecov.

francbartoli: @lucernae thanks for your explanation. I'm fine with Codecov and much more happy to know that it can support both test cases. I think it's worth for geonode.
Can you just explain it through the dev mailing list so after the approval we can then definitively remove coveralls?

lucernae: Sure, I will send email soon.

PR #3379, opened 2017-10-28
Patch paver setup

francbartoli:


afabiani:
@francbartoli can you please update with master? There is a test using an external service which randomly fails. A patch on master should fix the problem. Thanks.

PR #3390, opened 2017-10-31
 - Removing pdb dependency causing problems with Apache2 and updating 'pip -r requirements.txt' string

afabiani:


capooti:
hi @afabiani there are unrelated files in this PR.
Curious: what kind of problem is giving pdb with httpd?

afabiani:
Hi @capooti, no unrelated files, the proposal is to change

`pip install -r requirements.txt`

to 

`pip install -r requirements.txt --no-deps --no-cache -U`

everywhere.

About the second question, on a fresh Ubuntu 16 setup with Apache2, I get a dependency exception on WSGI if pdb is installed on virtualenv. Since pdb is not a strict requirements of the project, it is better to get rid of it. Developers that need it can install it manually if they want to debug.

What do you think about this?

PR #3391, opened 2017-11-02
Removing a duplicate in requirements which is causing an error on pip install

capooti:
Removing a duplicate in requirements which is causing an error on pip install


capooti:
It came out that this was caused by an old version of pip. Update pip is more resilient. Though we can still make same cleaning

PR #3396, opened 2017-11-04
Replace default crs with 3857

### LINKED ISSUE #3367: Remove 900913 definition in remote services
Apologies in advance if I'm using any terminology etc. incorrectly, I'm rather new with GeoNode.

Steps to reproduce: 
1) Install GeoNode 2.6 on Ubuntu 16.04 per instructions here http://docs.geonode.org/en/master/tutorials/install_and_admin/geonode_install/index.html
2) Add the following Remote Service: http://www.agr.gc.ca/atlas/services/mapservices/aafc_watershed_2013/MapServer/WMSServer
3) Browse any layer, see pink tiles

When checking GetCapabilities for the WMS server, you can see the following coordinate systems defined:

<CRS>CRS:84</CRS>
<CRS>EPSG:4326</CRS>
<CRS>EPSG:3857</CRS>

<CRS>EPSG:102100</CRS>

But when you examine the actual URL being generated by GeoNode for layer images

http://www.agr.gc.ca/atlas/services/mapservices/aafc_watershed_2013/MapServer/WMSServer?LAYERS=14&TRANSPARENT=TRUE&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&STYLES=&FORMAT=image%2Fpng&TILED=true&SRS=EPSG%3A900913&BBOX=-15028131.255,5009377.085,-12523442.7125,7514065.6275&WIDTH=256&HEIGHT=256

SRS=EPSG%3A900913

Is there something special that needs to be done with this type of remote service?

francbartoli:
Fixes #3367

### DIFF
--- a/geonode/services/views.py
+++ b/geonode/services/views.py
@@ -667,8 +667,9 @@ def _register_indexed_layers(service, wms=None, verbosity=False):
             srid = None
             # Some ArcGIS WMSServers indicate they support 900913 but really
             # don't
-            if 'EPSG:900913' in wms_layer.crsOptions and "MapServer/WmsServer" not in service.base_url:
-                srid = 'EPSG:900913'
+            if 'EPSG:900913' in wms_layer.crsOptions and "MapServer/WmsServer"\
+                    not in service.base_url:
+                srid = 'EPSG:3857'
             elif len(wms_layer.crsOptions) > 0:
                 matches = re.findall(
                     'EPSG\:(3857|102100|102113)', ' '.join(
--- a/geonode/services/views.py
+++ b/geonode/services/views.py
@@ -286,8 +286,12 @@ def _process_wms_service(
     if wms is None:
         wms = WebMapService(url)
     try:
-        base_url = _clean_url(
-            wms.getOperationByName('GetMap').methods['Get']['url'])
+        method_url = [
+            method['url']
+            for method in wms.getOperationByName('GetMap').methods
+            if method['type'] == 'Get'
+        ][0]
+        base_url = _clean_url(method_url)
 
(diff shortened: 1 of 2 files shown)


afabiani:
@francbartoli did you check that having 3857 as default in settings still works with default map layers in GeoExplorer? Since this is the default setup we need to be sure that clean installations still work well.

francbartoli:
@afabiani I run all tests but actually I did not make any further tests manually. Going to check that out now.

francbartoli:
@afabiani effectively everything seems to work fine apart from the geoexplorer viewer which is created with a zoom and a bbox with no correspondence to the uploaded layer. I'm probably missing something here where those settings are set, perhaps within geoexplorer?

capooti:
One thing to consider here is also existing GeoNode instances. I can't check this right now, but we should make sure existing maps are not broken. If that is the case, we should provide a migration script I think.

francbartoli:
Totally agreed, thanks @capooti for the hints. I haven't thought about that

capooti:
who knows @francbartoli, maybe it would keep working. I don't remember if srs is serialized somewhere in db. It would require some investigation

afabiani:
Will have a look as soon as I have some available time-slot

afabiani:
@francbartoli starting from your branch I have updated the PR in order to correctly handle the CRS from GeoExplorer also. See this PR

https://github.com/GeoNode/geonode/pull/3417

francbartoli:
Nice @afabiani thanks

afabiani:
@francbartoli I'm going to close this since changes here are part of #3417

PR #3406, opened 2017-11-07
Fixed geonode layers analytics and w*s layers analytics

### LINKED ISSUE #3403: monitoring: frontend layer analytics is missing data
Layer and WxS analytics tables at the bottom are not showing correct data. This is two-element bug:
 * frontend view with layer/WxS stats is not updated when filtering elements (layer, time window) was selected,
 * backend api is collecting data but assigned them to wrong parameters due to variable not reset during loop, also rate values aggregation is returning wrong values.

### LINKED ISSUE #4303: SITEURL not set in SPC
After uploading a layer I see broken thumbnails: sometimes the URL is wrong, sometimes it is just a placeholder ("no image"). The path is correct, but hostname and port are wrong (localhost:8000).

I am using SPCGeonode. `HTTP_HOST` and `HTTP_PORT` are correctly set.

cezio:
fix #3403

rebuilded bundle.js for monitoring
monitoring: reset ows service for metric collection
monitoring: calculate rates aggregates


afabiani:
@cezio it seems that travis is failing

PR #3409, opened 2017-11-08
Switch to https

Vampouille:
This PR use 'OpenStreetMap Mapnik' as default source for baselayer which allow http and https connection. This will remove browser warning about mixed content when using HTTPS to connect to geonode.


afabiani:
@Vampouille can you do the PR on master? We will do a backport on 2.7 in a few days.
Thanks.

Vampouille:
Same PR on master : https://github.com/GeoNode/geonode/pull/3413

PR #3414, opened 2017-11-09
release to stable download and gdal stable versions

stern0:
- Changed apt-repository from release to stable 
The release download isn't in apt-repository which fails while adding, the stable version downloads successfully.

- Changed gdal from 1.10.1 to 1.11.3.3 since it is the stable version the former had broken dependencies causing run-time errors.


afabiani:
@stern0 thanks for the update.

Would you mind to update your PR with master code in order to let Travis pass the failing test?

stern0:
@afabiani all done.

PR #3422, opened 2017-11-10
 - Avoid storing access_token (which have expiration) into the map co…

afabiani:
…ntext, and inject them dinamically instead


afabiani:
Ready to be merged. It fixes somethings on the map management along with OAUTH2 access_tokens:

- First it is not safe to save into DB the access_tokens since they have a validity period. Those must be injected as query params instead. I injected them on base url of layers since different JS frameworks translate params to capital case. OAuth2 may complain about that.

- There were a small issue with map details. Parameters must be propagated to all layers of the map instead of just the selected ones.

PR #3428, opened 2017-11-13
 - Relax base_file checks: Unchecked GDAL Info failure

afabiani:
Removing assertion since if the original file does not exists anymore it won't be possible to update Metadata anymore.

This kind of checks must be improved in the future. The code cannot just die if a check fails.


afabiani:
@francbartoli agree with your statement. I'm going to open an enanchement request to review the whole workflow.

PR #3468, opened 2017-11-26
use settings.DEFAULT_WORKSPACE

ahmednoureldeen:
use settings.DEFAULT_WORKSPACE  so upload layer and styles are stored in configured workspace

### DIFF
--- a/geonode/api/resourcebase_api.py
+++ b/geonode/api/resourcebase_api.py
@@ -321,7 +321,6 @@ class CommonModelApi(ModelResource):
 
         # Filter by Type and subtype
         if type_facets is not None:
-
             types = []
             subtypes = []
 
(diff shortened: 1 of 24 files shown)


afabiani:
Hi, thanks for the proposal. I'll test it asap. Even though I would envisage the possibility of using global styles also instead of forcing using the ones bounded to the default workspace only.

Other than this, I guess that for a change like this would be good to have few test cases consolidating the use case.

afabiani:
@ahmednoureldeen I have ported your changes here https://github.com/GeoNode/geonode/pull/3488 by adding also a check in order to fetch the SLD from glolbal catalog other than default workspace.

Therefore I'm closing this one which is now replaced by mine above.

PR #3519, opened 2017-12-19
[Fixes #3518] [Monitoring plugin] - Collection of statistics should take into account TimeZones

### LINKED ISSUE #3518: [Monitoring plugin] - Collection of statistics should take into account TimeZones
Currently the collection and storage of statistics does not handle correctly the different timezones between the local server and running services.

All the dates should be compared by taking into account possible differences between datetimes TimeZones.

afabiani:
Fixes #3518

### DIFF
--- a/geonode/contrib/monitoring/collector.py
+++ b/geonode/contrib/monitoring/collector.py
@@ -19,6 +19,7 @@
 #########################################################################
 import logging
 import re
+import pytz
 from datetime import datetime, timedelta
 from decimal import Decimal
 from itertools import chain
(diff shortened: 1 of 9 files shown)


geonode/contrib/monitoring/collector.py
```diff
@@ -130,7 +131,8 @@ def get_mem_label(*args):
                               )
                              )
 
-        collected_at = datetime.now()
+        utc = pytz.utc
+        collected_at = datetime.utcnow().replace(tzinfo=utc)
```
cezio: Other way is to use `tzinfo.localize`, so in this case, `collected_at = utc.localize(datetime.now())`

PR #3541, opened 2017-12-31
Geosites- support group per site

ahmednoureldeen:
same implementation as Users


francbartoli:
@ahmednoureldeen can you take a look why tests are failing? You should make travis happy with tests before somebody more expert on geosites than me can review and merge

ahmednoureldeen:
@francbartoli I will check :)

PR #3547, opened 2018-01-08
 - Fix GeoNode 2.6.3 to 2.7.x+ Migration

afabiani:


capooti:
just a note: it is possible to create a data migration in case like this
https://docs.djangoproject.com/en/2.0/topics/migrations/#data-migrations

afabiani:
sure, but is it actually needed? 2.6.3 and 2.7 schemas (at least for the common models) are almost aligned.

Well it would be useful to have for alternate and typename, but since we have inherited resources, it won't work. This is why I had to create the management command to be executed at the end of paver sync.

capooti:
I didn't think that it won't work for inherited resources. Sounds good

PR #3559, opened 2018-01-11
Remove Double Migrations [Master]

Coop56:
Remove fields that were added in two separate migrations causing errors when you stand up geonode.  This PR needs to go into master as well as 2.7.x so I will be creating two PRs for it.  You can see the other instance of these same migrations that I kept at https://github.com/GeoNode/geonode/blob/master/geonode/base/migrations/26_to_27.py.


cuttlefish:
A migration shouldn't change once it's been shipped. I've double checked and these are the only changes that were made to it.

PR #3602, opened 2018-02-08
[Fixes #3600] [GNIP] APIs to decouple import and load of GeoNode Client Libraries

### LINKED ISSUE #3600: [GNIP] APIs to decouple import and load of GeoNode Client Libraries
Currently the load of a GIS client library on GeoNode, is done through a set of if - else statements on the templates, like, as an instance:

```
  {% if preview == 'geoext' %}
    {% include "layers/layer_geoext_map.html" %}
  {% elif preview == 'OL3' %}
    {% include "layers/layer_ol3_map.html" %}
  {% elif preview == 'react' %}
	{% include 'geonode-client/layer_map.html' %}
  {% else %}
    {% include "layers/layer_leaflet_map.html" %}
  {% endif %}
```

The proposal here would be to somehow include the client libraries from settings by dynamically loading the inclusion template headers through a custom tag library.

What I would like to achieve is something like changing the client library by simply adding the dependency to GeoNode and doing something like this on local settings

`GEONODE_CLIENT_HOOKSET = "mygeonode.client.hooksets.MyNewReactHookSet"`

and have the pages automatically switch the GIS client.

afabiani:
This PR makes some code cleanup, by centralizing some methods and remove code redundancies, but mostly propose a way to make GeoNode GIS client libraries pluggable.

This done through a `geonode.client` app which should be used not only to collect the GIS client templates provided by default with GeoNode, but also defines a set of templatetags which can be used to easily plugin any other different client into the layer details or map templates.

How to use it. Instead of doing something like this:

```
  {% if preview == 'geoext' %}
    {% include "layers/layer_geoext_map.html" %}
  {% elif preview == 'OL3' %}
    {% include "layers/layer_ol3_map.html" %}
  {% elif preview == 'react' %}
	{% include 'geonode-client/layer_map.html' %}
  {% else %}
    {% include "layers/layer_leaflet_map.html" %}
  {% endif %}
```

through this new mechanism you can now do this

```
...
{% load client_lib_tags %}
...
{% get_layer_detail %}
...
```

the tag is translated into a method of a `HookSet` that simply translates the call into a custom template inclusion by passing the `context ` to it.

Plugging in you own Hookset is easy as

1. Load your new client lib on `INSTALLED_APPS`
2. Set the variable `GEONODE_CLIENT_HOOKSET` pointing to your own Hookset

And that's it. By default, if no `GEONODE_CLIENT_HOOKSET` is defined, GeoNode will use the GeoExt based one

`GEONODE_CLIENT_HOOKSET = "geonode.client.hooksets.GeoExtHookSet"`

This is the default one, and it is quite straight. Just defines which templates to include

```
class GeoExtHookSet(object):

    # Layers
    def layer_detail_template(self, context=None):
        return 'geoext/layers/layer_geoext_map.html'

    def layer_new_template(self, context=None):
        return 'geoext/layers/layer_geoext_map.html'

    def layer_view_template(self, context=None):
        return 'geoext/layers/layer_geoext_map.html'

    def layer_edit_template(self, context=None):
        return 'geoext/layers/layer_geoext_map.html'

    def layer_update_template(self, context=None):
        return 'geoext/layers/layer_geoext_map.html'

    def layer_embed_template(self, context=None):
        return 'geoext/layers/layer_geoext_map.html'

    def layer_download_template(self, context=None):
        return 'geoext/layers/layer_geoext_map.html'

    # Maps
    def map_detail_template(self, context=None):
        return 'geoext/maps/map_include.html'

    def map_new_template(self, context=None):
        return 'geoext/maps/map_geoexplorer.js'

    def map_view_template(self, context=None):
        return 'geoext/maps/map_geoexplorer.js'

    def map_edit_template(self, context=None):
        return 'geoext/maps/map_geoexplorer.js'

    def map_update_template(self, context=None):
        return 'geoext/maps/map_geoexplorer.js'

    def map_embed_template(self, context=None):
        return 'geoext/maps/map_geoexplorer.js'

    def map_download_template(self, context=None):
        return 'geoext/maps/map_geoexplorer.js'
```

On your own client library, you can manipulate the context as you wish and pass back to your templates any new variable.


afabiani:
@capooti can you please approve again? I had to merge with master branch

PR #3617, opened 2018-02-13
Add ability to control thumbnail generation implementation

DBlasby:
This allows different implementations of thumbnail generators. This PR configures Geonode to use the old/current implementation - so there should be no changes.

I did, however, make a minor change in ResourceBase#save_thumbnail(). The old implementation (theoretically) allowed multiple thumbnails (just call #save_thumbnail() with different filenames), however, only one is really supported. For example, see ResourceBase#thumbnail_url, ResourceBase#get_thumbnail_url(), and the rest of Geonode just using ResourceBase#thumbnail_url. This makes it more explicit.

There is also a minor change in the layer#post_save() handler. I put a guard on the call to generate a thumbnail. This is because an alternate implementation could be to schedule a celery task to create the thumbnail in the future. This will need to call layer#save() so that search indexers can be made aware that the layer#thumbnail_url has been changed. This could cause a re-generation of the thumbnail (the second call will not cause the thumbnail_url to change so layer#save() will not be called).

### DIFF
--- a/geonode/base/models.py
+++ b/geonode/base/models.py
@@ -1036,6 +1036,8 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin, ItemBase):
         """Determine if the thumbnail object exists and an image exists"""
         return self.link_set.filter(name='Thumbnail').exists()
 
+    # Note - you should probably broadcast layer#post_save() events to ensure
+    # that indexing (or other listeners) are notified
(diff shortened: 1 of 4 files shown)


afabiani:
@DBlasby p.s.: see https://github.com/GeoNode/geonode/pull/3602 for reference on the current work I started to delegate as much as possible to the client library.

PR #3640, opened 2018-02-21
Fixes #3639 - fix layer import to get title and abstract from metadata xml

### LINKED ISSUE #3639: Title and Abstract are never set from metadata xml on import
If a valid metadata file is uploaded along with valid layer files, the abstract and title appear to never be set. 

/geonode/layers/utils.py doesn't seem to try to do this. 

Suggest adding:
title = defaults.get('title',title)
abstract = defaults.get('abstract',abstract)

at about line [482](https://github.com/GeoNode/geonode/blob/374a1c3699c17f2d18054cdb15c32733dd976d71/geonode/layers/utils.py#L482) inside the if statement which starts on line [458](https://github.com/GeoNode/geonode/blob/374a1c3699c17f2d18054cdb15c32733dd976d71/geonode/layers/utils.py#L458)

This would get the title and abstract from the loaded xml, or defaults back to the previous title and abstract variables set previously.

An additional enhancement would be to add a tick box on the layer import screen to allow a user to choose whether the title is derived from the metadata.

JonDHo:
Fixes #3639 by setting the title and abstract from xml metadata on layer import if they exist. Defaults back to previously set variables if title and abstract are not found in the metadata. Thanks to @erinkenna.

### DIFF
--- a/geonode/layers/utils.py
+++ b/geonode/layers/utils.py
@@ -514,6 +514,9 @@ def file_upload(filename, name=None, user=None, title=None, abstract=None,
             else:
                 defaults[key] = value
 
+        title = defaults.get('title',title)
+        abstract = defaults.get('abstract',abstract)
+
     regions_resolved, regions_unresolved = resolve_regions(regions)
     keywords.extend(regions_unresolved)
 


afabiani:
@JonDHo I do not understand what you are trying to do here.
'defaults' are already set to title and abstract. The new ones must be taken from the XML.

Can you please explain?

JonDHo:
All line reference below are to the current 2.7.x commit - https://github.com/GeoNode/geonode/blob/6afd8ce28b70ba5d842b24b065e5c3f7c0e51616/geonode/layers/utils.py

@afabiani what I want to achieve is to get the title and abstract from a metadata xml file if the file and those values exist. Currently this doesn't happen and the title is just set to the slugified file name (same as `name`) and the abstract is either empty or "No abstract has been defined" (or something like that, I can't remember).

I agree that `defaults` does indeed hold the correct title and abstract from the metadata, but that content is never used.

From what I can tell:
- the `title` and `abstract` variables are initially set to None (line 393 - `def upload`)
- "set metadata" starts on L492 (`if 'xml' in files:`)
- the metadata is read and is loaded to the `vals` object
- `vals` is transferred into `defaults` with a few tweaks along the way (L505 onwards - `for key, value in vals.items():`)
- ultimately any `vals.items` which don't need special handling are written out to defaults at L515 (`defaults[key] = value`)

We therefore now have title and abstract existing as values in `defaults`, so that is all good.

But, there is nothing which gets those values out of `defaults` and sets them to the previously declared `title` or `abstract` variables from L393.

The title and abstract are added to `to_update` (L596 and L599 respectively) which is used to update the layer and ResourceBase, but they are still set to None as they have never actually been retrieved from the metadata content.

The two lines I added set the title and abstract named variables to the values in `defaults` if they exist, otherwise, they default back to themselves.

This then has the effect (which is working on our system) of setting a layers abstract and title from the xml file if it is uploaded alongside a shapefile or geotiff.

Hopefully I haven't just blatantly missed something!!! (entirely possible).

Cheers,
Jonathan.

I note that there is an extra section which interacts with title and abstract inside the NLP_ENABLED section of code (L524&525) and actually does exactly what my lines do, but that writes title and abstract to `nlp_metadata`, bit I don't think that is used anywhere either. We aren't using NLP anyway.

afabiani:
@JonDHo thanks.

I'm going to close this as it is replaced by this PR here

https://github.com/GeoNode/geonode/pull/3646

Issue #3561, opened 2018-01-12
collectstatic error on initial install in Stable version 2.6.3


rgdamalerio:
**# when I run sudo -u geonode python manage.py collectstatic**

Traceback (most recent call last):
  File "manage.py", line 29, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 328, in execute
    django.setup()
  File "/usr/local/lib/python2.7/dist-packages/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python2.7/dist-packages/django/apps/registry.py", line 108, in populate
    app_config.import_models(all_models)
  File "/usr/local/lib/python2.7/dist-packages/django/apps/config.py", line 198, in import_models
    self.models_module = import_module(models_module_name)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/geonode/geonode/geonode/base/models.py", line 56, in <module>
    from geonode.utils import bbox_to_wkt
  File "/home/geonode/geonode/geonode/utils.py", line 31, in <module>
    from slugify import Slugify
ImportError: cannot import name Slugify

rgdamalerio:
@lukeman @paltman  @chrisjones-brack3t  @brylie  @benmccall

poligarcia:
Same happens on 2.6.1

lalo73:
Geonode 2.6.x has `"transifex-client>=0.10" as dependency.
The last version of that library is *0.13.0*, which was released 28 days ago.
[This version](https://github.com/transifex/transifex-client/blob/2885996d75b79cbb07ce830462c1caecd3be64f1/requirements.txt) introduces `requests` and  `python-slugify` as dependencies.

`python-slugify` overlaps `awesome-slugify` [dependency](https://github.com/GeoNode/geonode/blob/debian/2.6.1%2Bthefinal0/setup.py#L96).

Install `transifex-client<=0.12.5`, uninstall `python-slugify` *and* `awesome-slugify` and reinstall `awesome-slugify==1.6.5`

poligarcia:
:raised_hands: thanks @lalo73 !

rgdamalerio:
thanks for the help.

[...]

PR #3650, opened 2018-02-28
fix bug cannot import slugify

hishamkaram:
take a look here https://github.com/GeoNode/geonode/issues/3561


afabiani:
@hishamkaram you should update "setup.py" also imho

PR #3666, opened 2018-03-15
consider layer workspace in updating/uploading and geofence rules

ahmednoureldeen:


afabiani:
@ahmednoureldeen hi, currently this seems to be misaligned with master branch. It looks like just some methods have been moved to another class, but also some improvements have been already pushed which probably fix already the issue.

Can you please check? Would like me to take a look?

ahmednoureldeen:
@afabiani Hi, I will align my branch with master and push again.
Thanks.

PR #3714, opened 2018-04-05
fix broken pip install

olivierdalang:
When running `pip install geonode==2.8.0` I get 
```
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-cmEX1w-build/setup.py", line 33, in <module>
        version=__import__('geonode').get_version(),
      File "geonode/__init__.py", line 23, in <module>
        from .celery_app import app as celery_app
      File "geonode/celery_app.py", line 24, in <module>
        from celery import Celery
    ImportError: No module named celery
```

That's because of `from .celery_app import app as celery_app` in __init__.py which I think was forgotten there by mistake (not sure?)

### DIFF
--- a/geonode/__init__.py
+++ b/geonode/__init__.py
@@ -20,8 +20,6 @@
 
 import os
 
-from .celery_app import app as celery_app
-
 __version__ = (2, 9, 0, 'unstable', 0)
 __all__ = ['celery_app']
 


lucernae:
Probably because celery was not installed beforehand?

afabiani:
This not going to work I'm afraid. Probably it will prevent celery tasks to run?

Just thinking out loud... you should install at least Django and celery first, similarly at what we need to do with geonode-project.

olivierdalang:
I checked on Celery's doc, and indeed they suggest the `from .celery_app import app as celery_app` line in `__init__.py` to be able to use shared tasks.

But are you sure it's needed even if running celery with `--app=geonode.celery_app:app` (as currently done, at least in docker-compose) ? It seems to specify the celery_app submodule already.

I'm not sure as celery is a bit a magic box for me... I'll test if it works and ask on celery's group.

lucernae:
The command you specify is for the worker.  Their entry point script is thus specified on that script.

On the other hand the import probably needed by GeoNode to load the entry point script and read celery settings. Because GeoNode is ran using wsgi or manage.py. They still need to read celery settings somehow to send the tasks to the correct broker used by the worker. 

That sums up the difference.

PR #3723, opened 2018-04-05
Fix QGIS Server settings sample

lucernae:
Previously, there are some errors detected in the sample settings file. This could lead to decreased coverage and unittests on QGIS Server side.

This PR contains the following changes:

- Disable QGIS Server Backend by default. User needs to enable it manually
- Update local_settings.py.qgis.sample
- Include static folder in .dockerignore

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,9 @@
 geonode/uploaded
 geonode/qgis_layer
 geonode/qgis_tiles
+geonode/static_root
+geonode/static/.components
+geonode/static/node_modules
 docs
 .coverage
 .celerybeat-*
--- a/geonode/local_settings.py.qgis.sample
+++ b/geonode/local_settings.py.qgis.sample
@@ -29,10 +29,15 @@ try:
 except ImportError:
     pass
 
-# GEONODE_APPS += ("geonode.qgis_server", )
+qgis_server_apps_package = "geonode.qgis_server"
+
 GEONODE_APPS = list(GEONODE_APPS)
-# INSTALLED_APPS += ("geonode.qgis_server", )
+if qgis_server_apps_package not in GEONODE_APPS:
+    GEONODE_APPS += ("geonode.qgis_server", )
+
 INSTALLED_APPS = list(INSTALLED_APPS)
+if qgis_server_apps_package not in INSTALLED_APPS:
+    INSTALLED_APPS += ("geonode.qgis_server", )
 
 # QGIS Server Backend settings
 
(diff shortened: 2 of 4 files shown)


.dockerignore
```diff
@@ -1,6 +1,9 @@
 geonode/uploaded
 geonode/qgis_layer
 geonode/qgis_tiles
+geonode/static_root
+geonode/static/.components
+geonode/static/node_modules
```
lucernae: Slightly improved docker build by excluding static development folders.

geonode/local_settings.py.qgis.sample
```diff
@@ -29,10 +29,15 @@ try:
 except ImportError:
     pass
 
-# GEONODE_APPS += ("geonode.qgis_server", )
+qgis_server_apps_package = "geonode.qgis_server"
+
 GEONODE_APPS = list(GEONODE_APPS)
-# INSTALLED_APPS += ("geonode.qgis_server", )
+if qgis_server_apps_package not in GEONODE_APPS:
```
lucernae: We now only include `geonode.qgis_server` if it is not included in  modified settings file.

geonode/local_settings.py.qgis.sample
```diff
@@ -44,14 +49,16 @@ if 'geonode.geoserver' in INSTALLED_APPS:
 if 'geonode.geoserver' in GEONODE_APPS:
     GEONODE_APPS.remove("geonode.geoserver")
 
-if 'LOCAL_GEOSERVER' in locals():
-    MAP_BASELAYERS.remove(LOCAL_GEOSERVER)
-    del LOCAL_GEOSERVER
-
-if 'PUBLIC_GEOSERVER' in locals():
+if 'PUBLIC_GEOSERVER' in locals() and \
```
lucernae: Fix syntax of excluding Geoserver basemap if we use QGIS Server backend (because GeoServer doesn't exists)

geonode/settings.py
```diff
@@ -298,7 +298,8 @@
     'geonode.services',
 
     # QGIS Server Apps
-    'geonode.qgis_server',
+    # Only enable this if using QGIS Server
+    # 'geonode.qgis_server',
```
lucernae: I suggested we disabled this by default. GeoNode uses GeoServer by default and I personally never tested both apps enabled at the same time. So I disable this to avoid problems with GeoServer.

lucernae: By the way the package will still be tested in Travis on the QGIS Server matrix environment. The apps is included via local_settings.py.qgis.sample

PR #3750, opened 2018-04-11
get geometry type for layers

ahmednoureldeen:


ahmednoureldeen:
after check , test layers has no attributes , also raster layers 
I think we need to add some test data for attributes.

Issue #3754, opened 2018-04-13
Geonode 2.8 on Ubuntu 16.04


t-book:
Hi,

http://docs.geonode.org/en/master/tutorials/install_and_admin/geonode_install/install_geonode_application.html
is speaking about installing geonode 2.8 but startproject refers to 2.9.x which ends in a 404.

`$ django-admin.py startproject --template=https://github.com/GeoNode/geonode-project/archive/2.9.x-rev.zip -e py, rst,json,yml my_geonode`

I would expect that documentation is saying: 
a) Install geonode latest on Ubuntu 16.04 or 
b) consequently stay with 2.8

`$ django-admin.py startproject --template=https://github.com/GeoNode/geonode-project/archive/2.8.0.zip -e py, rst,json,yml my_geonode`

( By the way. Who installs geonode on Ubuntu Desktop as docs explaining? Ubuntu server is much more suitable and mostly used in my opinion ... )

PR #3756, opened 2018-04-15
#3754: corrected geonode Version. Fixed tomcat reference

t-book:
This PR is based  #3754 
- apt install: libgdal-dev has been set twice 
- the documentation title says install **geonode 2.8** but as master is used one will currently end with 2.9 which can be missleading. I would suggest to  consistently stay with 2.8 (stable)
- I´ve extended the pip part comments for  installation of pygdal as this might be a hard step for not advanced users. as the user already opens requirements.txt I think it´s easiest to set the correct pygdal version in this step as well. As alternative I´m using [this](https://gist.github.com/t-book/ccc2f8d6e2c3b735e76c9b779152ee39) with vagrant but think it´s better to keep things simple and understandable for users in docs.
- the documentation is saying tomcat7 restart. which should be tomcat8 restart

that´s it.

cheers,

toni

### DIFF
--- a/docs/tutorials/install_and_admin/geonode_install/install_geonode_application.txt
+++ b/docs/tutorials/install_and_admin/geonode_install/install_geonode_application.txt
@@ -38,7 +38,7 @@ packages
 
     $ sudo apt-get update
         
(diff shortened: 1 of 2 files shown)


afabiani:
@t-book tanks very much for your contribution

PR #3764, opened 2018-04-19
populate service.last_check in monitoring in autoconfigure

cezio:
This will populate `last_check` fields in autoconfiguration. Otherwise, collector will not collect data.

### DIFF
--- a/geonode/contrib/monitoring/models.py
+++ b/geonode/contrib/monitoring/models.py
@@ -1595,6 +1595,7 @@ def do_autoconfigure():
                 name=geonode[0],
                 url=geonode[1],
                 host=host,
+                last_check=datetime.now(),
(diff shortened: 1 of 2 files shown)


geonode/contrib/monitoring/models.py
```diff
@@ -1595,6 +1595,7 @@ def do_autoconfigure():
                 name=geonode[0],
                 url=geonode[1],
                 host=host,
+                last_check=datetime.now(),
```
afabiani: We must use 

from django.utils.timezone import now

in order to take timezone into account and not

datetime.now

cezio:
@afabiani I've removed explicit `last_check` population and use default value (`auto_now_add=True`) in model instead.

Issue #3227, opened 2017-08-23
setup.py has hardcoded versions


ingenieroariel:
This makes it harder to use GeoNode with some depedencies coming from a PPA and others coming from PIP.

I believe setup.py should instead only limit minimum or maximum versions that come with incompatibilities. For reproducible builds we can already specify `==` on requirements.txt so there is no needed to make setup.py more restrictive than it should.

/cc @kalxas @afabiani @francbartoli @simod

kalxas:
I agree with @ingenieroariel 
setup.py should not have strict version numbers

simod:
Yes, agree.

capooti:
I am not sure about this. Please remember me the reason why we are using ppa packages vs pypi ones in debian setup

afabiani:
@ingenieroariel @kalxas @simod @capooti @francbartoli 

Accordingly to the issue statement, can we start from here
https://github.com/GeoNode/geonode/pull/3240

can you please take a look?

cezio:
Could we use one single source point for dependency? It can be `requirements.txt`, setup.py can just read it, parse to list and put to `install_requires`. 

As for strict versions, it could be relaxed, especially for patchlevel versions.

ingenieroariel:
The reason there are two is because we need one for reproducible builds
(known configuration that works) with exact version numbers in order to be
able to do releases. Those exact packages can live in a text file (called
requirements.txt, and used by pip directly) or can live in a PPA as debian
packages to simplify installation.

The other use case is running GeoNode as a dependency in a bigger system.
For example downstream projects, in this case you want to be able to use
GeoNode which whatever versions you are using downstream (an older
pinax-theme because you liked it more or a newer version of a package
because it fixes an issue you had)l In that case, when you do pip install
geonode or setup.py install GeoNode should not go and try to download the
latest versions of everything if stuff if the packages are already
available.

What this points to is a setup.py that just lists the package names that
are strictly needed to be able to run GeoNode, not including the
dependencies of contrib modules and only excludes major versions that are
known to be incompatible because there was a breaking change. And a
requirements.txt file that is super strict and we can use to make sure we
are all testing the same thing during the release cycles.

ingenieroariel:
One good example is the one below. GeoNode will work with literally any
paver version for the past 7 seven years, but setup.py insists in
downloading a particular version (not higher, not lower). This is
unnecessary and potentially problematic if you need to use a particular
[...]

PR #3766, opened 2018-04-20
use same deps for requirements.txt and setup.txt

olivierdalang:
[please review carefully as opinion is discutable]

Hi !

As discussed in https://github.com/GeoNode/geonode/issues/3227#issuecomment-333076257 this PR makes setup.py use requirements.txt, so that we have only once source of truth for dependencies.

This means you'd install exactly the same dependencies whether you `pip install geonode` or you install `pip install -r requirements.txt`.

I think this would help us curate our dependencies in a better way (in the past, `pip install geonode` was installing incompatible versions very often, sometimes silently as you don't necessarily get the exceptions on startup, which is particularly disruptive).

As raised by @ingenieroariel, this has the main drawback that we wouldn't have a pinned requirements file anymore for the releases. If that's needed, maybe we would need to do a pip freeze only as part of the release process ? (but I'm not familiar at all with the release process)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -176,6 +176,7 @@ Faker<=0.8.4
 more-itertools==4.1.0
 parse==1.8.2
 parse-type==0.4.1
+poster~=0.8.1
 py==1.5.3
 pytest==3.5.0
 pytest-bdd==2.20.0


setup.py
```diff
@@ -23,11 +23,15 @@
...
+REQUIREMENTS = open('requirements.txt').read().splitlines()
+
+# Hack for read the docs (TODO : explain why this is needed)
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 if on_rtd:
-    shapely_dep = "Shapely<1.5.13"
-else:
-    shapely_dep = "Shapely>=1.5.17,<1.6.dev0"
+    for i, line in enumerate(REQUIREMENTS):
+        if "Shapely>=1.5.17,<1.6.dev0" in line:
```
cezio: Why this special case?

afabiani:
@olivierdalang agree on the rationale. I will review and test your proposal. For the time being it is not passing travis tests.

setup.py
```diff
@@ -23,11 +23,15 @@
 
 from setuptools import find_packages
 
+
+REQUIREMENTS = open('requirements.txt').read().splitlines()
```
cezio: I'd suggest using more strict parsing here. While base contents of `requirements.txt` is similar to what `install_requires` accepts, file contains also comments (and other elements). You can extract requirements with `pip.req.parse_requirements`: 
https://github.com/pypa/pip/blob/master/src/pip/_internal/req/req_file.py#L60

setup.py
giohappy: This was a fix introduced because readthedocs cannot build the Shapely -> GEOS libs.
It's not required anymore because we introduced the dedicated requirements_docs.txt for RTD.

giohappy: https://github.com/GeoNode/geonode/pull/3767

olivierdalang: Thanks, I removed it

setup.py
olivierdalang: Ok done

olivierdalang:
Ok I made the changes (use pip.req.parse_requirements and removed special case for rtd).

There's still one failing test, which I think is highlighting some deps problem. I'll try to fix it.

PR #3774, opened 2018-04-23
 - Fix reproj issue on bbox_to_projection

afabiani:


capooti:
thanks @afabiani for fixing this, I was almost writing the same code here :)

PR #3849, opened 2018-07-19
Change environment variable name for database user

lucernae:
I assume this is originally a typo?
I appended `USER` for database user environment.
This way, dbname and user doesn't have to be the same.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -125,7 +125,7 @@ def _container_exposed_port(component, instname):
 
 
 def _update_db_connstring():
-    user = os.getenv('GEONODE_DATABASE', 'geonode')
+    user = os.getenv('GEONODE_DATABASE_USER', 'geonode')
     pwd = os.getenv('GEONODE_DATABASE_PASSWORD', 'geonode')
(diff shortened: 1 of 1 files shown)


giohappy:
hi @lucernae it isn't a typo. It wa a choice to name the DB users as the DB names. If a new env var is introduced it must be reflected in other places too (docker-compose, settings, etc.).
The connection and the user are kept private inside the service stack created by docker. IMHO we can keep it as it was wihtout introducing new vars.

lucernae:
I see we can close this then.
Basically if we create docker orchestration, we have to make sure DB name and user is the same.

giohappy:
It is the way the PostGIS Docker creates things. Look at here: https://github.com/GeoNode/postgis-docker/blob/master/initdb-geonode.sh

lucernae:
Yes, that's ok. We can easily change postgis user in our orchestration. I was unsure about this, so I create this PR so it can be easily reviewed. I'm happy to follow along on my side. :+1:

PR #3867, opened 2018-07-31
Update GeoNode with latest Worldmap developments

capooti:


capooti:
@afabiani I will align this to 2.10 as soon as possible. The fact is that we are running two instances using 2.7.x, that is why I am primarily working on that.

PR #3875, opened 2018-08-23
Mitigate name clashes for WMS service names

tschorr:
WMS service names have to be unique, however they are determined from the `netloc` part of the service URL only and are restricted to 40 characters (the field definition in `services.models.Service` allows for 255 characters).

This can lead to name clashes if e.g. multiple services are available on the same network location. Examples:

http://www.lfu.bayern.de/gdi/wms/wasser/abwasser?request=GetCapabilities&service=WMS
http://www.lfu.bayern.de/gdi/wms/wasser/wsg?service=wms&request=GetCapabilities

are different WMS services using the same network location.

Suggestion:

Use the entire service URL for name determination and use all available 255 characters to avoid name clashes.

### DIFF
--- a/geonode/services/serviceprocessors/wms.py
+++ b/geonode/services/serviceprocessors/wms.py
@@ -115,7 +115,7 @@ class WmsServiceHandler(base.ServiceHandlerBase,
             INDEXED if self._offers_geonode_projection() else CASCADED)
         # self.url = self.parsed_service.url
         # TODO: Check if the name already esists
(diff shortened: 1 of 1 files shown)


tschorr:
I have tried to sign the contributor agreement linked [here](https://github.com/GeoNode/geonode/blob/master/CONTRIBUTING) but keep getting a page error.

tschorr:
@afabiani just tried again using your link - still seeing an error page "We're sorry but something went wrong" when trying to submit the form after logging in with Github.

PR #3909, opened 2018-09-09
Manage featured layer in admin

ahmednoureldeen:


francbartoli:
@ahmednoureldeen the new rules from PSC provide that you should open this against `master` and then backport to `2.7.x`. Possibly with a short description of the PR or by referencing an issue

Issue #3872, opened 2018-08-15
Theme feature implementation issue


olivierdalang:
Hi,

The way the new theming feature is implemented is going to cause a lot of trouble.

Firstly, two concepts are mixed up :
1/ The ability to customize some elements of the homepage (e.g. change the jumbotron message, contact information, sponsor logos, etc.)
2/ The ability to enable/disable specific templates by the user.

But most importantly, the implementation looks extremely fragile to me :
- overwriting source files (static or templates) dynamically creates a huge mess, because it's basically modifying the code itself
- I think it's very unexpected to have collectstatic called from the code, and am pretty sure there's quite a few scenarios where this could cause a lot of trouble (e.g. when static files are hosted on a CDN).
- static files (such as css) should be served with a relatively long cache header (since they are supposedly... static !), so that changing a theme would require all users to empty their cache.

Also, I completely fail to understand how this approach is better than regular overriding Django templates/staticfiles, since anyway you need to add files to the filesystem to add new themes.

I suggest reverting completely that feature an rethink it's implementation.

1/ Should be very simple, as it's just a matter of storing some values in the database, and display them in the template (more or less Django 101). I think this is definitely a very useful feature, as it removes the need for a web developer to just make the homepage fit a project.

2/ Is probably more complex, and the need for it besides regular Django template/staticfiles overriding should be reconsidered. (also see https://github.com/GeoNode/geonode/issues/3520 which may make this slightly easier). If we really feel it's worth it (I don't), we should probably check existing Django theming apps to make sure we don't reinvent the wheel.

Cheers,

Olivier

t-book:
> 1/  ... I think this is definitely a very useful feature

> 2/ ...  If we really feel it's worth it (I don't), we should probably check existing Django theming apps to make sure we don't reinvent the wheel.

+1

afabiani:
-1

@olivierdalang I do not agree with the sentences:

1. Using embedded theming is not mandatory and it should be used only if not using a custom GeoNode project

2. Putting variables on the templates from the database would be even worse than having a fragile solution based on CSS

The concept here is exactly to avoid filling the templates with a lot of hardcoded variables.
I would suggest otherwise to make it an extension instead and enable it only if/when needed as a compromise, but I am completely against putting db variables on the templates.

davekennewell:
Hi folks, 

I'm a little lost as to what the proposed changes are here... is there a GNIP or PR you can share that would help me understand what is being proposed? 

thanks!
Dave

[...]

PR #3910, opened 2018-09-10
Fix for #3872

olivierdalang:
Hi,

Here's the implementation for customization I suggest instead of the one that's in master (fix for #3872).

It's not ready for merge, but I'd like to know if this approach is preferred to the current implementation before spending more time on this.

Still to do :
- Make the categories dynamic as well (take from DB rather than hardcoded, most of the portals don't use all of the categories), and probably also some other elements (call to action button link, etc.)
- Move the static CSS bits (for partners, contact, copyright...) to regular external CSS
- Improve display of site description
- Improve admin (e.g. order/description of fields, etc.)
- Update tests
- Document


afabiani:
In general, I'm not again this proposal.
Did not tested locally, but I'm not fully sure it will work as expected like it is now.
The PR removes the hooksets system, so I guess this must be fixed before moving on.

olivierdalang:
@afabiani 
Yes of course the intention was not to remove the hookset system... I'll restore it. (by the way, the logic for the hooksets could probably also be significantly simplified using normal django apps staticfile overrides)

Just to finalize the PR, I feel like it would be cleaner to move the customization code to another app (as it really has no relation with the hooksets). What do you think ? (this would mean a new migration)

Cheers,

Olivier

afabiani:
@olivierdalang yes, I guess moving it to an external app would be the best option. I'll try to craft a PR with a proposal in order to speed up things, if I have time.

olivierdalang:
I can do it but will be out of office next week, so it would be in two weeks... Assign the issue to me if this can wait.

afabiani:
@olivierdalang no worries, it is fairly easy to do that.

Please take a look at https://github.com/GeoNode/geonode/pull/3941

olivierdalang:
@afabiani Ok so I made the changes we talked about (move to another app, keep client hooksets)...

I didn't redo the `SeparateDatabaseAndState` trick but rather used normal migrations for the sake of simplicity/robustness, this means that existing data in the themes will be dropped.

I also added some other small improvements :
- add ability to customize call to action button
- document fields in the admin
- use CharField instead of TextField where appropriate
- removed site description (use welcome_content instead) as it conflicts with search box
- mentioned the feature in the doc

There are no tests so far, but not sure if tests would be useful, as there's really no complex logic...

It's all still a bit crude, but at least it won't cause trouble and we'll be able to enhance over time.

What do you think ?

afabiani:
@olivierdalang at a first glance it looks good. Give me some time to test it locally.

In the meantime, could you please fix formatting and Travis issues?

Also, about testing. Let me check the code in order to see if there's the possibility of regressions. If not we can skip test cases, even though in general I would prefer to have them. Just today I faced a regression due to the fact that there wasn't a test case for a particular use case.

In any case, let's try to not lower down the test coverage (which is still very low). Even if we don't add test cases, because we think they are not needed at all, we need to declare the app into the "ignore list" of codecov plugin.

I can help to refine the PR if you don't have enough time/resources to do it.

olivierdalang:
@afabiani Can you have a look at the proposed tests ?

I'll try to fix travis (I don't have a good setup to run tests locally currently so I'm counting on travis, hence very slow iterations...)

geonode/static/geonode/css/base.css
```diff
@@ -7130,6 +7131,12 @@ article.tab-pane {
 .home #showcase p {
   font-size: 18px;
 }
+.home .partner_logo {
```
afabiani: Remote .home from the partner_logo class prefix so that it applies globally

geonode/templates/base.html
```diff
@@ -29,7 +29,68 @@
...
+
+        {% if custom_theme.navbar_color %}
+        .home .navbar-inverse , .navbar-inverse {
+          background-color: {{ custom_theme.navbar_color }};
+        }
+        {% endif %}
+
+        {% if custom_theme.logo %}
+        .navbar-brand {
+          background-image: url({{ custom_theme.logo.url }});
```
afabiani: add to .navbar-brand also the following css in order to correctly fit the logo on the nav-bar

```
background-repeat: no-repeat;
background-position: center;
background-size: contain;
```

geonode/client/tests.py
```diff
@@ -1,111 +1 @@
...
-                value = base_html.read()
-                theme_regexp = re.compile(theme_html_template_regexp.format(theme.theme_uuid))
-                self.assertIsNone(theme_regexp.search(value))
-                base_html.close()
-        finally:
-            theme.delete()
-
-            self.assertIsNotNone(theme_css)
-            self.assertFalse(os.path.isfile(theme_css))
+# placeholder as paver task expects a tests.py module for each app
```
afabiani: Please add the copyright headers

```
# -*- coding: utf-8 -*-
#
# Copyright (C) 2018 OSGeo
#
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program. If not, see <http://www.gnu.org/licenses/>.
#
```

afabiani:
@olivierdalang I took the occasion to test the PR. It looks good, except for a few minor things I have put on the review.

Thanks.

geonode/static/geonode/css/base.css
olivierdalang: Ok

geonode/templates/base.html
olivierdalang: `base.less` has this already :
```
.navbar-brand {
	background: url(../img/logo.png) 0 10px no-repeat;
	background-size: 100%;
	text-indent: -9999px;
	height: 65px;
	width: 200px;
}
```
So we should be good... Or did you have some logo that don't fit properly ? (on my tests, logo displays correctly)

geonode/client/tests.py
olivierdalang: Ok

geonode/templates/base.html
olivierdalang: Sorry just tested again, you're right.

olivierdalang:
@afabiani Thanks for the review ! I made the changes you requested, if it's ok you can approve and I'll squash and merge.

afabiani:
Thanks @olivierdalang

PR #3911, opened 2018-09-11
[WIP] Add kubernetes manifest files

eggshell:
This will add a directory named deployment which contains all of the
necessary kubernetes manifest files to deploy onto a user's
managed kubernetes platform of choice. These have been written to
be platform-agnostic, but have not yet been tested for use with
Minikube.

Documentation will need to be added, but I want to get opinions
and have discussions with the geonode community before moving
further. I have a running deployment using these configs here:

https://eggshell.us-south.containers.appdomain.cloud/

Signed-off-by: Cullen Taylor <mctaylor@us.ibm.com>

### DIFF
--- a/scripts/kubernetes/local/celery-deployment.yaml
+++ b/scripts/kubernetes/local/celery-deployment.yaml
@@ -1,0 +1,107 @@
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: celery
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: celery
+    spec:
+      containers:
+      - args:
+        - celery
+        - worker
+        - --app=geonode.celery_app:app
+        - -B
+        - -l
+        - INFO
+        env:
+        - name: ALLOWED_HOSTS
+          value: '[''django'']'
+        - name: ASYNC_SIGNALS
+          value: "False"
+        - name: BROKER_URL
+          value: amqp://guest:guest@rabbitmq:5672
+        - name: CELERY_CMD
+          value: celery worker --app=geonode.celery_app:app --broker=amqp://guest:guest@rabbitmq:5672/
+            -B -l INFO
+        - name: C_FORCE_ROOT
+          value: "1"
+        - name: DATABASE_URL
+          value: postgres://postgres:postgres@db:5432/postgres
+        - name: GEODATABASE_URL
+          value: postgis://geonode:geonode@db:5432/geonode_data
+        - name: DEFAULT_BACKEND_DATASTORE
+          value: datastore
+        - name: DJANGO_SETTINGS_MODULE
+          value: geonode.settings
+        - name: DOCKER_ENV
+          value: production
+        - name: GEONODE_DATABASE
+          value: geonode
+        - name: GEONODE_DATABASE_PASSWORD
+          value: geonode
+        - name: GEONODE_GEODATABASE
+          value: geonode_data
+        - name: GEONODE_GEODATABASE_PASSWORD
+          value: geonode_data
+        - name: GEONODE_INSTANCE_NAME
+          value: geonode
+        - name: GEONODE_LB_HOST_IP
+        - name: GEONODE_LB_PORT
+        - name: GEOSERVER_LOCATION
+          value: http://geonode/geoserver/
+        - name: GEOSERVER_PUBLIC_LOCATION
+          value: http://geonode/geoserver/
+        - name: IS_CELERY
+          value: "true"
+        - name: SITEURL
+          value: http://geonode/
+        image: eggshell/geonode:latest
+        imagePullPolicy: Always
+        name: celery4geonode
+        securityContext:
+          privileged: true
+        stdin: true
+        volumeMounts:
+        - name: varlibdocker
+          mountPath: /var/lib/docker
+        - name: rundind
+          mountPath: /var/run/
+      - name: dind
+        image: docker:dind
+        securityContext:
+          privileged: true
+        args:
+          - dockerd
+          - --storage-driver=overlay2
+          - -H unix:///var/run/docker.sock
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: varlibdocker
+          mountPath: /var/lib/docker
+        - name: rundind
+          mountPath: /var/run/
+      restartPolicy: Always
+      volumes:
+      - name: varlibdocker
+        emptyDir: {}
+      - name: rundind
+        hostPath:
+          path: /var/run/dind/
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: celery
+spec:
+  ports:
+  - port: 8000
+    protocol: TCP
+    name: celery
+  selector:
+    app: celery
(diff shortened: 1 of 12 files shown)


eggshell:
I had to make some minor modifications to a few of the docker images in order to get this to work. I have opened pull requests in those repos to be thorough:

https://github.com/GeoNode/geonode-docker/pull/13
https://github.com/GeoNode/geoserver-docker/pull/43

afabiani:
@eggshell nice, very interesting.

Can you please create an issue with label "feature" associated to this PR describing better what this PR does?

Also, if feasible, would be cool to think to some integration testing. Something similar to what we do on .travis.yml when starting the docker-compose.
(not an expert of kubernetes, so just guessing here)

eggshell:
@afabiani thanks for your time! I will work on a writeup and post the issue here once it is submitted.

re: integration testing. I haven't personally done a kube deploy in travis, but it seems like [Minikube in Travis is the way forward on that](https://blog.travis-ci.com/2017-10-26-running-kubernetes-on-travis-ci-with-minikube). This is nice because I want people to be able to deploy to their local minikube cluster if they don't have PaaS resources, so I'll investigate that further.

francbartoli:
@eggshell This is great! Thanks for opening it. I will definitively have a try with my local minikube. Btw I have just tried to load a common shapefile on your public instance and it fails but we can investigate further when I have it deployed locally. Maybe is it a matter of the resources given to the k8s pod?
Out of curiosity is that running on IBM Cloud kubernetes cluster?

Again I would move the folder `deployment` under a new folder `scripts/kubernetes` for now.

eggshell:
@francbartoli Thanks for the feedback and functional testing! I debugged some today and fixed the issue with uploading shapefiles where you'd get a 413 from nginx. The problem there was that there was that the kubernetes ingress controller (which sits in front of geonode's static nginx) had its `client_max_body_size` set too low, so I went ahead and upped that to 100M to match geonode's config. Still trying to figure out a way to make it persist as I manually fixed it and reloaded the ingress controller's nginx config. Will update my commit as soon as I have that hammered out.

However, shapefile uploading is now hindered by a new error:
```
Unable to obtain connection: Cannot create PoolableConnectionFactory (Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.)
```
Which seems to be tomcat/catalina related. It would seem these are not pulling in environment variables to check where the postgres service actually is, but rather assuming it is living locally, which I definitely want to help work around, but am a little lost on how to fix at the moment. If you have any ideas of where this lingering config might be, let me know!

To answer your other question, this is indeed running on an IBM Cloud kubernetes cluster. I do want to make this is as platform-agnostic as possible, so will be testing with other major providers hopefully. And of course, I do want to get this running in minikube so others can play with it.

Will also update my commit to reflect your requested filepath change. Thanks again!

deployment/celery-deployment.yaml
```diff
@@ -0,0 +1,105 @@
...
+          value: "False"
+        - name: BROKER_URL
+          value: amqp://guest:guest@rabbitmq:5672
+        - name: CELERY_CMD
+          value: celery worker --app=geonode.celery_app:app --broker=amqp://guest:guest@rabbitmq:5672/
+            -B -l INFO
+        - name: C_FORCE_ROOT
+          value: "1"
+        - name: DATABASE_URL
+          value: postgres://postgres:postgres@db:5432/postgres
```
francbartoli: I would add below:

```yaml
- name: GEODATABASE_URL
   value: postgis://geonode:geonode@db:5432/geonode_data
```

deployment/django-deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+        - name: ASYNC_SIGNALS
+          value: "False"
+        - name: BROKER_URL
+          value: amqp://guest:guest@rabbitmq:5672
+        - name: COMPOSE_HTTP_TIMEOUT
+          value: "300"
+        - name: C_FORCE_ROOT
+          value: "1"
+        - name: DATABASE_URL
+          value: postgres://postgres:postgres@db:5432/geonode
```
francbartoli: I would add below:

```yaml
- name: GEODATABASE_URL
  value: postgis://geonode:geonode@db:5432/geonode_data
```

francbartoli:
> Which seems to be tomcat/catalina related. It would seem these are not pulling in environment variables to check where the postgres service actually is, but rather assuming it is living locally, which I definitely want to help work around, but am a little lost on how to fix at the moment. If you have any ideas of where this lingering config might be, let me know!

I have added a couple of comments, I don't know why but it seems to be that the variable `GEODATABASE_URL` has been removed from .env of production. This would be a bug because in this case the django container is not connecting to the postgres service name. Could you open a ticket in case my comments will solve the issue?

deployment/django-deployment.yaml
eggshell: added

deployment/celery-deployment.yaml
eggshell: added

eggshell:
@francbartoli so I went ahead and added the `GEODATABASE_URL` environment variable to the celery and django configs, as well as to the geoserver config just to be sure, and it is getting injected into the container successfully. However, something appears to still be overwriting it back to `localhost:5432`. Fixing this required logging into the geoserver webui and changing the datastore properties.

[correct_db_config](https://i.imgur.com/uckSRN9.png)

This seems to have propagated out to `/geoserver_data/data/workspaces/geonode/geonode_data/datastore.xml`, and uploading is now possible. I am wondering if there is some sort of timing issue where `GEODATABASE_URL` is not getting set before `settings.py` runs and parses it?

francbartoli:
@eggshell I’ll look into this, just to make sure that we are not getting into a false problem: did you test from scratch after the removal of all previous docker volumes?

francbartoli:
@eggshell I'm currently trying to run the deployment on minikube. I'm assuming the command to execute are:

```yaml
kubectl apply -f scripts/kubernetes/*.yaml
```

for each yaml file. Is that correct? Does the order matter? Anything else? Thanks for providing some guidance

francbartoli:
@eggshell sweet, that definitively would help!

afabiani:
@eggshell very nice, good job. Waiting for @francbartoli feedbacks.

@olivierdalang  might also provide some feedbacks on this.

afabiani:
@eggshell @francbartoli is this PR finished? Any other request?

francbartoli:
@afabiani I would wait for some docs as we don't have any urgency with this. What do you think?

eggshell:
@afabiani @francbartoli agree on docs, I am on some planes today but will throw them up asap. a few minor changes will be incoming as well but this should be wrapping up soon. Thank you again for your time!

francbartoli:
@eggshell did you have time to add some docs?

gannebamm:
> 
> 
> @eggshell did you have time to add some docs?

Just out of curiosity: What is the current status of this PR? Are these (https://github.com/eggshell/geonode-on-k8s) the mentioned docs?

giohappy:
@francbartoli I think these types of contributions can be accepted if active maintenance is guaranteed, otherwise it becomes trash. @eggshell seems to have abandoned it so I would be in favour to close this PR.

francbartoli:
@giohappy we can close it now and as soon as I have time I will be happy to add:
[ ] remove dind
[ ] docs
[ ] deploy on minikube

and then volunteer to maintain it

PR #3930, opened 2018-09-24
[Fixes #3929] Show documents also on Groups Activity Tabs

### LINKED ISSUE #3929: Show "documents" also on Groups Activity Tabs
Right now only Layers and Maps are shown on the Groups Activity Tab

afabiani:
Fixes #3929

### DIFF
--- a/geonode/groups/views.py
+++ b/geonode/groups/views.py
@@ -137,6 +137,7 @@ class GroupDetailView(ListView):
         context['object'] = self.group
         context['maps'] = self.group.resources(resource_type='map')
         context['layers'] = self.group.resources(resource_type='layer')
+        context['documents'] = self.group.resources(resource_type='document')
         context['is_member'] = self.group.user_is_member(self.request.user)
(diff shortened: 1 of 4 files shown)


geonode/groups/views.py
```diff
@@ -275,6 +276,14 @@ def getKey(action):
             for action in actions
             if action.action_object and action.action_object.group == self.group.group][:15]
         action_list.extend(context['action_list_maps'])
+        actions = Action.objects.filter(
+            public=True,
+            action_object_content_type__model='document')[:15]
+        context['action_list_documents'] = [
+            action
+            for action in actions
+            if action.action_object and action.action_object.group == self.group.group][:15]
```
olivierdalang: Shouldn't you filter in the queryset instead ? (not sure, very quick review without looking in depth what the change is about)

afabiani: Nope, in this case, the objects have been stored on "actions" objects. We are basically trying to display all the activities on documents related to a specific group.

PR #3936, opened 2018-09-25
[Closes #3935] celery.log file based handler causing issues if not correctly configured

### LINKED ISSUE #3935: celery.log file based handler causing issues if not correctly configured
local_settings samples (some of them) are declaring a celery file-based log handler.

On a default installation, it might prevent the server to start due to permission denied on filesystem.

My suggestion is to remove it by default and align the handler with the other ones, based on console one.

afabiani:


afabiani:
@francbartoli given that this change is just affecting a sample settings file, the change here is just to remove logging to a file done from Django. It will still logged through the console handler to the httpd server config files. So yes, I don't see the advantage of keeping another file which can also cause issues whenever Django is not able to write it for any reason.

Issue #3800, opened 2018-05-11
Uploading shapefiles without a datefield and time-enabled is False in importer settings fails in 2.7.x


ewsterrenburg:
Uploading shapefiles without a datefield and time-enabled is False in importer settings fails in 2.7.x.
In `check_step_view`, the completed step is set to '**time**' for these layers, yet this step is removed from the `_pages` dictionary in `utils.py` if option TIME_ENABLED is set to False. Result is an unexpected error (Index not found). Code below does fix this (yet it is a bit unclear to me at the moment to which branch PRs should be made).

```
def check_step_view(request, upload_session):
    import_session = upload_session.import_session
    assert import_session is not None

    if request.method == 'GET':
        layer = check_import_session_is_valid(
            request, upload_session, import_session)
        if upload_session.completed_step != 'error':
            if not layer:
                upload_session.completed_step = 'error'
                upload_session.error_msg = 'Could not access/read the uploaded file!'
            else:
                (has_time_dim, layer_values) = \
                    layer_eligible_for_time_dimension(request,
                                                      import_session.tasks[0].layer, upload_session=upload_session)
                # if _ALLOW_TIME_STEP is False, step 'time' is removed from the available 
                # list of steps in utils.py, which results in issues uploading a layer
                # which would be eligible for time dimension
                if has_time_dim or _ALLOW_TIME_STEP is False:
                    upload_session.completed_step = 'check'
                else:
                    # This command skip completely 'time' configuration
                    upload_session.completed_step = 'time'
    elif request.method != 'POST':
        raise Exception()
    return next_step_response(request, upload_session)
```

afabiani:
@ewsterrenburg hopefully this fixes the issue

https://github.com/GeoNode/geonode/pull/3830/files

ewsterrenburg:
@afabiani 
Seems to be working correctly. Think it can be incorporated into master & 2.7.x

PR #3939, opened 2018-09-25
set to completed step to check if no _ALLOW_TIME_STEP is false

ewsterrenburg:
Related to https://github.com/GeoNode/geonode/issues/3800
 
@afabiani With the current fix, the upload is completed successfully, however the layer is created in a new shapefile store in GeoServer, not in the (Postgis) store configured in the OGC_SERVER settings. 

After changing the timestep to 'check' instead of run, the layers are imported into the correct store.

Kind regards,
Erwin


afabiani:
@ewsterrenburg I see. Let me test it under several different conditions.

We need to be sure this does not introduce some regression on the current use cases. Also, there will be the need to add more integration and upload tests.

Most probably I will refactor this PR by adding more specific test cases too. It will take some time.
If you are able to do that in the meantime, it would be great.

afabiani:
@ewsterrenburg let me give it a look today. I will be back hopefully today with more details.

I'll ask you eventually to help me testing the fixes/changes on your specific use cases.

afabiani:
@ewsterrenburg so I took a look at those changed but I don't really understand how this might affect the original workflow.

Other than this, there's a potential deadlock at line 418... that change seems like calling recursively the "check_step_view" view.

I guess the issue you are describing on the first comment, might be a side effect of another issue. Could you please update GeoServer version? Notice also that now "master" is using GeoServer 2.14.x. Old versions won't work anymore, since the GeoFence paths have been changed.

geonode/upload/views.py
```diff
@@ -415,7 +415,7 @@ def check_step_view(request, upload_session):
                     upload_session.completed_step = 'check'
                 else:
                     # This command skip completely 'time' configuration
-                    upload_session.completed_step = 'time' if _ALLOW_TIME_STEP else 'run'
+                    upload_session.completed_step = 'time' if _ALLOW_TIME_STEP else 'check'
```
afabiani: This looks like an infinite recursive loop.

ewsterrenburg: It's not an infinite recursive loop (completed step is set to 'check', so the next step will be taken), so for shape-files, the next step that will be executed would be 'run'. With the current code in master, it would say 'run' is completed already, so it would directly go to final (which might be the correct behavior). None-the-less, I don't see why the uncorrect store is being used (since this is set in the check-step, I assumed this was not executed, yet apparently it is)

I will analyse this further, thanks for your efforts

afabiani:
@ewsterrenburg I did some more tests with this, but currently it worked on my side even without this patch. Did you manage to reproduce the issue? Can we try to dig a bit more into this?

ewsterrenburg:
@afabiani 

Preliminary conclusions after some testing:
- Issue seems **only present** using the **geonode.importer backend**, using geonode.rest is working fine. importer can handle more coordinate systems (for instance, it is not possible to use srs.identify_epsg() for the Dutch national coordinate system, probably why I set it to importer anyway).
- When using importer, for shapefiles without a date field, the store is set to the name of the shapefile, not to the Geonode PostGIS datastore (but I cannot fully pinpoint yet where)
- When changing 'run' to 'check', this issue is fixed, yet I do not fully understand why.
- When changing 'run' to 'check', the rest backend is still function correctly for shapefiles (did not test for other data types yet)

afabiani:
@ewsterrenburg are you using an updated version of GeoNode? I'm asking because I currently tested locally this functionality with importer.backend and time = False, but the shapefiles are correctly imported into the DB datastore. Also, which version of GeoServer are you using? I'm asking because I remember there was a mistake on the DS naming in the importer workflow some time ago.

I will do more tests anyway, trying to create a couple of specific test cases for this on the integration tests package.

ewsterrenburg:
@afabiani 

- GeoNode straight from 2.8.x from Github
- GeoServer:
image

afabiani:
@ewsterrenburg I'm afraid that could be an issue with GeoServer, probably.
Currently, I'm still not able to reproduce the issue on GeoNode master (2.10) and GeoServer 2.14.x

While I'm going to switch to your same versions of GeoNode and GeoServer, could you try to upgrade your GeoServer versions with the latest 2.13.x war built yesterday?

https://build.geo-solutions.it/geonode/geoserver/latest/

You don't need to touch anything of your current GEOSERVER_DATA_DIR, just replace the GeoServer binaries with the latest ones. Probably Importer some issues recently fixed on GeoServer 2.14.x have been backported to GeoServer 2.13.x since May 2018.

afabiani:
@ewsterrenburg I confirm the behavior with a fresh geonode 2.8.x setup.

Still I'm not sure what's different, the problem seems to be related to the importer session returned by GeoServer. Will investigate a bit more on this. If I won't be able to find a suitable solution, your patch could work anyway.

ewsterrenburg:
@afabiani Thanks again (it's getting repetitive :). 

It could very well be related to the importer session (I wasn't able to find in the GeoNode code where the wrong store was set yet).

I tried to update GeoServer today, yet before I can plug it in our environment, requires some additional efforts (I wasn't able to include both Postgis support for geofence & S3 support for Geowebcache without compiling it myself, and apparently I have forgotten how to do this.

PR #3941, opened 2018-09-27
[Fixes #3872] - Moving out themes stuff on a contrib app

### LINKED ISSUE #3872: Theme feature implementation issue

afabiani:
As discussed on https://github.com/GeoNode/geonode/pull/3910 this PR aims to move on a contrib app the theme customization stuff


afabiani:
> Are you sure about editing the migrations files directly ? The stale tables won't be deleted this way...
> 
> It would have been safer just to let makemigrations drop the tables (or even better to do it like this : https://stackoverflow.com/a/26472482 so that if anyone has used that feature already he doesn't loose data)

@olivierdalang I pushed the migrations changes as you requested

olivierdalang:
Maybe I'm confused but not sure what you're doing with migrations is correct.

I suggest to just move the models.py, without do anything manually about previous migrations, and run makemigrations, and that's it. This will just delete the table in the client app and create it in the new app. If you change previous migrations like you're doing, you run the risk of creating inconsistent state which is a huuuge pain to solve.

If you want for users not to loose data (if they used the feature already), just look at the link I pasted. Note that this solution doesn't touch previous migrations either.

I know this code was not formally released yet, so theoretically, no one is supposed to have run those migrations on production, but still, if I'm not mistaken we've had some release candidates containing this code, so I think we should play safe. I'm not sure about best practice here, but I think a migration that has landed in master should be considered permanent.

afabiani:
@olivierdalang I see, sorry, I didn't fully get your point. Nevertheless, I'm not convinced about this solution. This will leave dirty stuff on core which should stay in a contrib app instead.

Since those tables just store themes information, I don't think it would be a big deal to create a new one from scratch. But since you requested this change, if you think is better to leave the old model on client I will make changes accordingly.

afabiani:
@olivierdalang by the way, I just did some tests with the SeparateDatabaseAndState migration and this is the outcome:

django.db.migrations.exceptions.InconsistentMigrationHistory: Migration geonode_themes.0001_initial is applied before its dependency geonode_client.0005_auto_20180927_1015 on database 'default'.

afabiani:
@olivierdalang I should have found a solution, please take a look
https://github.com/GeoNode/geonode/pull/3941/commits/02b0000044656d1f769da97c5627d74286d3c59b

however we still need to keep dirty migrations inside geonode.client

@francbartoli I have created a new test case checking for contrib apps into tests.smoke

olivierdalang:
> Nevertheless, I'm not convinced about this solution. This will leave dirty stuff on core which should stay in a contrib app instead.

>however we still need to keep dirty migrations inside geonode.client

I think that's perfectly fine, it's really the point of migrations to represent an obsolete state. Without this, we run the risk of creating migrations issues, which really are the worst... I think the code was out long enough so that it's pretty sure a few users have deployed it already (especially since it was in release candidates).

> I should have found a solution, please take a look

At first glance it looks good, will check again tomorrow as it's late already.

olivierdalang:
Just a few questions :

- is it actually appropriate to have this as a contrib module ? I'd be in favor of keeping it in core, as it seems to me it's a feature that would be used by almost everybody (who installs a geonode without changing at least the title and adding their logo ?)
- what about the other problems mentioned in #3872 (use template tags instead of overwriting the template files), will you integrate those in this PR or leave that for later ?

IMO, if we use normal template tags in the core index.html as I suggested in my PR, it would feel a bit wrong to have that logic in a contrib module. And if we don't put it in the core index.html (but rather override that from the new contrib module), it's going to be hard to keep it DRY/not too complex.

afabiani:
@olivierdalang 

1. Nope, I really think putting this into "contrib" apps is the best choice, given the discussion and effort we put extracting this from the core. Until now "contrib" has been used historically to include directly into GeoNode extensions and plugin. Something similar to the GeoServer "extensions" module. For the moment my feeling is that this is the best place to put this feature, which might be disabled and substituted by another one in the future (or removed completely).

2. I guess we need a GNIP with votes before moving on with further development on this. Also, I'm not fully sure just one GNIP would be sufficient since I envisage non trivial changes to the templating system of GeoNode.

francbartoli:
+1 for moving into contrib apps.

olivierdalang:
@afabiani If we don't address 2. (which is what #3872 is about, it was not about moving to contrib), I think we should remove the feature altogether.

It will mess with your git (as it changes source files). I think it would be buggy in a dockerized context (since stateless, changes to the source are not permanent), in load balanced settings, with shared libraries (where you may not have the permission to change the source, and if you did, it could affect other sites), etc..... Moving it to contrib doesn't fix these, just makes it less probably that it will be used.

Removing the feature would be a shame, you seemed to agree with the approach I was suggesting in the PR, and I really think the feature would be super-useful for many users. I still think it should stay in core as it could be one of the nicest feature of next release in terms of user experience.

Also I was (and still am) willing to do the work (in 2 weeks) to have this mergeable, and don't think we need a GNIP on this as it's not a new feature but implementation detail.

afabiani:
@olivierdalang so what with this? Should we close it? Nevertheless *this* is the correct way to remove *themes* stuff from geonode.client, without breaking the hooksets. If you are planning to do another PR or updating the WIP one, please follow the same approach (except for the contrib module of course).

olivierdalang:
If you're sure about the contrib app, you can merge this, and I'll redo the changes on the new app.

The only thing I'm not happy with in the original proposal is that it requires changes in index.html (core), which seems a bit wrong for a contrib module (which should be more or less self-contained). Of course it's always possible to override index.html, but that's will be less DRY.

afabiani:
@olivierdalang as far as I understood this would be something that will be completely removed when your PR will be ready, right?

I'd suggest to just close this one and wait for your PR if it is ok with you. Or, alternatively, move this module completely outside GeoNode. The issue I envisage is that by refactoring the themes functionality will let this incompatible and might create nontrivial issues to people that started using it already.

The best approach would be to do a PR which is somehow backward compatible with this approach, but it does not seem realistic to me; at least this is my feeling.

olivierdalang:
> @olivierdalang as far as I understood this would be something that will be completely removed when your PR will be ready, right?

No my PR would include some changes on core templates (e.g. index.html or base.html, look at the PR). As said, it's not really possible to achieve exactly this while keeping the template DRY (as theme elements appear a bit everywhere in the page).
As said, I'd be in favor to keep it in core. But still OK if you insist to have it in contrib, in which case I'd be for non-DRY template override...

afabiani:
@olivierdalang nope, it is ok for me to keep in core too. So I'm going to close this one.

PR #3998, opened 2018-10-16
PR with Worldmap changes running on master (2.10.x)

capooti:
This PR sync master with changes in worldmap branch (previously we were running it on top of 2.8.x).


afabiani:
@capooti doh! This will mess up a bit the history :) 

250 commits !!!

capooti:
squash the commits man! :D

capooti:
it is always a good idea to squash. In this case even more because many commits comments refer to ticket open and closed in our repo

PR #4008, opened 2018-10-16
[DRAFT] - Tentative fix 2.6.3

francbartoli:


afabiani:
@francbartoli is this still valid? What was the intent here?

francbartoli:
Sorry, I had forgotten it @afabiani . The tentative to fix the 2.6.3 version was good locally at that time but travis didn't pass, actually, I don't remember the reason.
I believe we can close it and see if it might be interesting for someone who has to stay with that version.

PR #4023, opened 2018-10-24
Manage worldmap client using client hooksets standard approach. Refs #4019

capooti:


geonode/settings.py
```diff
@@ -1234,6 +1234,12 @@
 )
 """
 
+# To enable the WorldMap based Client enable those
+GEONODE_CLIENT_HOOKSET = "geonode.client.hooksets.WorldMapHookSet"
```
afabiani: I guess this should not be enabled by default, right?

capooti: definitely. Sorry I meant to comment it out before pushing it. Will do it right now

Issue #4047, opened 2018-10-30
PR 4035 caused a number of broken/incorrect urls


capooti:
So far I have seen:
* in layers list profiles link is broken
* in maps list profiles link and "view map" link are broken

PR #4048, opened 2018-10-30
Fixes some broken links as per #4047

capooti:
{{ SITEURL }} will not evaluated as it is an angular variable in this context

### DIFF
--- a/geonode/base/templates/base/_resourcebase_snippet.html
+++ b/geonode/base/templates/base/_resourcebase_snippet.html
@@ -62,7 +62,7 @@
           <div class="row">
             <div class="col-lg-12 item-items">
               <ul class="list-inline">
-                <li><a href="{{ SITEURL }}people/profile/{{ item.owner__username }}"><i class="fa fa-user"></i>{{ item.owner_name }}</a></li>
+                <li><a href="/people/profile/{{ item.owner__username }}"><i class="fa fa-user"></i>{{ item.owner_name }}</a></li>
                 <li><a href="{{ item.detail_url }}#info"><i class="fa fa-calendar-o"></i>{{ item.date|date:'d MMM y' }}</a></li>
                 <li><a href="{{ item.detail_url }}"><i class="fa fa-eye"></i>{{ item.popular_count }}</a></li>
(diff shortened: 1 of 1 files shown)


geonode/base/templates/base/_resourcebase_snippet.html
```diff
@@ -62,7 +62,7 @@ <h4>
           <div class="row">
             <div class="col-lg-12 item-items">
               <ul class="list-inline">
-                <li><a href="{{ SITEURL }}people/profile/{{ item.owner__username }}"><i class="fa fa-user"></i>{{ item.owner_name }}</a></li>
+                <li><a href="/people/profile/{{ item.owner__username }}"><i class="fa fa-user"></i>{{ item.owner_name }}</a></li>
```
afabiani: @capooti I cannot currently reproduce the issue, could you please check your settings and context processor? On the other side, reverting this won't allow us to deploy GeoNode on URIs different from the root "/"

capooti: @afabiani I cannot figure out how it could work for you. You maybe meant to use the siteUrl javascript variable in place of SITEURL?

This is what is happening to me, using GeoNode master, if I look at a snippet in the maps list page:
* the owner link is returned in this form: http://localhost:8000/maps/people/profile/admin (should be http://localhost:8000/people/profile/admin)
* the view map link is returned in this form: http://localhost:8000/maps/maps/8/view (should be http://localhost:8000/maps/8/view)

This is happening because {{ SITEURL }} isn't evaluated.

afabiani: Oh yes, I see now. You are right, I didn't notice the different url for profile.

In that case let me fix this by introducing an angular variable instead of reverting. This is why in the case you will need to publish GeoNode on a sub-path instead of root ("/") it won't work anymore.

capooti: sounds good, and the possibility to publish geonode in a subpath is a great addition. Thanks @afabiani !

afabiani:
replaced by https://github.com/GeoNode/geonode/pull/4055

PR #4055, opened 2018-11-05
Fixes some broken links as per #4047

afabiani:


capooti:
Hi @afabiani 
why not using siteUrl which you introduced (the javascript var)? (just checking if there is a specific reason)

afabiani:
@capooti well I just thought that this approach is more reliable and robust. I wouldn't trust that much on a global JavaScript variable, which might be easily removed / overridden by a wrong template.

What do you think?

Issue #4091, opened 2018-11-20
GNIP Edit/Add data functionality


karakostis:
I am working on the migration of a contrib app to the latest GeoNode (2.8). The app is called Edit/Add Data.

The Edit Data functionality was designed and implemented in order to give the possibility to the user to:

1. View data of a layer in a tabular format.
2. Make a simply search on the data (it uses the search functionality of Datatables.js)
3. Edit the data by changing its non spatial attributes.
4. Edit the spatial data (currently works only for points). E.g. change position of a point.
5. Delete a row from the layer.
6. Add a new feature (point, line or polygon).

Below I attach some of the use cases of the app together with screenshots:

**Edit the data by changing its non spatial attributes**

1. Click the 'Edit Attributes' button.
2. Edit the data by clicking on each cell.
3. Click the 'Save' button or click 'Cancel' to cancel the changes.

![1](https://user-images.githubusercontent.com/17905221/48782134-c7464980-ecdd-11e8-8778-d7bae3235e36.png)
![2](https://user-images.githubusercontent.com/17905221/48782146-cad9d080-ecdd-11e8-9602-096abd60172a.png)

**Edit the spatial data of a layer (currently works only for point geometries)**

1. Click the 'Edit Geometry' button.
2. A new window will open with a map. To move the point click the button on the left side of the map.
3. Place the point where you want.
4. Click 'Save' (on the top left side of the map).
5. Click the 'Save' button to save all changes. To cancel the changes select the 'Close' button. 

![1](https://user-images.githubusercontent.com/17905221/48782197-ed6be980-ecdd-11e8-85d8-830e2e0c9c72.png)
![2](https://user-images.githubusercontent.com/17905221/48782201-efce4380-ecdd-11e8-80c1-04a260d5aba7.png)
![3](https://user-images.githubusercontent.com/17905221/48782206-f2c93400-ecdd-11e8-92ec-d325ca73ecfa.png)

**Add a new feature (point, line or polygon)**

1. Click the 'Add a new entry' button
2. Type the values for each field.
3. Select the 'Add Geometry' button.
4. A new window will open with a map. To add a new point, Line or Polygon click the button on the left side of the map.
5. Place the Point, Line or Polygon.
6. Click the 'Ok' button.

![1](https://user-images.githubusercontent.com/17905221/48782295-315eee80-ecde-11e8-9295-5b1924735315.png)
[...]

PR #4127, opened 2018-12-17
Edit/Modify data functionality

karakostis:
The Edit Data functionality was designed and implemented in order to give the possibility to the user to:

View data of a layer in a tabular format.
Make a simply search on the data (it uses the search functionality of Datatables.js)
Edit the data by changing its non spatial attributes.
Edit the spatial data (currently works only for points). E.g. change position of a point.
Delete a row from the layer.
Add a new feature (point, line or polygon).
Below I attach some of the use cases of the app together with screenshots:

Edit the data by changing its non spatial attributes

Click the 'Edit Attributes' button.
Edit the data by clicking on each cell.
Click the 'Save' button or click 'Cancel' to cancel the changes.

https://github.com/GeoNode/geonode/issues/4091


geonode/contrib/edit_data/templates/edit_data/wfs_add_new_line.xml
```diff
@@ -0,0 +1,21 @@
+<wfs:Transaction service="WFS" version="1.1.0"
+  xmlns:wfs="http://www.opengis.net/wfs"
+  xmlns:gml="http://www.opengis.net/gml"
+  xmlns:geonode="{{ geonode_url }}">
+  <wfs:Insert>
+    <geonode:{{ layer_name }}>
```
hishamkaram: workspace name should be retrieved from django setting(DEFAULT_WORKSPACE )

geonode/contrib/edit_data/templates/edit_data/wfs_delete_row.xml
```diff
@@ -0,0 +1,9 @@
+<wfs:Transaction service="WFS" version="1.0.0"
+  xmlns:ogc="http://www.opengis.net/ogc"
+  xmlns:wfs="http://www.opengis.net/wfs">
+  <wfs:Delete typeName="geonode:{{layer_name}}">
```
hishamkaram: the same thing for the typeName

karakostis: @hishamkaram Thanks for the comment. I have updated the code. It is not clear to me why the tests are failing. Could you please have a look as well?

karakostis:
Thanks @afabiani 

I will look at these issues. I have one question regarding **adding specific leaflet plugins libraries in the:**

1. leaflet-plugins.min.js
2. leaflet-plugins.min.css

**I follow the below steps:**

1. add the dependencies in bower
2. run "bower cache clean"
3. run "bower update"
4. add the dependencies in grunt file
5. run "grunt production"

Although I can see the updated files with the corresponding code included, the libraries are not loaded (I am currently in the dev env).

Right now I have modified the settings.py file in order to make this work but I will also need to fix this before committing:

`if not DEBUG_STATIC:
    # if not DEBUG_STATIC, use minified css and js
    LEAFLET_CONFIG['PLUGINS'] = {
        'leaflet-plugins': {
            'js': 'lib/js/leaflet-plugins.min.js',
            'css': 'lib/css/leaflet-plugins.min.css',
            'auto-include': True,
        },
        'leaflet-areaselect': {
            'css': 'lib/css/leaflet-areaselect.css',
            'js': 'lib/js/leaflet-areaselect.js',
            'auto-include': True,
        },
    }`

giohappy:
@karakostis AFAICS you didn't provide an alternative to the conditional check in layer_detail. I totally agree with @afabiani that contrib apps' stuff should never leak into the core code.
The same applies for the static assets. The contrib apps are in charge of serving them, they shouldn't be placed in the base folders.

PEP8 errors are still there.

karakostis:
@giohappy thanks for the tips. I have fixed the PEP8 errors and modify the app in order to serve the static files. Regarding what @afabiani recommends, it makes sense and I will try to do this asap although right now I haven't thought of a way. Do you have any recommendations on this?

Issue #4120, opened 2018-12-13
GeoNode 2.8 Upload of Shp Layer failing – GeoServer Error: CoordinateArraySequenceFactory


t-book:
#### Expected behavior
Successful upload of vector layer (attached)

#### What happens instead 
Upload fails with:

> Unable to usecom.vividsolutions.jts.geom.impl.CoordinateArraySequenceFactory@47999b66 to produce CoordinateSequence with dimension 2

Full trace:  https://gist.github.com/t-book/d42e5c86a0ee7f1554b77fd96f0f6ec2

#### Setup
- Ubuntu 16.04  
- GeoNode 2.8.1 rc0 installed by use of geonode-project
- Full Stack for production
- GeoServer Version: 
  - Implementation-Version: 2.13-SNAPSHOT
  - Build-Timestamp: 2018-10-17T14:26:11Z
- JVM: Oracle Corporation: 1.8.0_191 (OpenJDK 64-Bit Server VM)
- JVM renderin engine: sun.java2d.pisces.PiscesRenderingEngine

[tiger_roads.zip](https://github.com/GeoNode/geonode/files/2675945/tiger_roads.zip)

t-book:
@afabiani I´ve just tested with an older 2.13 release from Mai which is working. So this issue might fit better with GeoServer?

This is working
> Implementation-Version: 2.13-SNAPSHOT
> Build-Timestamp: 2018-05-23T16:52:21Z
> Git-Revision: 934d6bd11928e28c96a8313d37eb94afecf15d29

This does not:
>Implementation-Version: 2.13-SNAPSHOT
> Build-Timestamp: 2018-10-17T14:26:11Z
> Git-Revision: 686c780a08965f06f0ab8a33261feee00ed90f40

afabiani:
Uhm, different GeoServer with same GeoNode you mean?

t-book:
Exactly, I´ve installed latest Geoserver 2.13 which produces error. Than I´ve replaced GS with a 2.13 build from May which works without problems. Same shape file same encoding ;(

t-book:
Did something change between 2.13.1 and 2.13.3 regarding CoordinateArraySequenceFactory or in generell with importer and shp handling?

afabiani:
Maybe the version of the JTS.... need to check.

I'll try to take a look as soon as possible. In the meantime, I think it is safe to use the older one.

t-book:
As always thanks Alessio!

[...]

PR #4143, opened 2018-12-20
Restored older GeoServer Version #4120

t-book:
The latest GeoServer 2.13 build has problems with uploading of shape files. In agreement with Alessio we restored an older working build. This can remain for 2.8 branch until the problem is fixed. See #4210 for more.


afabiani:
@t-book I'm going to fix the GeoServer 2.13.x Docker image also accordingly

https://github.com/GeoNode/geoserver-docker/tree/2.13.x

PR #4148, opened 2019-01-02
Removed Windows from docs+danger message for apt

t-book:
The docs need an update! – Until somebody finds time I would suggest to at least prevent users burning time by trying apt or windows installers:

1) removed windows from readme
2) removed windows from docs
3) danger message for apt added


t-book:
@afabiani Any idea why these changes do not show up at: http://docs.geonode.org/en/master/tutorials/install_and_admin/index.html
I guess this is readthedocs with some on commit hook? 

*I´m asking as we again have a case were somebody might have fallen in the apt trap (https://lists.osgeo.org/pipermail/geonode-users/2019-January/005285.html)*

afabiani:
@t-book the docs build is failing due to

```
Could not find a version that satisfies the requirement mock==2.0.0 (from -r ./requirements_docs.txt (line 104)) (from versions: )
No matching distribution found for mock==2.0.
```

afabiani:
@t-book even pushing a small fix for this, it looks like the readthedocs trigger does not starts automatically and I'm not currently listed on the maintainers list.

Let's try to ask to @capooti or @simod if they can hit a build.

t-book:
I´m a bit puzzled with this. The error points to https://github.com/GeoNode/geonode/blob/master/requirements_docs.txt but if I´m right it´s coming from https://github.com/rtfd/readthedocs.org/blob/master/requirements/pip.txt#L81 ?
Strange mock==2.0 resolves without problems.

t-book:
@afabiani You were faster ;) Thanks for the info, let´s wait what Paolo and Simon think.

PR #4158, opened 2019-01-16
Replace exit with a valid command

frafra:
`exit 0` fails becaouse `exit` is not a command available (it is a valid shell builtin command). `true` instead can be found at `/usr/bin/true`.

```
$ man true | head -n4
TRUE(1)                                            User Commands                                           TRUE(1)

NAME
       true - do nothing, successfully
```


olivierdalang:
Thanks !

Issue #3563, opened 2018-01-15
Add status message on GeoNode session "access_token" validity


afabiani:
OAuth2 access_tokens currently have a timeout. Once their validity expires, protected layers are no more accessible by the logged in user, unless he logs out and in again.

It would be very helpful to let GeoNode check for access_token validity periodically and send a warning message to the user.

GeoNode should also provide a small button to refresh it and either generate a new one or extend its validity.

t-book:
+1 
- Addition: If last_activity < allowed_inactive_time (timeout),
 log-out user and redirect to login page with some message what happend.

capooti:
I think this is not trivial and has caused headaches to many users for one of our instance. What is the OAuth2 access_token timeout? Isn't it possible to just logout the user when the token expires?

capooti:
Instead than sending the warning, wouldn't be possible to have Django automatically request a new token if the other is expired?

afabiani:
There should be a way to refresh the tokens. Nevertheless, GeoNode should somehow open a pop-up coherently asking the user to refresh the session under a certain time IMHO.
As soon as I'm able to find some time and resource to dig more into this I'll propose a more detailed solution.

capooti:
Sound great! thanks @afabiani

capooti:
This is really causing confusion to users, as when the token epires:
* private layers become not accessible
* data editing stops working
* style editing stops working

Looking at https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#authorization-code-expire-seconds I am confused about ACCESS_TOKEN_EXPIRE_SECONDS and AUTHORIZATION_CODE_EXPIRE_SECONDS. I believe that increasing ACCESS_TOKEN_EXPIRE_SECONDS should help, while I am unsure of AUTHORIZATION_CODE_EXPIRE_SECONDS which is recommended to be set at 600 seconds.

According to this: https://github.com/jazzband/django-oauth-toolkit/issues/93
we should be able to set ACCESS_TOKEN_EXPIRE_SECONDS to a very high value to solve this issue.

Any thoughts on this?
For now I am trying to increase ACCESS_TOKEN_EXPIRE_SECONDS to a very high value and let's see if this helps.

giohappy:
Generally having an automic token refresh mechanism would be the best option. This is supported by oauth toolkit and by default REFRESH_TOKEN_EXPIRE_SECONDS is set to None, which means that the refresh token can be always used to obtain a fresh access token.
[...]

PR #4164, opened 2019-01-21
 - Making OAUTH2 Access Token expiration seconds configurable by settings/env

afabiani:
Ref to #3563

### DIFF
--- a/geonode/base/models.py
+++ b/geonode/base/models.py
@@ -1464,12 +1464,16 @@ def do_login(sender, user, request, **kwargs):
             # Lets create a new one
             token = generate_token()
 
+            # 1 day expiration time by default
+            _expire_seconds = getattr(settings, 'ACCESS_TOKEN_EXPIRE_SECONDS', 86400)
+
+            # Let's create the new AUTH TOKEN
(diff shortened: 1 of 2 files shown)


geonode/base/models.py
```diff
@@ -1464,12 +1464,16 @@ def do_login(sender, user, request, **kwargs):
...
+            _expire_seconds = getattr(settings, 'ACCESS_TOKEN_EXPIRE_SECONDS', 86400)
+
+            # Let's create the new AUTH TOKEN
             AccessToken.objects.get_or_create(
                 user=user,
                 application=app,
                 expires=datetime.datetime.now(timezone.get_current_timezone()) +
                 datetime.timedelta(
-                    days=1),
+                    seconds=_expire_seconds),
```
capooti: oh now I see why changin ACCESS_TOKEN_EXPIRE_SECONDS didn't work for me (ref https://github.com/cga-harvard/geonode/issues/583#issuecomment-455326558). Thanks for pointing it out

capooti:
thanks this should fix the problem, I will check it out.

PR #4172, opened 2019-01-23
configure if links should be available inside the layers list from API

giohappy:


giohappy:
Integration test must be changed accordingly. I'll fix it ASAP.

PR #4185, opened 2019-01-25
[Closes GeoNode#4183] Fix and improve LOCKDOWN_GEONODE mechanism

afabiani:


afabiani:
> I would add /api/layers to the exempted URLs

@giohappy  done

PR #4190, opened 2019-01-29
fixed: Anyone permission stores as an empty string instead of * at GeoFence Data Rules

khaledboka:


khaledboka:
Thanks 
I will give it another try with another GeoServer WAR

afabiani:
@khaledboka I'm going to close this PR since this is not valid.

Please, feel free to keep the discussion live on the mailing list and eventually create issues on the github repo. After having more clear where the problem is, we can envisage a good PR.

PR #4193, opened 2019-01-29
Fix geoserver location

giohappy:
The current two options GEOSERVER_LOCATION and GEOSERVER_PUBLIC_LOCATION are not enough to define the location of the Geoserver Web UI. GEOSERVER_LOCATION is reserved for internal use, GEOSERVER_PUBLIC_LOCATION points to the proxied OWS services. 
A new **GEOSERVER_WEB_UI_LOCATION** option is provided and by default it points to SITEURL/geoserver/.


giohappy:
closing this PR to remove trask commits. Sorry

PR #4201, opened 2019-01-30
proxy auth refactoring with extended checks

giohappy:
This PR includes:

- a bit of refactoring of the auth_token retrieval logix inside the proxy view
- extend token in case a valid session is available but the token is expired
- introduces an oauth.py module to gather oauth and access_token related utilities


giohappy:
Travis is failing [here](https://travis-ci.org/GeoNode/geonode/jobs/486448412#L3322) but I don't understand why. In my manual tests it works as expected, since request.session['access_token'] shoud store an AccessToken object, but Travis reports that a string is returned in that test.

Issue #4207, opened 2019-02-05
Proxy base URL suffix for Docker is 'geoserver' instead of 'gs'


frafra:
It seems to me that using `/gs/` would be a better default, in order to keep the Docker setups more similar to the standard ones. It also raises some compatibility reasons, because I can see some hardcoded references to `/gs/` in various part of this repository (`MapLoom-*.js` for example).

Related discussion: https://github.com/GeoNode/geonode/issues/4206#issuecomment-460669614

giohappy:
I agree. The proxy base url should be updated wherever it's refernced (also in geonode-project).
The recent change is meant to make all requests pass thoruhg the GeoNode internal proxy in order to support Basic Auth, Bearer token and session based authentication.

gannebamm:
I can not express how much I like this kind of refactoring. +100

frafra:
Ok, I understood that the `/gs` suffix is just for external URLs, while for internal connection `/geoserver` has to be used.

```
$ grep -r 'localhost:8080/geoserver/' geonode | wc -l
20
```

PR #4208, opened 2019-02-05
Use gs instead of geoserver for URLs in Docker

frafra:
WIP pull request for #4207.

### DIFF
--- a/scripts/docker/env/production/celery.env
+++ b/scripts/docker/env/production/celery.env
@@ -15,7 +15,7 @@ IS_CELERY=true
 C_FORCE_ROOT=1
 SITEURL=http://localhost/
 # replaced with defaults in settings
-GEOSERVER_PUBLIC_LOCATION=http://localhost/geoserver/
+GEOSERVER_PUBLIC_LOCATION=http://localhost/gs/
 GEOSERVER_LOCATION=http://geoserver:8080/geoserver/
 OGC_REQUEST_TIMEOUT=300
 STATIC_ROOT=/mnt/volumes/statics/static/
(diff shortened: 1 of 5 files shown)


scripts/docker/env/development/django.env
```diff
@@ -1,7 +1,7 @@
 GEONODE_INSTANCE_NAME=geonode
 BROKER_URL=amqp://guest:guest@rabbitmq:5672/
 DJANGO_SETTINGS_MODULE=geonode.settings
-GEOSERVER_INTERNAL_URL=http://geoservices:8080/geoserver/
+GEOSERVER_INTERNAL_URL=http://geoservices:8080/gs/
```
afabiani: Internal URLs (using port 8080) must be /geoserver/

giohappy:
I think there's some confusion about the semantics of the locations.
**GEOSERVER_LOCATION**  as its internal URL to communicate with Geoserver. In Docker this should translate to the overlay network URL of the target service (geoserver).
**GEOSERVER_PUBLIC_LOCATION**  is instead the URL clients use to connect to Geoserver services (WMS, WFS, GWC, etc.). Since some commits ago this location points to the internal /gs proxy.
I need to review all the convoluted paths of these envs inside our compose files but this commit doesn't seem to fix the semantics mismatch.

frafra:
Sorry, this was a WIP pull request; it seems that there still something to be changed, because in recent revisions I get a gateway timeout.

frafra:
My pull request is broken. 
@giohappy could you please have a look at it and see if you can detect the error?

PR #4209, opened 2019-02-06
 Includes additional layer fields in the search index

ahmednoureldeen:


ahmednoureldeen:
@afabiani there are 2 build jobs stopped, I think it will pass if restarted.

PR #4210, opened 2019-02-06
Async thumbnail

hishamkaram:
This PR contain:
- add async_thumbnail task to generate thumbnails to improve layer upload time


giohappy:
IMHO the celery_enabled check should be only done during application startup and the result stored, globally.  Otherwise a test connection is opened on each thumbnail creation.
Morever the "celery or thread" pattern could be generalized to be used elsewhere.

giohappy:
I want to hear comments from others on this, but I'm concerned of the thread spawning. Without proper management (pools, etc.) we risk to have tons of threads running and pending in case of batch imports / updates.

And I think this pattern should in case be moved to a generic place (utils module or else) 

What's your opinion @afabiani?

afabiani:
> I want to hear comments from others on this, but I'm concerned of the thread spawning. Without proper management (pools, etc.) we risk to have tons of threads running and pending in case of batch imports / updates.
> 
> And I think this pattern should in case be moved to a generic place (utils module or else)
> 
> What's your opinion @afabiani?

Yes, I agree.
For the time being if celery is not enabled I would keep the usual sync mechanism until we can spend some more time and resources to envisage a better and more stable threading mechanism.

capooti:
Personally I'd avoid to generate thumbnails when changing styles for now, as it is time consuming and it can happen pretty often in the map composer. When needed the user can manually sync the thumbnail from the layer details view.

hishamkaram:
@afabiani @capooti please take a look at the last commit

geonode/geoserver/tasks.py
```diff
@@ -18,6 +18,8 @@
 #
 #########################################################################
 
+import threading
```
afabiani: This most probably will cause pep8 formatting issues

afabiani:
@hishamkaram

still pep8 issues
```
flake8 geonode
geonode/geoserver/signals.py:47:1: F811 redefinition of unused 'thumbnail_task' from line 24

```

The overall idea is good though. Thanks for your contribution.

giohappy:
@hishamkaram please run flake8 before submitting new commits.

afabiani:
Thanks @hishamkaram

hishamkaram:
@afabiani you are welcom

frafra:
This pull request contains a commit which introduces a regression: a18bbd4.

hishamkaram:
@frafra please review the PR comments

gannebamm:
> @frafra please review the PR comments

@hishamkaram what do you mean by that? Which comment should be reviewed and why? I am puzzled.

PR #4215, opened 2019-02-09
Fix missing replacement of geoserver public location

### LINKED ISSUE #4207: Proxy base URL suffix for Docker is 'geoserver' instead of 'gs'

francbartoli:
Closes #4207

### DIFF
--- a/docs/tutorials/install_and_admin/geonode_update/26_to_27/index.txt
+++ b/docs/tutorials/install_and_admin/geonode_update/26_to_27/index.txt
@@ -135,7 +135,7 @@ Upgrade Development Environment
             64
             65  GEOSERVER_PUBLIC_LOCATION = os.getenv(                                                                                  49  GEOSERVER_PUBLIC_LOCATION = os.getenv(
             ------------------------------------------------------------------------
-            66  #    'GEOSERVER_PUBLIC_LOCATION', '{}gs/'.format(SITEURL)                                                    <>
+            66  #    'GEOSERVER_PUBLIC_LOCATION', '{}geoserver/'.format(SITEURL)                                                    <>
             67      'GEOSERVER_LOCATION', 'http://localhost:8080/geoserver/'                                                            50      'GEOSERVER_PUBLIC_LOCATION', 'http://localhost:8080/gs/'
             68  )
             ------------------------------------------------------------------------
--- a/docs/tutorials/install_and_admin/geonode_update/26_to_27/index.txt
+++ b/docs/tutorials/install_and_admin/geonode_update/26_to_27/index.txt
@@ -135,8 +135,8 @@ Upgrade Development Environment
             64
             65  GEOSERVER_PUBLIC_LOCATION = os.getenv(                                                                                  49  GEOSERVER_PUBLIC_LOCATION = os.getenv(
             ------------------------------------------------------------------------
-            66  #    'GEOSERVER_PUBLIC_LOCATION', '{}geoserver/'.format(SITEURL)                                                    <>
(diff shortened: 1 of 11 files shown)


docs/tutorials/install_and_admin/geonode_update/26_to_27/index.txt
```diff
@@ -135,8 +135,8 @@ Upgrade Development Environment
             64
             65  GEOSERVER_PUBLIC_LOCATION = os.getenv(                                                                                  49  GEOSERVER_PUBLIC_LOCATION = os.getenv(
             ------------------------------------------------------------------------
-            66  #    'GEOSERVER_PUBLIC_LOCATION', '{}geoserver/'.format(SITEURL)                                                    <>
-            67      'GEOSERVER_LOCATION', 'http://localhost:8080/geoserver/'                                                            50      'GEOSERVER_PUBLIC_LOCATION', 'http://localhost:8080/geoserver/'
+            66  #    'GEOSERVER_PUBLIC_LOCATION', '{}gs/'.format(SITEURL)                                                    <>
+            67      'GEOSERVER_LOCATION', 'http://localhost:8080/geoserver/'                                                            50      'GEOSERVER_PUBLIC_LOCATION', 'http://localhost:8080/gs/'
```
afabiani: This URL is not correct, it should be either:

http://localhost:8080/geoserver/

or 

http://localhost:8000/gs/

not a mix of the two above.

francbartoli: @afabiani right, sorry I did not realize of the sphinx diff visualization. If I understand correctly you mean:

```
65  GEOSERVER_PUBLIC_LOCATION = os.getenv(                                                                                  49  GEOSERVER_PUBLIC_LOCATION = os.getenv(
            ------------------------------------------------------------------------
            66  #    'GEOSERVER_PUBLIC_LOCATION', '{}geoserver/'.format(SITEURL)                                                    <>
            67      'GEOSERVER_LOCATION', 'http://localhost:8080/geoserver/'                                                            50      'GEOSERVER_PUBLIC_LOCATION', 'http://localhost:8080/gs/'
            68  )

```

right?

giohappy:
this diff example refers to an older version of GN. The use of /gs for GEOSERVER_PUBLIC_LOCATION was introduced only recently.
So IMHO the fix is not relevant for that file...

docs/tutorials/install_and_admin/geonode_update/26_to_27/index.txt
afabiani: Right. You should also make sure this not happens on other lines below.

francbartoli:
> this diff example refers to an older version of GN. The use of /gs for GEOSERVER_PUBLIC_LOCATION was introduced only recently.
> So IMHO the fix is not relevant for that file...

Yep but the /gs location was not present at that time so it is correct to preserve the old value and make the user aware to edit it to the new /gs.

However, despite the diff is actually useful, I feel the sphinx rendering a little bit confusing IMHO

giohappy:
But the file refers to 26_to_27 migration. The new endpoind should be used for {old}_2.10.
Anyway I agre with you that the sphinx output is not that great and confusing....

PR #4216, opened 2019-02-10
Make sure geogig can be created in a schema other than public

ahmednoureldeen:


afabiani:
@ahmednoureldeen I'm not a geogig expert unfortunately, and also I don't have an environment ready to test this change.

I can just "trust" you and merge this PR or you mught request a review from someone using it.
Just let me know.

giohappy:
I don't see what's the benefit of using the custom DB options, instead of the current "SCHEMA" field. Does the geogic module ever interact with the geogig datastore direclty from Django?

In any case the extraction of the scheme option should be more robust. You simply split the "options" field by "=", but it doesn't take into account cases where more options are included in the options string.
A regex should be used instead, matching the following string portion: "-c search_path=myschema,public".

ahmednoureldeen:
"SCHEMA" field is not working with postgresql - may be it was working in older versions of postgres & psycopg-  , so the proper way to define schema name in search_path options.
 by `schema = map(str.strip, search_path.split(','))[0]` we select first schema defined.
but we need to handle if options rather than search_path.

but I think it is sufficient for now, we can enhance it later.

hishamkaram:
@afabiani  working fine for me 
screen shot 2019-02-13 at 1 32 58 pm

giohappy:
> but I think it is sufficient for now, we can enhance it later.

@hishamkaram I don't agree. If someone include an additional custom option things will break. So this should be strnghten from the beginning.

hishamkaram:
@giohappy @afabiani please take a look on the last commit?

giohappy:
looks good to me now

PR #4221, opened 2019-02-12
auto-create Tastypie ApiKey objects

hishamkaram:
This PR include:
- add signal to auto-create Tastypie ApiKey objects after creating new user ([docs](https://django-tastypie.readthedocs.io/en/latest/authentication.html))

> 
> Tastypie includes a signal function you can use to auto-create ApiKey objects. Hooking it up looks like:
> 
> ```
> from django.contrib.auth.models import User
> from django.db.models import signals
> from tastypie.models import create_api_key
> 
> signals.post_save.connect(create_api_key, sender=User)
> ```


giohappy:
@hishamkaram IMO the API authentication should use the tokens already managed for Oauth AuthN/AuthZ. These tokens are provided by the django-oauth-toolkit, to provide authentication to Geoserver (and any third party Oauth2 / OpenID client), and we do custom management during [login/logout](https://github.com/GeoNode/geonode/blob/master/geonode/people/models.py#L254) and inside [proxied requests](https://github.com/GeoNode/geonode/blob/master/geonode/proxy/views.py#L64). Recently I started to organize these utils inside a dedicated [auth](https://github.com/GeoNode/geonode/blob/master/geonode/base/auth.py) module.

To align the Tastypie APIs with this (that I think should be moved to DRF in the future!) we should implement a custom Authentication class for Tastypie to extract the oauth tokens as we do inside the proxy view.

So, from my point of view, this PR should be closed.

hishamkaram:
@giohappy in this case I don't know if we need to keep this [auth class](https://github.com/GeoNode/geonode/blob/master/geonode/api/authorization.py#L80),[authentication](https://github.com/GeoNode/geonode/blob/master/geonode/api/resourcebase_api.py#L678) , for Tastypie or not.

hishamkaram:
@giohappy if you decided to keep ApiKeyAuthentication  I think this method is useless because you need to run this command `python manage.py backfill_api_keys` to create tokens for the users to be able to deal with the api

hishamkaram:
What's your opinion @afabiani?

afabiani:
Well, the thing is... if we decide to keep ApiKeyAuth for Tastipye than the signal could be useful in order to create keys for users at creation.

However, what @giohappy was suggesting (I guess) is to make the OAuth2 Framework generating those keys instead of using the tastipye custom ones. I fully agree with this since the same keys could be used also as valid OAuth2 access tokens.

So, long story short, your suggestion is good, but we should connect a custom method to the signal in order to get or create a valid access_token through the OAuth2 utility methods and then use the latter as API key too.

giohappy:
we already have those signals for login/logout. The tokens are already there, so there's no need to create the tokens again.
We should only change  [GeonodeApiKeyAuthentication](https://github.com/GeoNode/geonode/blob/master/geonode/api/authorization.py#L80) extract the token from headers as the proxy view already does [here](https://github.com/GeoNode/geonode/blob/master/geonode/proxy/views.py#L159)

PR #4222, opened 2019-02-13
fix filter by type for haystack

ahmednoureldeen:


geonode/layers/search_indexes.py
```diff
@@ -99,7 +99,7 @@ def prepare_type(self, obj):
     def prepare_subtype(self, obj):
         if obj.storeType == "dataStore":
             if obj.has_time:
-                return "vectorTimeSeries"
+                return "vector_time"
```
afabiani: Why we need to change this? Such change will break the context menu counter.

ahmednoureldeen: actually , this fixes the context menu counter when haystack is enabled, 
while haystack indexes 'vectorTimeSeries' context menu uses 'vector_time' to filter.

PR #4224, opened 2019-02-13
Add rabbitmq volume

frafra:
[RabbitMQ Dockerfile](https://github.com/docker-library/rabbitmq/blob/master/Dockerfile-alpine.template) contains a `VOLUME` command, thus creating a volume even if `docker-compose.yml` do not use it apparently. It is better to be aware of the existence of that volume to avoid further problems and permission issues with third party software (volumes created by containers have different permissions from the ones defined by the compose file in Portainer).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,8 @@ services:
     labels:
         org.geonode.component: rabbitmq
         org.geonode.instance.name: geonode
+    volumes:
+      - rabbitmq:/var/lib/rabbitmq
 
   celery:
     restart: unless-stopped
@@ -138,4 +140,5 @@ volumes:
     name: ${COMPOSE_PROJECT_NAME}-dbdata
   dbbackups:
     name: ${COMPOSE_PROJECT_NAME}-dbbackups
-
+  rabbitmq:
+    name: ${COMPOSE_PROJECT_NAME}-rabbitmq
--- a/scripts/spcgeonode/docker-compose.yml
+++ b/scripts/spcgeonode/docker-compose.yml
@@ -156,7 +156,9 @@ services:
 
   # Vanilla RabbitMQ service. This is needed by celery
   rabbitmq:
-    image: rabbitmq:3.7-alpine 
+    image: rabbitmq:3.7-alpine
+    volumes:
+      - rabbitmq:/var/lib/rabbitmq
     restart: on-failure
 
 volumes:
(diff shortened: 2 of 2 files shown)


giohappy:
@frafra can you elaborate on this? what conflictual situations can happen when a container define its own VOLUME? The rest of Docker environment (orchestrators, containers, etc.) shouldn't now anything about its own internal stuff if its not exposed, IMHO.
A simple argument in favour of this is: what if RabbitMQ changess the RABBITMQ_DATA_DIR env var?

frafra:
> @frafra can you elaborate on this? what conflictual situations can happen when a container define its own VOLUME? The rest of Docker environment (orchestrators, containers, etc.) shouldn't now anything about its own internal stuff if its not exposed, IMHO.

Better to be explicit than implicit in my opinion; developers should be aware of which volumes are used without having to look at the Dockerfile, at their parents and so on.

Brief summary of what happened today to me: we had a VM with GeoNode + Portainer (because sysadmin do not want to give to developers root rights to a production systems and we do not have a sophisticated orchestrator at the moment), where all the volumes have been replaced by bind mounts (for external backups etc.); both me and the sysadmin didn't evaluate the possibility that there could be a volume hidden inside a Dockerfile, so after few weeks we end up with a VM with a disk that is full because RabbitMQ was storing data inside a VM that was not meant for storage. In addition to this, I never seen the RabbitMQ volume that was causing issue on Portainer, because volumes created by `VOLUME` inside a docker file and volumes created by a compose file had different permission by default (the first was visible to the admin only). Even when not using Portainer, the RabbitMQ volume is not named with the same prefix like the others, causing confusion on systems with many volumes.

If making that volume explicit is considered a no-go, I propose to add a comment instead.

> A simple argument in favour of this is: what if RabbitMQ changess the RABBITMQ_DATA_DIR env var?

It would not be an issue as long as developers test geonode when they upgrade an external dependency (like moving to a new version of RabbitMQ), but that has to be done anyway. Same issue would arise if Postgres decide to store its data under a different directory.

giohappy:
ok, so it doesn't cause any issues by itself. You're proposing this just to facilitate the life to admins :)
That's fine to me.

francbartoli:
@frafra can you tweak the root docker-compose file too? I would keep this kind of changes aligned

frafra:
The CI test failed for a random issue that is not related to the commit. Please run it again.

PR #4225, opened 2019-02-14
tastypie OAuth Backend

hishamkaram:
This PR include:
- implement OAuth Backend for tastypie


geonode/api/authentication.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+#########################################################################
+#
+# Copyright (C) 2016 OSGeo
```
afabiani: Could you set this to 2019

hishamkaram: fixed

giohappy:
it looks good but as @afabiani asked you should include tests

hishamkaram:
@giohappy @afabiani  ok the next two days are holiday in Egypt so it's a good time to write the test cases

hishamkaram:
@afabiani

PR #4226, opened 2019-02-14
Use geoserver instead of gs (partial revert)

frafra:
Fix and complete https://github.com/GeoNode/geonode/pull/4208.
I am not sure of the logic behind it, but this is the smallest patch that makes possible again to upload layers in GeoNode.

### DIFF
--- a/scripts/spcgeonode/django/docker-entrypoint.sh
+++ b/scripts/spcgeonode/django/docker-entrypoint.sh
@@ -23,7 +23,7 @@ else
     fi
 fi
 
-export GEOSERVER_PUBLIC_LOCATION="${SITEURL}/gs/"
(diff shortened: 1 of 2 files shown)


afabiani:
@frafra I trust you here, I have no time/resources to test spcgeonode docker setup now

frafra:
I found that the previous pull request introduced a regression using https://github.com/GeoNode/geonode-selenium and those two changes are sufficient to make it work again.

The one applied to `docker-compose.yml` seems reasonable to me, while the other is a bit strange, and it seems to me and there is probably a naming consistency issue.

afabiani:
@frafra I can just merge this PR, but maybe it is worth to write down an issue for the memories. I have the feeling that this is something that should be inspected with more attention.

PR #4228, opened 2019-02-14
fix Object of type Map is not JSON serializable

hishamkaram:


hishamkaram:
@afabiani fix for Object of type Map is not JSON serializable (celery4)

afabiani:
oh! Thanks @hishamkaram 

this is why tests are important and very welcome!

hishamkaram:
@afabiani you are right

simboss:
...and where are tests in this case?

PR #4237, opened 2019-02-18
Setup Postgres password for SPC Geonode

frafra:
This patch allows to setup a PostgreSQL password for SPC Geonode.

Password is not set by default (the variable is empty, thus avoiding breaking backwards compatibility), but if sets makes safer to expose Postgres port over the internal network (useful if a centralized backup system is in use).

@olivierdalang Do you know how to fix the pgdumper cron script to use an environmental variable for the password?

### DIFF
--- a/scripts/spcgeonode/.env
+++ b/scripts/spcgeonode/.env
@@ -26,7 +26,7 @@ ADMIN_USERNAME=super
 ADMIN_PASSWORD=duper
 ADMIN_EMAIL=admin@example.com
 
-# PostgreSQL superuser password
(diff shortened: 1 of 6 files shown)


olivierdalang:
@frafra The only way I see would be to install the cronjob in the entrypoint, and using `envsubst` to replace the variables (you can look in the rclone entrypoint how it's done).

I'd suggest adding as a comment in the `.env` that the password is optional and only needed if postgres is to be exposed.

frafra:
Done, thank you @olivierdalang :-)

PR #4242, opened 2019-02-19
Update GeoServer to 2.14.2

frafra:
Pull request for SPC Geonode. No regression detected with `geonode-selenium`. Using `ARG` to parametrize the Dockerfile.


frafra:
Added patch to further improve the geonode Dockerfile to support additional plugins.

Issue #4232, opened 2019-02-15
Unable to upload layers anymore


frafra:
Running SPCGeonode, GeoNode commit 287708d101732cb665096d257126594089f384c9 (Jan 15 2019). I am unable to upload layers. Data comes from a 2.8 instance. I tried to generate and run migrations to see it that could be the issue, but it is not (and the makemigrations warning is shown every time). GeoNode and GeoServer can talk each other and the configuration seems just fine.

It worked fine, now I have the same issue on two different servers. The only thing that happens is that the containers get restarted from time to time.

The first relevant error is the following one (I am logged in as administrator):

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/oauth2_provider/oauth2_validators.py", line 380, in validate_bearer_token
    access_token = AccessToken.objects.select_related("application", "user").get(token=token)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 380, in get
    self.model._meta.object_name
DoesNotExist: AccessToken matching query does not exist.
```

Then there is a database error and an additional error trying to manage the previous one.

Full logs below:

```
-----------------------------------------------------
STARTING DJANGO ENTRYPOINT Fri Feb 15 09:47:29 UTC 2019
-----------------------------------------------------
Running initialize.py...
-----------------------------------------------------
1. Running the migrations
Operations to perform:
  Apply all migrations: account, actstream, admin, agon_ratings, announcements, auth, avatar, base, contenttypes, dialogos, documents, geonode_client, geonode_themes, groups, guardian, invitations, layers, maps, oauth2_provider, people, pinax_notifications, services, sessions, sites, socialaccount, taggit, tastypie, upload, user_messages
Running migrations:
  No migrations to apply.
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
-----------------------------------------------------
2. Creating/updating superuser
superuser successfully updated
-----------------------------------------------------
3. Create/update an OAuth2 provider to use authorisations keys
oauth2 provider successfully updated
-----------------------------------------------------
4. Loading fixtures
Installed 300 object(s) from 1 fixture(s)
-----------------------------------------------------
5. Running updatemaplayerip
-----------------------------------------------------
 Collecting static files
-----------------------------------------------------
7. Securing GeoServer
Master password was already changed. No changes made.
-----------------------------------------------------
[...]

PR #4245, opened 2019-02-20
Import SSL Certificate for GeoServer

frafra:
> @frafra no matter if the certificate is valid or not, you need to add it to the java truststore in order to make GeoServer recognize it and avoid throwing that exception.
> 
> See if this [link ](https://docs.geoserver.org/stable/en/user/community/oauth2/index.html#ssl-trusted-certificates) might be helpful somehow.

https://github.com/GeoNode/geonode/issues/4232#issuecomment-465524029


abersh:
Is there anyone solve this error to upload layers  
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)

Issue #4129, opened 2018-12-17
Migrate Geonode Frontend from AngularJS to React


Coop56:
This issue will serve as a GNIP, and it's something we're working on. We’re looking to do the work to migrate the geonode front end away from Angular to ES6 and React. We’re planning to do the work on a fork and PR it into master geonode when it’s ready. 

The goal will be to replace the current frontend Javascript and AngularJS code with modern ES6 and React components.  

General Steps to Complete this GNIP

1. Configure project to use Webpack & Yarn.
2. Update Docker configuration to use Webpack & Yarn.
3. Replace Require.js with ES6 module syntax.
4. Update Geonode JS to modern ES6 syntax.
5. Refactor existing modules into React Components
6. Update Geonode documentation.

Our goal is to remove AngularJS as a dependency entirely, and get all of the Javascript up to date with modern ES6 syntax.  We'll also be able to streamline the JS asset management in the process.  If there are any questions I'll be happy to answer with what we're planning, and once we have a branch up and going I'll edit this post and include a link to the branch so the community can see the work in progress.

francbartoli:
@afabiani @simod @capooti @olivierdalang can you kindly express your vote?

afabiani:
Looks a good plan... a bit afraid it could break some things, especially on contribs.
Nevertheless:

1. I would be very happy to see GeoNode cleaned from all those mixed stuff and technologies and have a clean and unique way to write frontend logic.

2. I would be very happy to cleanup obsolete contrib apps and, if possible, remove them from trunk and attach as external libraries instead, with the positive effect of cleaning up the settings too.

Hopefully this GNIP would be a good occasion to start the positive trend.

giohappy:
@Coop56 we ([GeoSolutions](https://github.com/geosolutions-it)) are working on the integration of several new components for the maps previews and views, based on our React based framework [MapStore](https://mapstore2.geo-solutions.it/mapstore/). It integrates charts widgets, a new timeline, a style editor, and several other features.
We're going to merge this client soon in January on master, but it's already working and deployed on some of our projects. 

So we support the idea of migrating all the GeoNode's frontend from Angular. It is advisable to coordinate the offert, at least to agree on a toolchain (Webpack and Node versions, etc.) and, hopefully, on some basic componentes and patterns. (e.g. we adopt Redux, RxJS). 
This would help a lot to obtain a seamless integration and provide a coherent structure to the frontend architecture.

Coop56:
[...]

PR #4246, opened 2019-02-20
[DRAFT] Remove AngularJS from Geonode Search.

trepagnier:
This PR is connected to https://github.com/GeoNode/geonode/issues/4129. It is an early draft that will complete the items listed below. The list will grow in number and detail as work on this PR progresses:

- [x] Integrate Webpack into the Geonode static directory.
- [x] Configure Webpack to transpile ES6 to ES5 for browser support.
- [ ] Remove all references to AngularJS inside the Geonode search codebase and replace with  a mixture of React, jQuery, and vanilla ES6.
- [ ] Introduce a central search controller that consists of minimal declarative code. All imperative code will be abstracted away to helper functions inside of a `functions` directory.
- [ ] Split search code into smaller, more maintainable modules.
- [ ] Componentize key search features.
- [x] Fix the extent bug in the Leaflet sidebar map that causes bounds to grow infinitely as one scrolls.
- [x] Introduce a `Search` factory function. An instance of `Search` will be used to store all data related to searching the Geonode catalogue. This instance should be regarded as the single source of truth for all search related data. Any stateful components or modules should receive their state from the Search instance. 
- [x] Implement a pubsub paradigm for the search codebase. Communication between components should be handled via publications and subscriptions in order to ensure loose coupling between components. 
- [x] Provide an Angular shim that syncs data between Angular's `$scope` and a given `Search` instance, which will be used until Angular is entirely removed from the project.


francbartoli:
@darinacosta @Coop56 I believe a connected issue with this is required as per our latest rules from PSC. Also minimal e2e tests for the UI features touched by this would be actually useful

Coop56:
@francbartoli The issue connected with this PR is https://github.com/GeoNode/geonode/issues/4129 - do we just need to add that to the description of the PR?

francbartoli:
Yes please, reference it somewhere

francbartoli:
Another comment is about jquery. Do we need to still maintain it? Can we replace with something like [Axios](https://github.com/axios/axios) for ajax calls?

darinacosta:
@francbartoli Regarding the use of jQuery: there are a couple of components rely on it, such as [TextSearchForm](https://github.com/MapStory/geonode/blob/react-migration/geonode/static/geonode/js/search/components/TextSearchForm.js) and [SelectionTree](https://github.com/MapStory/geonode/blob/react-migration/geonode/static/geonode/js/search/components/SelectionTree.js). Perhaps we can address  its removal in a future phase of refactoring?

francbartoli:
> @francbartoli Regarding the use of jQuery: there are a couple of components rely on it, such as [TextSearchForm](https://github.com/MapStory/geonode/blob/react-migration/geonode/static/geonode/js/search/components/TextSearchForm.js) and [SelectionTree](https://github.com/MapStory/geonode/blob/react-migration/geonode/static/geonode/js/search/components/SelectionTree.js). Perhaps we can address its removal in a future phase of refactoring?

Main concern was about still using if for ajax call even though getting rid of jquery would be better. Thanks for addressing bdd tests

afabiani:
@trepagnier closing this since it is stale for a long time and has conflicts.
Please re-send if you want by following the correct path:
 * Create an issue describing what you want to achieve
 * Make sure the PR respects all the requirements

Thanks,
@afabiani

Issue #4255, opened 2019-02-25
Python Requirements.txt Dependency Audit


Coop56:
The requirements.txt file has grown rather long over the years, and many of the dependencies listed there are no longer relevant.  I'll be preparing a PR to remove all of the ones that are no longer used.  I'll ensure that the current tests are in the same state as they are on master, and manually check that each dependency that I remove is no longer used anywhere in the Python codebase.

PR #4256, opened 2019-02-26
Removes Old Python Dependencies

Coop56:
Removes unnecessary dependencies from the requirements.txt and settings.py files.  This PR is connected to issue https://github.com/GeoNode/geonode/issues/4255.  I verified that the tests pass locally.  They're also working on Travis CI now as well.

All of these dependencies were checked against the Python codebase to make sure they weren't being used, and the tests seem to agree.  I also did some manual smoke testing to make sure nothing was broken.  If the Travis CI tests show anything different, or if anyone sees a dependency they think we still need I'll take a look and add them back if necessary.

### DIFF
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -354,29 +354,19 @@ INSTALLED_APPS = (
     'leaflet',
     'bootstrap3_datetime',
     'django_filters',
-    'django_extensions',
     'django_basic_auth',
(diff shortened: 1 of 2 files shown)


afabiani:
seems good to me

afabiani:
Thanks @Coop56

PR #4258, opened 2019-02-27
[DRAFT][Closes #4251] Introduce a SessionExpiredMiddleware in order to checkfor access_token validity

### LINKED ISSUE #4251: Introduce a "SessionExpiredMiddleware" in order to check for access_token validity
Related to #3563

The idea is to introduce a Middleware class looking for Token expiration and session validity.

Accordingly to the settings it will either force the user to login again or refresh/extend the token validity.

afabiani:


afabiani:
@frafra if I start GeoNode locally and run geonode-selenium/test.sh (no spcgeonode, just a standard one) tests are good, but somehow here are failing... any idea why?

frafra:
@afabiani That is because of different user/pass/url compared to spcgeonode.
Update your `geonode-selenium` and run it like this:
```
GEONODE_USER=admin GEONODE_PASS=admin GEONODE_URL=http://localhost ./test.sh
```

We could run the test against the standard docker compose files, in order to improve the test coverage until we merge the two version.

afabiani:
@frafra that would be very useful, until we decide which one use as default.

PR #4259, opened 2019-02-28
Test standard docker setup with geonode-selenium

frafra:
As we have 2 docker setups at the moment (SPC and standard), let's test both of them.


afabiani:
@frafra from the logs I guess this curl option `--retry-connrefused` is not recognized

frafra:
@afabiani Yes, you are right, that's the last thing to be fixed :)

afabiani:
@frafra almost there :)

I guess this file https://github.com/GeoNode/geonode/blob/master/pytest.ini interfere with selenium tests when running geonode-selenium script

frafra:
@afabiani I updated geonode-selenium: please build it again :-)

afabiani:
@frafra now started but still errors... is it possible that somehow the url of GeoServer is wrong? From the logs it looks to me it is not able to render the page.

frafra:
@afabiani If I upload the data manually it works, but I get this error in the test: `Unexpected exception Could not access to uploaded data.`. Do you have any clue?

afabiani:
@frafra uhm not sure, but it comes from the following statements

```
        tempdir = tempfile.mkdtemp(dir=settings.FILE_UPLOAD_TEMP_DIR)
        logger.info("valid_extensions: {}".format(form.cleaned_data["valid_extensions"]))
        relevant_files = _select_relevant_files(
            form.cleaned_data["valid_extensions"],
            req.FILES.itervalues()
        )
        logger.info("relevant_files: {}".format(relevant_files))
        _write_uploaded_files_to_disk(tempdir, relevant_files)
        base_file = os.path.join(tempdir, form.cleaned_data["base_file"].name)
```

Is it possible that somehow the `settings.FILE_UPLOAD_TEMP_DIR` variable is pointing to an unaccessible store?

frafra:
@afabiani It looks like it is not enough to wait for GeoNode to go up with the standard Docker setup. I added a `sleep` instruction just as a workaround.

frafra:
In SPCGeonode the Django container is the one to look at to see if the system is up and running, but that's not the case with the default compose file. It would be better to implement some health checks in the compose file, to determine when the system is actually ready to be tested.

afabiani:
@frafra do you need me to take a look?

frafra:
Thanks @afabiani; I can do it by myself, but I actually trying to investigate and fixing some other issues I have at the moment, so it cold require a bit of time before I fix this pull request.
For the moment I just wrote down what I discovered yesterday (there has been a discussion about it in chat) :-)

afabiani:
@frafra no worries, thanks for your effort. Let me know how I can help!

afabiani:
@frafra I did some more tests with selenium tests, but they look to me quite unstable.
Don't know exactly why, but from my local tests most of the times I experience some missing data/user accounts, even though the fixtures seem to be created successfully.
If I purge all the volumes and try again at least the login test passes, the upload ones not.
Could it be related to some misconfigured volumes?

frafra:
> @frafra I did some more tests with selenium tests, but they look to me quite unstable.
> Don't know exactly why, but from my local tests most of the times I experience some missing data/user accounts, even though the fixtures seem to be created successfully.
> If I purge all the volumes and try again at least the login test passes, the upload ones not.
> Could it be related to some misconfigured volumes?

I do not think so. selenium tests seems very reliable on spcgeonode, but on fresh standard Docker setup I wasn't able to produce something reliable; it looks like there is something that is still going on, but I didn't had the time to work more on it.

frafra:
Things are going better. It still looks to me that the standard Docker setup is a bit unreliable; maybe because of different DEFAULT_BACKEND_UPLOADER settings?

PR #4270, opened 2019-03-07
Run dos2unix on .env files

### LINKED ISSUE #4267: DOS vs UNIX newlines

frafra:
Fix #4267.

There are other files that would need the same treatment, but `.env` and Bash files are the most important.


afabiani:
It would be good to run dos2unix on all files sooner or later.

We will need dedicated PRs tough, otherwise will be impossible to understand what's changed or not.

PR #4281, opened 2019-03-12
add missing settings in settings.py

ahmednoureldeen:
update settings.py with settings defined in local_settings.py.geoserver.sample 
to be able to run with default settings as local_settings.
updated settings related to site host and port and geoserver port and location.


ahmednoureldeen:
@afabiani I think we shouldn't check against hard-coded url as below

```
                if link['scheme'] == 'OGC:WMS':
                    self.assertEqual(
                        link['url'],
                        'http://localhost:8000/gs/ows',
                        'Expected a specific OGC:WMS URL')
```

for default GEOSERVER_PUBLIC_PORT and GEOSERVER_PUBLIC_HOST the url should be http://localhost:8080/geoserver/ows

geonode/local_settings.py.geoserver.sample
```diff
@@ -131,7 +131,7 @@ GEOSERVER_PUBLIC_PORT = os.getenv(
     'GEOSERVER_PUBLIC_PORT', 8080
 )
 
-_default_public_location = 'http://{}:{}/gs/'.format(GEOSERVER_PUBLIC_HOST, GEOSERVER_PUBLIC_PORT) if GEOSERVER_PUBLIC_PORT else 'http://{}/gs/'.format(GEOSERVER_PUBLIC_HOST)
+_default_public_location = 'http://{}:{}/geoserver/'.format(GEOSERVER_PUBLIC_HOST, GEOSERVER_PUBLIC_PORT) if GEOSERVER_PUBLIC_PORT else 'http://{}/gs/'.format(GEOSERVER_PUBLIC_HOST)
```
afabiani: This must be `http://{}:{}/gs/` instead of `http://{}:{}/geoserver/`

geonode/settings.py
```diff
@@ -736,12 +740,22 @@
...
 
+GEOSERVER_PUBLIC_HOST = os.getenv(
+    'GEOSERVER_PUBLIC_HOST', SITE_HOST_NAME
+)
+
+GEOSERVER_PUBLIC_PORT = os.getenv(
+    'GEOSERVER_PUBLIC_PORT', 8080
+)
+
+_default_public_location = 'http://{}:{}/geoserver/'.format(GEOSERVER_PUBLIC_HOST, GEOSERVER_PUBLIC_PORT) if GEOSERVER_PUBLIC_PORT else 'http://{}/gs/'.format(GEOSERVER_PUBLIC_HOST)
```
afabiani: This must be `http://{}:{}/gs/` instead of `http://{}:{}/geoserver/`

ahmednoureldeen: the result url is not found for default values ..  http://localhost:8080/gs
i think it should refere to geoserver url as the definition is for public port & host for geoserver meanwhile /gs is geonode url( act as a proxy for geoserver)

afabiani: Yes, because of the default GEOSERVER_PUBLIC_PORT. It should be 8000 instead of 8080.

The policies are not to pass through the GeoNode proxy.

francbartoli: I have the impression that here we introduced a regression:

if `_default_public_location` ends always up with the form `http://{}:{}/gs/` then the default result of GEOSERVER_WEB_UI_LOCATION is wrong because neither django urls nor jetty/tomcat have such a location `/gs`

@afabiani @giohappy what do you think?

afabiani: GEOSERVER_WEB_UI_LOCATION should be a variable with it's own value

francbartoli: Is it currently wrong indeed?

afabiani: @francbartoli not if we are going to use GeoNode proxy

PR #4288, opened 2019-03-14
[Closes #4106] [Remote Services] Add the possibility of filtering the list of resources

### LINKED ISSUE #4106: [Remote Services] Add the possibility of filtering the list of resources
When harvesting resources associated to a Remote Service, we just get a long list of layers.
Even though the list is paginated, with hundreds of layers it is difficult to find the resources we need (see the example below)

image

It would be nice to have at least a filter search box providing a simple plain text match on title and abstract in order to refine the list of resources shown on the page.

afabiani:


atherashraf:
can you please elaborate the issue. as I cant see details here...

PR #4290, opened 2019-03-16
 - Hardening GeoNode and cleanup: http_client uses Bearer Auth whenev…

afabiani:
…er it's possible


geonode/geoserver/utils.py
```diff
@@ -50,12 +55,15 @@ def geoserver_requests_session():
     session = requests.Session()
     access_token = None
     try:
-        from django.contrib.auth import get_user_model
-        from geonode.base.auth import get_or_create_token
-        _u = get_user_model().objects.get(username=_user)
-        access_token = get_or_create_token(_u)
```
giohappy: why don't you keep using the base.auth module? It was meant to unify and hide the access token retrieval logic

afabiani: get_user_model() causes issues with selenium tests somehow... but I will keep your suggestion and maybe update base.auth in order to allow filtering by username too.

PR #4313, opened 2019-03-26
[Closes #4312] Remove GeoGig and api_basemaps contrib apps integrated stuff

### LINKED ISSUE #4312: [Refers #4311] Remove GeoGig and api_basemaps contrib apps integrated stuff

afabiani:


francbartoli:
@afabiani what about to create at the same time the correspondent django package repository for each deleted app? I mean not a working app package but only the github repo with a cookiecutter django package bootstrap and the rough addition of the portion of code deleted from geonode. In this way people who liked a specific app can be addressed easily

afabiani:
@francbartoli yes this is the idea. In this case, I don't think it is possible since most of it was integrated into the core code. Since I don't think anyone is using geogit, I propose to just remove it. If someone would like to restore it in the future, we will evaluate a more clean and correct way to do it than just put "if-else" around the code.

francbartoli:
@afabiani actually not concerned about geogig but a little bit with api_basemaps which was slightly used and so I wouldn't totally throw away.
I totally agree with you that the sane way to add them again has to be by installing external django packages related to geonode and then enable them.

afabiani:
@francbartoli in which cases are "api_basemaps" used? The code seem to be quite unuseful, except maybe for the taglibs, which simply injects api keys into the context.

The rest looks like a GeoExt bridge, which IMHO is a workaround for something that is not used anymore.

francbartoli:
@afabiani I remember it addressed in the users list several times when users were asking for custom basemaps. It was my only concern but if you think it's not useful at all then you have my +1

afabiani:
@francbartoli I propose to get rid of it. It is really bounded to old GeoExt stuff. If someone needs custom basemaps, those must be hooked into the client library and as GeoExt extensions.

Issue #4125, opened 2018-12-14
GeoFence cache seems to be refreshed every time a layer is uploaded


capooti:
GeoFence cache seems to be refreshed every time a layer is uploaded.
This slows down the upload of a layer in our case, with thousands of layers. If I do import the layer with GS importer, logged as the GeoServer admin, it takes just a few seconds.
If I upload it with GeoNode it takes up to 3 minutes.
Is there anything which can be done to mitigate this?

afabiani:
The PR https://github.com/GeoNode/geonode/pull/4326 should fix the issue.

capooti:
@afabiani it seems now faster (a layer is uploaded in one minute). But still slow when compared to previous GeoNode version. One thing I have noticed: the geoserver configuration during a layer upload seems being reloaded: is it possible? That would explain the slowness at this point

afabiani:
@capooti nope as far as I can say. There's no command issuing a GeoServer configuration realod.
I kindly suggest to always use the 'geoserver.importer' upload method though, it is more resilient and efficient.

About the slowness, make sure you have enough UWSGI processes and connections available. Remember that now GeoServer asks GeoNode to validate the access tokens. If all the uwsgi processes are busy, it will have to wait the "harakiri" to free new ones.

afabiani:
@capooti you were right. I have found an unattended catalog.reload() while uploading a layer.
By removing that instruction the upload time has been significantly reduced.

capooti:
yes, @afabiani , it is much much faster now, thanks for fixing that!

Issue #3944, opened 2018-09-27
GNIP Make GeoFence efficient for GeoNode instances with a large number of layers


capooti:
After testing GeoFence with a GeoNode instance with a very large number of layers (about 25,000) and users (for a total of more than 700,000 GeoFence data rules) it came out that GeoFence performances are heavily affected.

There are 2 major issues in such a scenario: 

1) setting the permissions for a single layer from the GeoNode "set permissions" widget become very slow (up to 2/3 minutes in our tests)

2) every time an action that changes the GeoNode permission system happens (a layer upload, a layer creation, setting the permissions for a specific layer), this will invalidate the GeoFence cache and the GeoNode user has to wait for GeoFence to re-generate the whole cache - which typically slows down with the increase of the layers number (up to 4 minutes in our tests)

After some debugging and some discussion with @afabiani on the #geonode gitter channel, these are our findings in order to mitigate the two issues:

1) instead than setting the GeoFence permissions using the GeoFence REST API, use a direct connection to the database. We have a prototype of this here (will work only for GeoFence on PostGIS, not on GeoFence for H2 which is the default): https://github.com/GeoNode/geonode/blob/3f62d58a533a806c06ae86b5e6fb6c9327808426/geonode/security/utils.py#L315

2) Asynchronous GeoFence rules propagation. Unfortunately the GeoFence cache invalidation cannot be avoided and cannot be done on a layer by layer basis. Therefore a solution would be to "buffer" the cache invalidation calls and run them asynchronously instead of sending them synchronously every time there is a permission change on a layer.

afabiani:
Just few clarification and updates on this, following @capooti proposal and @olivierdalang observations.

* The idea is to make this optional. Of course, this is still a proposal, so technical details must still come. First implementation examples from @capooti show the possibility of controlling the GeoFence rules backend and switch between REST APIs and direct DB access. Don't how much work would be to make the DB backend configurable. A first implementation using PostGIS would be the best choice anyway.

* A lot of work has been done on Dockerization, both on GeoNode and GeoNode Project recently. Right now the docker setup supports Celery and Async mode by default. It can be controlled through env variables.

[...]

PR #4326, opened 2019-04-01
[GNIP #3944] Make GeoFence efficient for GeoNode instances with a large number of layers

afabiani:
This PR also would like to address the issue https://github.com/GeoNode/geonode/issues/4125
"GeoFence cache seems to be refreshed every time a layer is uploaded"


geonode/geoserver/management/commands/sync_geonode_layers.py
```diff
@@ -43,10 +43,10 @@ def sync_geonode_layers(ignore_errors, filter, username):
             print 'Syncing layer %s/%s: %s' % (count, layers_count, layer.name)
             # sync permissions in GeoFence
             perm_spec = json.loads(_perms_info_json(layer))
+            # re-sync GeoFence security rules
             layer.set_permissions(perm_spec)
             # recalculate the layer statistics
             set_attributes_from_geoserver(layer, overwrite=True)
-            layer.save()
```
capooti: This command is not only related to permissions, but to general synchronization between GeoServer/GeoFence/Django. I believe without this layer.save() thumbnails and styles won't be synced

afabiani: @capooti this was my intention. I mean, it seems to me dangerous and also a waste of time to regenerate all the thumbnails for the layers synced. It seems to me that the thumbnail, except the first time, should be manually updated by the user from the gui or at most we should envisage a specific management command for this.

capooti: well this command was thought as a sort of updatelayers alternative version which doesn't use GetCapabilities (which can be broken and it is broken and really slow in our case). The use case for us was to use it as a replacement to it.
We are also using it to migrate GeoNode itself, which explains why we want to update the thumbnails - in fact they are being generated for the first time in such a case. How about to add it as a command option that launch layer.save() if enabled?

afabiani: @capooti yes, good point. Make sense to add an explicit option which will warn users to eventually provide a filter whenever they want to explicitly update the thumbnails too. I will push the change in a few unless there are other objections.

giohappy: @afabiani I would add the following optional arguments: --updatepermissions, --updatethumbnails
If none are set all actions will be executed, otherwise only the ones set are executed.

capooti: great! thanks @afabiani and @giohappy

t-book:
I´ve just tested with a large number of layers (12.500). It´s  working much smoother now (https://github.com/GeoNode/geonode/issues/4125) Thanks @afabiani 

+1 for @giohappy suggestions regarding --updatepermissions, --updatethumbnails

capooti:
@t-book that is great to know. How long takes a layer upload for you?

t-book:
> @t-book that is great to know. How long takes a layer upload for you?

@capooti  ~10 – 20 seconds but this is not a valid measurement as 
- I´ve tested locally 
- Tested with very small test-shape-files (3-5 points each)

So basically I can say for sure:

**Old behavior:**
- changing permissions for one layer affected rules for all layers (**slow not usable for production**)

**New behavior:**
- changing permissions for one  layer only affects specific rules

I will repeat/do more testing on a remote staging instance with around 4000 layers. (Unfortunately at earliest next week)

geonode/geoserver/management/commands/sync_geonode_layers.py
```diff
@@ -41,12 +41,17 @@ def sync_geonode_layers(ignore_errors, filter, username):
         try:
             count += 1
             print 'Syncing layer %s/%s: %s' % (count, layers_count, layer.name)
-            # sync permissions in GeoFence
-            perm_spec = json.loads(_perms_info_json(layer))
-            # re-sync GeoFence security rules
-            layer.set_permissions(perm_spec)
-            # recalculate the layer statistics
-            set_attributes_from_geoserver(layer, overwrite=True)
+            if updatepermissions or not updatethumbnails:
```
giohappy: @afabiani I would make these arguments True by default and simply check is they're set, without any permutive check. I.e. if none are explicitly set by the user, ot both are set, both the actions should be performed.

afabiani:
@capooti @t-book with the latest commits I have updated the management command as suggested by @giohappy and added a few more improvements which should allow GeoNode to work with a very large number of datasets.

Currently, on a test instance with 22K layers the behavior is good.
The only things I noticed is that the api for the categories often seems to fail due to a memory exception. Not sure if it is a DB issue or, most probably, a memory leak with tastypie. This will need more investigation and it is bread for another issue.

I kindly suggest, in a production environment, to configure uWSGI wisely.

Typically try to have many processes with 1/2 threads each, since threads consume a lot of memory.

Also, be careful about the dynamic memory allocation. Spawning a lot of processes may easily consume all the available memory, therefore limit them. Moreover, in the case, GeoServer is running on the same machine, consider also the max heap memory needed by the Virtual Machine. If no more RAM is available, the JVM will crash.

geonode/geoserver/management/commands/sync_geonode_layers.py
afabiani: @giohappy done

giohappy: @afabiani what was something between this and the previous. Ok the defaults to true but if any of them is set the other must be False, unless they are explicitely true.

t-book:
Thanks a lot @afabiani ! Good point having an eye on uWSGI memory usage. In production, I normally set threads depending on available CPUs. 

Regarding memory I would add lowering logging to critical issues only ( `disable-logging = True` ) .
Also it might be worth checking process refreshing f.e. `max-worker-lifetime = 30`. The newer the process the faster it is (viewer memory management tasks). 

Thanks again!

PR #4336, opened 2019-04-03
[Fixes #4333] Upload from GUI doesn't handle projection properly

### LINKED ISSUE #4333: Upload from GUI doesn't handle projection properly
Comparison of detected projection from GeoServer is not correctly managed

afabiani:


francbartoli:
I have this native bbox

`('-180.0', '180.00000000000003', '-89.99892578124998', '83.59960937500006', 'EPSG:4326')`

and the check

```python
if -180 <= minx <= 180 and -180 <= maxx <= 180 and \
           - 90 <= miny <= 90 and -90 <= maxy <= 90:
```

is false then the step 6. is failing. I would propose to convert those variables to integer, I can send an update if we agree

afabiani:
> I have this native bbox
> 
> `('-180.0', '180.00000000000003', '-89.99892578124998', '83.59960937500006', 'EPSG:4326')`
> 
> and the check
> 
> ```python
> if -180 <= minx <= 180 and -180 <= maxx <= 180 and \
>            - 90 <= miny <= 90 and -90 <= maxy <= 90:
> ```
> 
> is false then the step 6. is failing. I would propose to convert those variables to integer, I can send an update if we agree

Uhm, quite strange, it should not fail just because of int conversion... I just noticed that there is a space between '-' and '90' --> - 90
Could this typo be the cause of your issue?

francbartoli:
Nope, this condition is false:

`-180 <= maxx <= 180`

I'm going to add the conversion, do you have some particular layer that we are happy to challenge?

PR #4342, opened 2019-04-04
GeoServer web UI url default fixed

### LINKED ISSUE #4315: GEOSERVER_WEB_UI_LOCATION default value is wrong
GeoNode web interface uses `GEOSERVER_LOCATION` for the menu (SPC). The hostname cannot be `nginx` because that domain is reachable from the container only.

@gannebamm this issue could be interesting for you.

frafra:
Fixes #4315.

Tested on both Docker setup successfully. I will add a new test on geonode-selenium too.


afabiani:
@frafra about the failing test, for the time being just remove the assert

`self.assertTrue(len(o['long-array-array']) > 0)`

Issue #4348, opened 2019-04-06
read the docs failing


t-book:
Read the docs are failing with:
```
Collecting Django==1.11.20 (from -r requirements_docs.txt (line 1))
  Using cached https://files.pythonhosted.org/packages/8e/1f/20bbc601c442d02cc8d9b25a399a18ef573077e3350acdf5da3743ff7da1/Django-1.11.20-py2.py3-none-any.whl
Collecting six==1.10.0 (from -r requirements_docs.txt (line 2))
  Using cached https://files.pythonhosted.org/packages/c8/0a/b6723e1bc4c516cb687841499455a8505b44607ab535be01091c0f24f079/six-1.10.0-py2.py3-none-any.whl
Collecting amqp==2.4.2 (from -r requirements_docs.txt (line 3))
  Using cached https://files.pythonhosted.org/packages/42/ec/cbbaa8f75be8cbd019afb9d63258e2bdc95242f8c46a54bb90db5fef03bd/amqp-2.4.2-py2.py3-none-any.whl
Collecting arcrest==10.3 (from -r requirements_docs.txt (line 4))
  Using cached https://files.pythonhosted.org/packages/17/2b/13e80c2ae5a4e6d1723f2c04c17c013c49db3d00fa701eecdb420e1ab44e/arcrest-10.3.tar.gz
Collecting python-slugify==3.0.1 (from -r requirements_docs.txt (line 5))
  Using cached https://files.pythonhosted.org/packages/a0/ff/3bd652257e8de19e09bd83544317d5b2e0f1d0b7e35633d75a3a215ce3ca/python-slugify-3.0.1.tar.gz
Collecting beautifulsoup4==4.7.1 (from -r requirements_docs.txt (line 6))
  Using cached https://files.pythonhosted.org/packages/8b/0e/048a2f88bc4be5e3697df9dc1f7b9d5c9c75be62676feeeb91d2e896c5ea/beautifulsoup4-4.7.1-py2-none-any.whl
Collecting boto==2.49.0 (from -r requirements_docs.txt (line 7))
  Using cached https://files.pythonhosted.org/packages/23/10/c0b78c27298029e4454a472a1919bde20cb182dab1662cec7f2ca1dcc523/boto-2.49.0-py2.py3-none-any.whl
Collecting celery==4.2.1 (from -r requirements_docs.txt (line 8))
[...]

PR #4349, opened 2019-04-06
#4348 fix failing read_the_docs

t-book:
As #4348 describes read the docs is failing because of Shapely. I think it could be fixed by installing libgeos-dev before building. Anyways I do not see a way how to force read_the_docs to do so.

Hence I would comment Shapely and pycsw (shapely is a requirement) in requirements_docs.txt. (_I do not see a reason why those would be needed for sphinx._)

Closes: #4348

### DIFF
--- a/requirements_docs.txt
+++ b/requirements_docs.txt
@@ -75,7 +75,7 @@ poster==0.8.1
 psutil==5.6.1
 psycopg2==2.7.7
 pycountry
-pycsw==2.2.0
+#pycsw==2.2.0
 pyopenssl==19.0.0
 pyproj==1.9.5.1
 pytest==4.3.1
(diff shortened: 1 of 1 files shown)


t-book:
I just found another way how we could solve this issue:

> This happens because our build system doesn’t have the dependencies for building your project. This happens with things like libevent, mysql, and other python packages that depend on C libraries. We can’t support installing random C binaries on our system, so there is another way to fix these imports.

import sys
from unittest.mock import MagicMock

```
class Mock(MagicMock):
    @classmethod
    def __getattr__(cls, name):
        return MagicMock()

MOCK_MODULES = ['pygtk', 'gtk', 'gobject', 'argparse', 'numpy', 'pandas']
sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
```
https://docs.readthedocs.io/en/stable/faq.html

But if you agree I would stay with removing shapely.

PR #4355, opened 2019-04-11
fixes #4354 - changed database backend for spc geonode to match default geonode

### LINKED ISSUE #4354: SPC Geonode uses Filestorage instead of PostGIS ingest if uploading a Layer
The docker setup for SPC Geonode will save the uploaded shapefile as file on the system, instead of ingesting it into the geodatabase datastore.

## expected behaviour:

The shapefile will get processed and a geometry enabled table created in the database geonode_data. Geoserver will read the table from the geonode_data PostGIS datastore.

## actual behaviour:

The shapefile will get processed and stored in the geoserver datadir as shapefile. Geoserver will serve the file through a shapefile datastore.

---

## current investigation:

SPC Geonode does not use the geonode/postgis-docker container but a plain [mdillon/postgis](https://github.com/GeoNode/geonode/blob/master/scripts/spcgeonode/docker-compose.yml#L156) variant. Therefore the whole initializiation script [initdb-geonode.sh](
https://github.com/GeoNode/postgis-docker/blob/master/initdb-geonode.sh) will not be called. In there the geonode_data geodatabase is created. The tasks.py from [geonode/geonode-docker](https://github.com/GeoNode/geonode-docker/blob/master/tasks.py#L139-L160) is updating the database connection strings for the django db and geodata db and  *substituting* the variables in settings.py through `os.getenv()` calls. Later the variables will be used by geonode (the plain geonode repo - **not** the docker repo 'geonode-docker') entrypoint.sh to migrate and applying the fixtures.

SPC is using [initialize.py](https://github.com/GeoNode/geonode/blob/master/scripts/spcgeonode/django/initialize.py) instead. The `GEODATABASE_URL` parameter will not be set.

I think this is the reason why GeoNode will 'just' upload the files to GeoServer-DataDir instead of ingesting them into a geodatabase.

## possible solution (TBD and tested)

Since settings.py will search for the env var `GEODATABASE_URL` we could just set it via compose variable. The current initialize.py has to be modified to create the geodatabase instance.

gannebamm:
fixes Issue #4354
To enable data ingestion into the postgis database some environmental variables had to be set. Some of the env variables were reformatted. The rework also adds some hints about SSL for GeoServer with commented https endpoints.


scripts/docker/env/production/celery.env
```diff
@@ -39,4 +39,4 @@ MAX_DOCUMENT_SIZE=2
 # GEOSERVER_ADMIN_PASSWORD=admin
 # See https://github.com/geosolutions-it/geonode-generic/issues/28
 # to see why we force API version to 1.24
-DOCKER_API_VERSION="1.24"
+DOCKER_API_VERSION="1.24"
```
frafra: I suppose that there is something wrong with your editor :)

scripts/spcgeonode/docker-compose.yml
```diff
@@ -22,10 +22,25 @@ x-common-django:
...
     - SECRET_KEY=${SECRET_KEY}
+
+    ## UNCOMMENT FOR SSL
+    #- GEOSERVER_PUBLIC_LOCATION=https://${HTTPS_HOST}/gs/
+    #- GEOSERVER_WEB_UI_LOCATION=https://${HTTPS_HOST}/geoserver/
+    #- SITEURL=https://${HTTPS_HOST}
+    ## COMMENT FOR SSL
+    - GEOSERVER_PUBLIC_LOCATION=http://${HTTP_HOST}/gs/
+    - GEOSERVER_WEB_UI_LOCATION=http://${HTTP_HOST}/geoserver/
+    - SITEURL=http://${HTTP_HOST}
```
frafra: This improves the current situation, but it would be a different way to specify URL (using full URLs instead of HTTP{,S}_{HOST_PORT}. I prefer your approach actually, but it would be nice to avoid mixing things together.

scripts/spcgeonode/docker-compose.yml
```diff
@@ -22,10 +22,25 @@ x-common-django:
...
+
+    # database info, ${variables} are read from .env
+    - DATABASE_URL=postgres://${GEONODE_DATABASE}:${GEONODE_DATABASE_PASSWORD}@postgres:5432/${GEONODE_DATABASE}
+    - GEODATABASE_URL=postgis://${GEONODE_GEODATABASE}:${GEONODE_GEODATABASE_PASSWORD}@postgres:5432/${GEONODE_GEODATABASE}
+
     # hardcoded
+    - DEFAULT_BACKEND_UPLOADER=geonode.importer
+    - DEFAULT_BACKEND_DATASTORE=datastore
```
frafra: What about a comment for those who want to store the layers on files?

scripts/spcgeonode/docker-compose.yml
```diff
@@ -22,10 +22,25 @@ x-common-django:
...
+    #- GEOSERVER_WEB_UI_LOCATION=https://${HTTPS_HOST}/geoserver/
+    #- SITEURL=https://${HTTPS_HOST}
+    ## COMMENT FOR SSL
+    - GEOSERVER_PUBLIC_LOCATION=http://${HTTP_HOST}/gs/
+    - GEOSERVER_WEB_UI_LOCATION=http://${HTTP_HOST}/geoserver/
+    - SITEURL=http://${HTTP_HOST}
+
+    # database info, ${variables} are read from .env
+    - DATABASE_URL=postgres://${GEONODE_DATABASE}:${GEONODE_DATABASE_PASSWORD}@postgres:5432/${GEONODE_DATABASE}
+    - GEODATABASE_URL=postgis://${GEONODE_GEODATABASE}:${GEONODE_GEODATABASE_PASSWORD}@postgres:5432/${GEONODE_GEODATABASE}
```
frafra: So, 3 different users for the same database  with two different password? Is there a way to make it simpler?

scripts/spcgeonode/docker-compose.yml
```diff
@@ -153,19 +169,22 @@ services:
...
-    image: mdillon/postgis:9.6-alpine
+    image: geonode/postgis:10
     environment:
       - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
+      - GEONODE_DATABASE=${GEONODE_DATABASE}
+      - GEONODE_DATABASE_PASSWORD=${GEONODE_DATABASE_PASSWORD}
+      - GEONODE_GEODATABASE=${GEONODE_GEODATABASE}
+      - GEONODE_GEODATABASE_PASSWORD=${GEONODE_GEODATABASE_PASSWORD}
+      - DATABASE_URL=postgres://${GEONODE_DATABASE}:${GEONODE_DATABASE_PASSWORD}@postgres:5432/${GEONODE_DATABASE}
+      - GEODATABASE_URL=postgis://${GEONODE_GEODATABASE}:${GEONODE_GEODATABASE_PASSWORD}@postgres:5432/${GEONODE_GEODATABASE}
```
frafra: You could use `x-common-db` similarly to `x-common-django` to avoid duplicated variables.

scripts/spcgeonode/docker-compose.yml
```diff
@@ -153,19 +169,22 @@ services:
 
   # PostGIS database.
   postgres:
-    image: mdillon/postgis:9.6-alpine
+    image: geonode/postgis:10
```
frafra: As this is a custom image, it would be better to have a `build` parameter.

scripts/spcgeonode/docker-compose.yml
```diff
@@ -153,19 +169,22 @@ services:
...
       - database:/var/lib/postgresql/data/
     restart: on-failure
-
-  # Vanilla RabbitMQ service. This is needed by celery
-  rabbitmq:
-    image: rabbitmq:3.7-alpine
-    volumes:
-      - rabbitmq:/var/lib/rabbitmq
-    restart: on-failure
+    stdin_open: true
```
frafra: This option is useful during development/debug, but I would avoid having it set by default.

scripts/docker/env/production/celery.env
gannebamm: oh, I use Sublime on Windows. This file does not need any changes

scripts/spcgeonode/docker-compose.yml
gannebamm: The current geonode/imageName setups do not reside in the main repo. I would have to merge Geonode/postgist-docker into this repo. We should discuss if this is wanted/ smart.

scripts/spcgeonode/docker-compose.yml
gannebamm: For me this is ok: one superuser (postgres), one just for django stuff (genode) and one for the geospatial data (geonode_data).

Actually there are two databases involved: `geonode` for the django stuff and `geonode_data` for spatial data. I just took the default geonode docker setup as concept and inserted it into the spc setup. 

You could do it all with just one account, too. Do you think that would be better?

scripts/spcgeonode/docker-compose.yml
gannebamm: So what exactly is your proposal?

In theory HTTP and HTTPS Host are equal (lets say geonode.domain.com). What is different is the protocoll. One solution could be to use something like:

`GEOSERVER_PUBLIC_LOCATION=${PROTOCOLL}://${HOST}/gs/`

and where HTTP is needed (container to container communication:
`GEOSERVER_LOCATION=http://nginx/geoserver/`

The default geonode composition does use some shell scripting with `if ... fi` blocks to concatenate URLs.

scripts/spcgeonode/docker-compose.yml
gannebamm: Leftover. Will be deleted

scripts/spcgeonode/docker-compose.yml
gannebamm: good idea. Will enhance the setup with more commons and yml anchors.

scripts/spcgeonode/docker-compose.yml
t-book: @gannebamm what about a third database for geofence rules?

t-book: (I think in this context it would suffice to have just one user for Django and postgis, just my two cent)

gannebamm: @t-book The Geofence rules database is a good idea, but imho not part of this pull request.

scripts/spcgeonode/docker-compose.yml
frafra: I would say that we can export those variables in the entrypoint, based on the if/fi block you are mentioning.

scripts/spcgeonode/docker-compose.yml
frafra: No problem having different databases, but I am not sure having more than 2 users (superuser + regular user) would have an impact on security, as long as the same application has access to these passwords.

Nothing against this solution by the way, it is just for discuss and understand if there is something we can improve.

scripts/spcgeonode/docker-compose.yml
frafra: Dockerfiles for spc are inside the repository; I would like to avoid duplication too, and it would make sense to me to have these file in the main repository. We should open a new issue to discuss about that I suppose.

scripts/spcgeonode/docker-compose.yml
gannebamm: Will be set in django/docker-entrypoint.sh

scripts/spcgeonode/docker-compose.yml
gannebamm: Was added as comment to the compose file

scripts/spcgeonode/docker-compose.yml
gannebamm: This would require changes to the dockerfile which currently do not reside in this repository. It should get its own issue with label minor imho

scripts/spcgeonode/docker-compose.yml
gannebamm: Two anchors are not working as discussed on gitter on [April 15, 2019 4:44 PM](https://gitter.im/GeoNode/general?at=5cb498c793fb4a7dc233185b). Which is a pitty.

The proposed setup has only one point of configuration: the global .env file

All changes should be done in there and will be ingested into the appropriate places via docker-compose and shell scripts.

scripts/spcgeonode/docker-compose.yml
gannebamm: I have sent an email to the dev list to discuss this idea

afabiani:
@frafra @gannebamm what is the status of this PR?

gannebamm:
I use the postgis 10 and geodatabase ingest on my setup with an SSL nginx and have not encountered any errors. I do not think the old travis fails are connected to the PR not working tbh, but did not recheck it locally, sorry.

afabiani:
@gannebamm but are still there comments to fix? Is it mergeable in case travis succeed?

@frafra @t-book comments?

gannebamm:
Hi, I have no experience with travis, but the used xml defines:
```
addons:
  hosts:
    - geonode
  postgresql: 9.6
  apt:
    packages:
      - postgresql-9.6-postgis-2.4
      - nginx
  firefox: "latest-esr"
```

as PostgreSQL version which is altered in this commit:
https://github.com/GeoNode/geonode/pull/4355/files/ad656562cb046ed6efa3a92f7ba4f167ea49e4c4#diff-0922f675ec674e962ce7edecf4ef5726

Maybe this leads to the build failing?

frafra:
@gannebamm it is not related to it, because the geonode-selenium test uses its own Docker images and versions, as defined by the compose file. The layer upload failed. You can replicate it locally following the instruction provided by geonode-selenium. I will try to do the same meanwhile.

PR #4366, opened 2019-04-23
[Fixes #4357] tags do not support unicode text

### LINKED ISSUE #4357: tags do not support unicode text
If using unicode characters in tags the tags will get updated, but the metadata assistent form is not able to load anymore:
unicode_tags_error

afabiani:


t-book:
@afabiani  Just tested locally works nice as expected. Thanks!

Bildschirmfoto 2019-04-23 um 17 59 52

Bildschirmfoto 2019-04-23 um 18 01 28

Issue #4376, opened 2019-04-29
UI glitch in explore people - text overflows


gannebamm:
If the postion string is to long it will overflow the container in 12er and 8er bootsstrap:
![grafik](https://user-images.githubusercontent.com/6172324/56887883-ff16a180-6a72-11e9-8772-22e3868b522c.png)

On 6er it is working fine (even though it is not 'nice'):
![grafik](https://user-images.githubusercontent.com/6172324/56887906-0dfd5400-6a73-11e9-8105-11b42095d2d7.png)

t-book:
This should solve it:

```
.profile-details small {
    word-break: break-word;
}
```

gannebamm:
> 
> 
> This should solve it:
> 
> ```
> .profile-details small {
>     word-break: break-word;
> }
> ```
Almost. Should be word-**wrap** : break-word
```
.profile-details small {
    word-wrap: break-word;
}
```

t-book:
Of course! Thanks for pointing!

gannebamm:
closed by #4380

PR #4380, opened 2019-04-30
fixes profile text overflow UI glitch #4376

gannebamm:
added .profile-details small with word-wrap break words as suggested by @t-book

### DIFF
--- a/geonode/static/geonode/css/base.css
+++ b/geonode/static/geonode/css/base.css
@@ -7442,6 +7442,9 @@ article.tab-pane {
   background: white;
   color: #333333;
 }
+.profile-details small {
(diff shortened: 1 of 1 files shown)


t-book:
@gannebamm So this is just the less file, can you rebuild the static files and commit those as well?

gannebamm:
@afabiani @t-book So the PR was allready merged and closed, therefore I am done? Shall I still do anything with the LESS etc pp?

t-book:
I think it would be okay to just wait until somebody else recreates/commits them. But let´s keep it clean. See: https://github.com/GeoNode/geonode/pull/4382

Issue #4092, opened 2018-11-21
Changing style is very slow


capooti:
Changing style from default geoexplorer client in my local GeoNode master instance takes almost 10 seconds. This is what I see in logs:

```
[geoserver_proxy] Updating Style ---> url http://localhost:8080/geoserver/rest/workspaces/geonode/styles/route_2.xml?raw=true
/home/ubuntu/camp-2.10.x/geonode/geonode/geoserver/helpers.py:1593: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  if not elm_user_style_title:
Loading user configuration
running configuration <ConfigParser.SafeConfigParser instance at 0x7f0cfd876e18>
Setting language
Loading custom repository mappings from /home/ubuntu/camp-2.10.x/geonode/geonode/catalogue/backends/pycsw_local_mappings.py
Loading outputschemas
Setting MaxRecordDefault
Querying repository with ids: daf8c75c-e9c0-11e8-ad9e-02d8e4477a33
Request processed
Writing response.
/home/ubuntu/camp-2.10.x/env/local/lib/python2.7/site-packages/owslib/iso.py:117: FutureWarning: the .identification and .serviceidentification properties will merge into .identification being a list of properties.  This is currently implemented in .identificationinfo.  Please see https://github.com/geopython/OWSLib/issues/38 for more information
  FutureWarning)
/home/ubuntu/camp-2.10.x/env/local/lib/python2.7/site-packages/owslib/iso.py:482: FutureWarning: The .keywords and .keywords2 properties will merge into the .keywords property in the future, with .keywords becoming a list of MD_Keywords instances. This is currently implemented in .keywords2. Please see https://github.com/geopython/OWSLib/issues/301 for more information
  FutureWarning)
/home/ubuntu/camp-2.10.x/env/local/lib/python2.7/site-packages/geoserver/style.py:73: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  if not user_style:
/home/ubuntu/camp-2.10.x/env/local/lib/python2.7/site-packages/geoserver/style.py:77: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  if user_style:
Loading user configuration
running configuration <ConfigParser.SafeConfigParser instance at 0x7f0cfc9f9638>
Setting language
[...]

PR #4430, opened 2019-05-23
Add UPDATE_THUMBS_ON_STYLE_CHANGE setting - refs #4092

capooti:


afabiani:
So I added few comments... nevertheless I'm not sure to fully understand the goal of this PR.
It seems to avoid to save the layers accordingly to a variable, but the thumbnails are created on a post-save signal only if the resource has been just created, accordingly to my previous PR.

Are you sure this affects the thumbnail creation and addresses the issue connected to this PR?

capooti:
I can't find your comments @afabiani 
in any case while you are right that now the thumbnail is not generated anymore (thus the name for this variable is inappropriate), it is still very slow a change of style action (even a few seconds) as the layer with current code is saved twice (1755 in helpers.py and 554 in geoserver.views) and for a style change doesn't need to be saved at all as nothing changes in django db, so I would rather suggest to remove that part of the code: what do you think?

To recap, I would remove this line (first unneeded save, as layer will be still saved at 555 in geoserver.views):
https://github.com/GeoNode/geonode/blob/master/geonode/geoserver/helpers.py#L1755

Also I would be inclined to remove the save here:
https://github.com/GeoNode/geonode/blob/master/geonode/geoserver/views.py#L556

If this is required for updating the thumbnails, I would then suggest to return None in affected_layers if we just update a style.

In our instance the difference is dramatic after this change: style editor now works properly, before it was faulty and many times didn't work - particularly when making continue changes in the style editor.

afabiani:
@capooti I agree with you about removing redundant saves. I would close this one and do the changes directly on the lines you indicated above.

capooti:
thanks @afabiani it definitely makes more sense not to have that variable and just remove the unneeded calls.

Issue #4443, opened 2019-05-30
Index page translation issue


t-book:
The current index page has minor issues regarding translations:

1) the base class which is used in body tag is used for styling. It shouldn't be translated:
https://github.com/GeoNode/geonode/blob/master/geonode/templates/index.html#L8

2) the search box placeholder should be translated
https://github.com/GeoNode/geonode/blob/master/geonode/templates/index.html#L43
https://github.com/GeoNode/geonode/blob/master/geonode/templates/base.html#L294

Issue #4440, opened 2019-05-29
Cannot edit profile on mobiles


ppasq:
On small screens (e.g. mobiles) the Resources div in the user profile page seems to cover the list group where many links are placed such as "Edit profile". So  you cannot click on it.

![Screenshot 2019-05-29 at 14 51 36](https://user-images.githubusercontent.com/3024454/58558633-a711bd00-8221-11e9-8444-9b4f50844c15.png)

PR #4444, opened 2019-05-30
Fixes: #4443 issues translations on index

t-book:
The current index page has minor issues regarding translations:

1. the base class which is used in body tag is used for styling. It shouldn't be translated otherwise the css selector will fail:

https://github.com/GeoNode/geonode/blob/master/geonode/templates/index.html#L8

2. the search box placeholder should be translated

https://github.com/GeoNode/geonode/blob/master/geonode/templates/index.html#L43
https://github.com/GeoNode/geonode/blob/master/geonode/templates/base.html#L294

### DIFF
--- a/geonode/people/templates/people/profile_detail.html
+++ b/geonode/people/templates/people/profile_detail.html
@@ -176,7 +176,7 @@
   {% endif %}
 </div>
 
-<div class="col-md-9">
+<div class="col-xs-12 col-md-9">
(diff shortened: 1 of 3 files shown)


t-book:
@afabiani @ppasq Uh sorry I was a bit too fast. Just pushed a commit which solves #4440 
So this is PR is for both issues #4443 and #4440. Let me know if I should separate those commits again.

PR #4446, opened 2019-05-31
[New Docs] GeoNode Core > Ubuntu 18.04

afabiani:


afabiani:
@t-book my comments below:

1. Package installation; yes, doubles must be removed
2. Postgres; I confirm postgis are required only for "geonode_data" unless you have specific requirements
3. Getting an error with openjdk install; ok, I'll investigate this further. Didn't notices those errors.
4. At section: Update the settings in order to use the PostgreSQL Database; I always suggest to do this to users because you never know what they did before. Usually, they skip a lot of passages and steps... so better repeat!

t-book:
Thanks @afabiani I agree with 4.  If @francbartoli @frafra @giohappy do not have further requests  (?) I would merge it.

afabiani:
@t-book before merging would be better to fix the for the points above. Are you taking care of this? Otherwise, I can fix them and merge later on.

t-book:
@afabiani Done for 1–3. https://github.com/GeoNode/geonode/blob/d51aa384c7cc15819cd65bfcc5e5f3c3b8c2492f/docs/install/core/index.rst

Issue #4456, opened 2019-06-05
Thumbnail URL when using non-default file storage


GeoTob:
The URL for thumbnails (and other `ResourceBase` implementations?) is currently generated based on `SITEURL`:

```python
url_path = os.path.join(
    settings.MEDIA_URL,
    upload_to,
    filename).replace(
    '\\',
    '/')
site_url = settings.SITEURL.rstrip('/') if settings.SITEURL.startswith('http') else settings.SITEURL
url = urljoin(site_url, url_path)
```
https://github.com/GeoNode/geonode/blob/master/geonode/base/models.py#L1148 (current master)

However when using a non-standard storage provider (e.g. S3, OpenStack Swift, etc) thumbnails may be served from a different host and thus won't be found by Geonode. 

Django's File Storage API provides a `url(filename)` method to retrieve a full file URL and which should be implemented by all file storage providers: https://docs.djangoproject.com/en/1.8/ref/files/storage/#django.core.files.storage.Storage.url

Using this method would simplify the snippet above to `url = storage.url(upload_path)` and works dynamically against other file storage providers.

Suggest this as an improvement, though unsure whether there's an important reason for the current implementation?

PR #4457, opened 2019-06-05
Use Django file storage API to construct media URL

GeoTob:
Replace custom logic for generating media URLs in ResourceBase.save_thumbnail with call to Django File Storage's `.url` method, to support media file storages hosted remotely.

Fixes https://github.com/GeoNode/geonode/issues/4456

### DIFF
--- a/geonode/base/models.py
+++ b/geonode/base/models.py
@@ -25,7 +25,7 @@ import logging
 import traceback
 
 from pyproj import transform, Proj
-from urlparse import urljoin, urlsplit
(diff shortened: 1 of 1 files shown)


afabiani:
@GeoTob the idea is very good, thanks. Travis is failing actually. I'll try to find some time the next week to review.

GeoTob:
@afabiani thanks, I'm still working on getting the test suite to run locally (and failing so far). The two flake8s above are simple and I can push a change in a few minutes.

PR #4460, opened 2019-06-06
[Draft] Add list of Settings (Needs review and support)

t-book:
This PR adds the full list of settings as we had [here](http://docs.geonode.org/en/master/tutorials/advanced/geonode_settings/settings.html).
Further more explanations and removed contrib settings.

As this still needs a lot of work I would suggest leaving this Draft open until it's finalized. 

**Contributing descriptions and reviewing is very welcome and needed!**

---
**📖 For contributors** 

At each setting, we've started to add  

- the default value
- the env var (if existing)
-  link to external documentation (if the setting is coming from an external app )

---


afabiani:
@t-book I'll merge it for now, let's continue editing this directly to the new-docs beanch

PR #4479, opened 2019-06-12
provide .env template, prevent checking it in (contains secrets)

### LINKED ISSUE #4477: remove spcgeonode .env from git
In spcgeonode the .env file with secrets is under version control, but it better should not be versioned, since checking in changes means having secrets in version control.
Solution: 
* provide a template.env
* users must copy and change it
* add .env to .gitignore
* update the README

I will create a pull request for it.

Marwe:
This pull request closes #4477


t-book:
I think to make this work Travis has to be changed a bit (copy the template first). Otherwise (as it currently does), it will miss env vars for building.

frafra:
Right, but the `.env` file has not been removed either.
While secrets should not be committed and defaults should not be used in production, I wonder if we could start just having this as tip in the documentation, if it is not already there.

afabiani:
@frafra @t-book @Marwe the typical approach on GeoNode is to keep defaults on github and advertise people that they must change them when publishing on a public instance.

There will be a dedicated section on the docs ([here](http://docs.geonode.org/en/master/advanced/index.html#hardening-geonode)) where we will describe all the steps needed and related warning to harden and secure GeoNode.

The "idea" we would like to transmit to the people is that GeoNode is a complex architecture and needs experienced devops to be published on a server. GeoNode developers cannot and won't guarantee for any deployment around unless they have been involved in person.

t-book:
@afabiani D'accord.

Issue #4497, opened 2019-06-13
geoTIFF upload will fail for capital file endings (.TIF/.TIFF)


gannebamm:
## Expected Behavior
A valid geotiff file will get ingested through the web ui at /layers/upload and published as layer.

## Actual Behavior
through the ui you will get a 'Unexpected exception Could not access to uploaded data.'
![grafik](https://user-images.githubusercontent.com/6172324/59428094-9eea7d80-8ddc-11e9-86bc-8576163d861f.png)

the errormessage in django is:

```
django_1          | [pid: 42|app: 0|req: 17/26] 127.0.0.1 () {28 vars in 293 bytes} [Thu Jun 13 22:58:12 2019] GET / => generated 37345 bytes in 149 msecs (HTTP/1.1 200) 8 headers in 401 bytes (1 switches on core 0)
django_1          | [pid: 40|app: 0|req: 6/27] 172.30.218.209 () {44 vars in 828 bytes} [Thu Jun 13 22:58:21 2019] GET /upload/progress => generated 17 bytes in 6 msecs (HTTP/1.1 200) 7 headers in 236 bytes (1 switches on core 0)
nginx_1           | 172.30.218.209 - - [13/Jun/2019:10:58:21 +0000] "GET /upload/progress HTTP/1.1" 200 17 "http://172.30.218.218/layers/upload" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:67.0) Gecko/20100101 Firefox/67.0"
django_1          | uploaded_files: [<InMemoryUploadedFile: UTM2GTIF.TIF (image/tiff)>]
django_1          | valid_extensions: ('TIF',)
django_1          | relevant_files: []
django_1          | Unexpected error in upload step
django_1          | Traceback (most recent call last):
django_1          |   File "./geonode/upload/views.py", line 649, in view
django_1          |     resp = _steps[step](req, upload_session)
django_1          |   File "./geonode/upload/views.py", line 175, in save_step_view
django_1          |     charset=form.cleaned_data["charset"]
django_1          |   File "./geonode/upload/files.py", line 262, in scan_file
django_1          |     raise Exception(_("Could not access to uploaded data."))
django_1          | Exception: Could not access to uploaded data.
django_1          | Traceback (most recent call last):
django_1          |   File "./geonode/upload/views.py", line 649, in view
django_1          |     resp = _steps[step](req, upload_session)
django_1          |   File "./geonode/upload/views.py", line 175, in save_step_view
[...]

PR #4500, opened 2019-06-13
Add postgis v10 fixes Issue #4354

gannebamm:
added postgis v10, switched to geonode/postgis image (and added it to the dockerimage folders for spc), added a second spatially enabled database inside the db-container, made passwords configurable in the .env file, added a second pg_dumper container for the geo-db, updated both pg_dumpers to use alpine3.8, swtiched to geonode.importer instead of rest.importer to enable shp to postgis ingestion.

selenium test will fail caused by #4497


francbartoli:
I've merged the change with the lowercase suffix, let's see what happens

gannebamm:
From Travis
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install postgresql-9.6-postgis-2.4 nginx" failed and exited with 100 during .

Seems not related to my changes and btw geonode core does use postgis 10 and spc with my changes in this pr, too. Can't see why anything should pull postgres 9.6

gannebamm:
Failed with

Unable to locate element: //button[contains(., 'Editing Tools')]


Again not related to my code

frafra:
Everything is working on Travis CI. I have seen that error on the regular Docker setup, not with SPC, that is the one currently used by Travis.

gannebamm:
triggered an merge with master. As frafra said the Travis CI checks were successfull last time.

frafra:
What about importing existing files? Is there an easy way to digest previously inserted layers?

afabiani:
@gannebamm tests have passed but I still see a pending comment from @frafra

gannebamm:
@frafra about transforming the shape file layers to postgis ones:

I would like to make a gnip for that which is able to pull the layers info and data from one geonode and ingest it into an other. This would make an easy reingest of the files working. What do you think?

frafra:
> @frafra about transforming the shape file layers to postgis ones:
> 
> I would like to make a gnip for that which is able to pull the layers info and data from one geonode and ingest it into an other. This would make an easy reingest of the files working. What do you think?

I think that being able to easily import and export layer would be a feature that could be used in many different scenarios 👍

PR #4509, opened 2019-06-13
changed basemap for thumbnails to wikimedia, fixes #4459

### LINKED ISSUE #4459: Thumbnail basemap not rendering
Not sure if I'm the only one affected by this, but thumbnails are currently rendering without a basemap for me. Thumbnails on the official geonode demo also don't show a basemap currently.

Looking at the GeoServer logs I noticed the following warning:

```
Error fetching : http://a.tile.openstreetmap.org/xx/xxxxxx/xxxxxx.png, status is : 403
```

Loading the URL manually works just fine.

Searching for the issue online revealed this very recent answer on SO: https://stackoverflow.com/a/56440427/4033315 which hints at OSM blocking certain requests. Not sure if this is the problem here?

I changed `THUMBNAIL_GENERATOR_DEFAULT_BG` in `settings.py` to a different provider and got this to work again:

```python
THUMBNAIL_GENERATOR_DEFAULT_BG = r"https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png"
```

I currently can't spend the time to check the request headers sent by the `printng` extension to see if they complies with OSM policy, but maybe someone else can?

In the meantime I suggest switching the default for `THUMBNAIL_GENERATOR_DEFAULT_BG` to a more forgiving server.

gannebamm:
This PR will change the default **thumbnail** basemap to wikimedia. This will result in some legal constraints, eg:

**You License Freely Your Contributions** – you generally must license your contributions and edits to our sites or Projects under a free and open license (unless your contribution is in the public domain).

The full Terms of Use is listed here: https://foundation.wikimedia.org/wiki/Terms_of_Use/en

It fixes openstreet map tileserver responding with a 404 for the thumbnail creation requests as mentioned in #4459 .

Additionally the wikimedia thumbnails have a much better resolution which results in nicer thumbnails for intermediate sized devices (bootstrap md)

BEFORE MERGING: Please read the full terms of usage


gannebamm:
Travis fails due to
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install postgresql-9.6-postgis-2.4 nginx" failed and exited with 100 during.. 

Not related to the pr

frafra:
Nice!
Have a look at https://foundation.wikimedia.org/wiki/Maps_Terms_of_Use too.
We need to update the attribution string too.

gannebamm:
@frafra yes, but notice this will **only** change the tiles used for thumbnails. The basemap in mapstore/geoexplorer will stay the same.

gannebamm:
checks had passed succesful:
https://travis-ci.org/GeoNode/geonode/builds/546347826?utm_source=github_status&utm_medium=notification

triggered a merge with global master ..

frafra:
> 
> 
> @frafra yes, but notice this will **only** change the tiles used for thumbnails. The basemap in mapstore/geoexplorer will stay the same.

Right. I think we should have attribution on all these previews, as we are violating https://www.openstreetmap.org/copyright (which has to be followed for Wikipedia maps too). A different issue+PR maybe? :-D

frafra:
I am discussing with the Italian OSM community about that. Someone says that we could skip attribution on the preview, but only if that is a preview: if we use Wikipedia maps for the thumbnail and OSM maps for the data, I guess we will have to credit Wikipedia.

We can switch entirely to Wikipedia or attribute Wikipedia on the thumbnails I guess.

afabiani:
@frafra any chance to have confirmation that the previews do not need attribution? I don't think so as they are static snapshots, but I'm not sure.

We should probably add some disclaimers at the bottom of the portal or somewhere on some dedicated page. This is probably part of a new PR as it requires some additional work and decisions.

frafra:
@afabiani cannot confirm, but they should not.

I would say that we can approve this PR, as the licensing is exactly the same as OSM tiles and do another one when we are certain about how to behave in this kind of situation.

PR #4591, opened 2019-06-29
added ansible to docs

t-book:
Dear @afabiani 

I've merged the two existing ansible docs into one and removed some parts (but referenced the ansible docs where needed). 

Further added a section about galaxy role geonode.geonode and an example playbook. As the ansible role targets geonode project, the docs make most sense under /install/project to me. In kind, you prefer them elsewhere no problem.

It would be nice too especially review the role and playbook part ( I use the official stack very rarely ).

For your convenience find here the PR-build:
http://csgis.de/2load/html/start/quick/index.html#install-via-ansible
http://csgis.de/2load/html/install/project/index.html#install-with-ansible

### DIFF
--- a/docs/install/project/index.rst
+++ b/docs/install/project/index.rst
@@ -303,3 +303,223 @@ Run the containers in daemon mode
 .. code-block:: shell
 
   docker-compose -f docker-compose.yml -f docker-compose.override.example-org.yml up --build -d
+
+.. _install-with-ansible:
+
+Ansible
+=======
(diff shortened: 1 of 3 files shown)


afabiani:
Thanks, @t-book very nice.

About the docs organization, feel free to move sections as you think it would be better. Mine is just a proposal but definitively not the best one.

I'll try to test and improve the process if needed before the final release. That may take some time though.

PR #4600, opened 2019-07-01
added numpy for SPC

t-book:
see: https://github.com/GeoNode/geonode/pull/4599/files


t-book:
@francbartoli true! Sorry, this was a misunderstanding but you're absolutely right!

Issue #4726, opened 2019-08-09
GNIP: curated_thumbs contrib application


capooti:
Hello GeoNode developers

we have the following requirement: in many cases GeoNode users would like to be able to upload a static image that can be used in place of the automatically GeoNode generated thumbnail.

We would like to develop a contrib application which, when added in INSTALLED_APPS, will act in the following way:

* in the edit layer/map form, there will be a new button, named "Upload" in the Thumbnail section
* clicking on the button, the user will get to an "Upload thumbnail" page, identical for layers and maps
* the user will be able to upload a thumbnail to be associated with the layer/map. If an image is uploaded, then it will be used as the thumbnail for that layer/map. Otherwise the usual GeoNode behaviour will keep going

How GeoNode would be impacted when the contrib application is used:

* there won't be any schema change in existing tables
* there will be a new model (table) which will keep a 1-1 relation with base_resourcebase
* thumbnail_url will be dehydrated in geonode/api/resourcebase_api.py (only when the extension is used) in order to return the static image url in place of the thumbnail
* the edit layer/map form will have a new "upload thumbnail" button
* a new page with the upload thumbnail feature will be available

Please provide your feedback (maybe with changes/improvements), and I would soon send a PR with this new contrib application.

t-book:
Thanks, @capooti  I would very welcome this. +1.

_It would be great to have some settings option which filesize and types are allowed (maybe even  svg :)) By the way have you seen there is an upload option in mapstore  for thumbs as well nevertheless I like the idea of a viewer independent contrib app._

afabiani:
@capooti +1 nice feature. 

There's already a similar feature available for Documents by default since a long time

![image](https://user-images.githubusercontent.com/1278021/62864220-68ad7980-bd0b-11e9-977d-9445cf75c23f.png)

![image](https://user-images.githubusercontent.com/1278021/62864242-7531d200-bd0b-11e9-90d1-bd527b739ad7.png)

t-book:
@capooti  As the PR has already been merged. Can we close this issue?

capooti:
yes let's close this @t-book

PR #4731, opened 2019-08-13
Implements GNIP for curated-thumbs app

capooti:
This PR implements the GNIP for curated-thumbs #4726 

Here is an animated gif with the new feature:
http://g.recordit.co/JaKutevbrT.gif

It has an additional requirement: django-imagekit

Thumbnails are implemented for layers and maps. It would be very easy to implement them for documents as well but I see that there is already this feature (let me know if you want to replace it with this for homogeneity).

I decided to implement the application as an internal application for GeoNode as I believe this is useful for any GeoNode instance. If you believe it should rather go to a contrib application, I am ready to migrate it to geonode-contribs.


afabiani:
... moreover, I don't see any test

geonode/curated_thumbs/tests.py
```diff
@@ -0,0 +1 @@
+# -*- coding: utf-8 -*-
```
afabiani: Missing headers

afabiani: Missing tests

capooti:
Thanks Alessio
having this in Base definitely makes sense. I will also include tests and documentation.
I am closing this PR and will send a PR
cheers

PR #4736, opened 2019-08-14
Gnip 4726

capooti:
This is the PR implementing curated thumbnails for layers, maps and documents.
This is now done in geonode.base (and not in a different app) and I have added some documentation.

@afabiani I am trying to write some tests (in base/tests.py), however could you please point out what is now the recommended approach?

If I run "manage.py test geonode.base" it hangs...


afabiani:
@capooti try

`coverage run --branch --source=geonode manage.py test geonode.base.tests`

capooti:
@afabiani using that line I believe it is trying to connect to GeoServer and get stuck

what is now the recommended way to run tests at this time with GeoNode?
Once it was possible to run the whole suite using "paver run_tests" and to run tests for just an application following the django standard ("manage.py test appname"). Would love to learn the difference

thanks a lot!

t-book:
Dear @capooti hopefully I do not prevent @afabiani from answering as I can surely learn more on that. For local testing what works for me is setting the env var first:

```
export GEOSERVER_SERVER_URL=http://localhost:8080/geoserver/
paver start
./manage.py test geonode.base.tests  --noinput  --parallel=1
```
(You might set in the correct settings as well `DJANGO_SETTINGS_MODULE=geonode.settings`)
I often look at what travis does https://travis-ci.org/GeoNode/geonode/jobs/571890658 to see which vars needed. There might be an easier way ...

capooti:
Thanks, will try that @t-book and let you know
I believe restoring a run_tests command in paver - if possible - would be still a great thing

t-book:
> I believe restoring a run_tests command in paver - if possible - would be still a great thing

Definitely! And most likely there already is. Settings the env var before is just my way to do as I do not know better ;)

Issue #4713, opened 2019-07-31
Attributes are not added in layers_attributes table when a layer is uploaded


capooti:
## Expected Behavior

A shapefile is uploaded to GeoNode. If the layer is successfully uploaded, there must be one record in the layers_attributes table of each shapefile field.

## Actual Behavior

Randomly the attributes are not created. When this is happening I am observing that the problem seems occurring at this line: 
https://github.com/GeoNode/geonode/blob/master/geonode/geoserver/helpers.py#L915

```python
req, body = http_client.get(dft_url)
```

checking the body value after this line when the error is occurring this is what I see:

```
<?xml version="1.0" ?>
<ServiceExceptionReport
   version="1.2.0"
   xmlns="http://www.opengis.net/ogc"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://www.opengis.net/ogc
http://schemas.opengis.net/wfs/1.0.0/OGC-exception.xsd">
   <ServiceException code="InvalidParameterValue">
      Could not find type: {http://www.geonode.org/}subwaylines_p_odp_4
</ServiceException></ServiceExceptionReport>
```

It looks like a problem of GeoNode/GeoServer communication. In our case everything works well with GeoNode/GeoServer interaction (data edit, style edit ecc) but this is still randomly occurring.

## Steps to Reproduce the Problem

Can't provide a specific sequence of steps. This is occurring randomly on 3 different instance we administer.

## Specifications

  - GeoNode version: master
  - Installation method (manual, GeoNode Docker, SPCGeoNode Docker): manual
  - Platform: Ubuntu 18.04 LTS
  - Additional details:

capooti:
@afabiani until we cannot find a solution to this, how about if I send a PR with attributes synced from GDAL on layer upload?

afabiani:
@capooti so your idea would be to dump the attributes list by using GDAL?

capooti:
@afabiani would it be a big work to include attributes in the layer definition from the GeoServer REST endpoint? that way we could use gsconfig for the purpose
Using gdal is more complicated particularly once the layer is already uploaded to GeoNode
Any idea why the HttpClient would fail? I am struggling why that is happening

capooti:
by the way @afabiani I noticed demo is also affected: https://dev.geonode.geo-solutions.it/layers/geonode:roads0#/

[...]

PR #4750, opened 2019-08-19
Fixes for #4713

capooti:
This will use PostGIS metadata - only if a layer is in PostGIS - to create the attribute_map if this is created empty from GeoServer because of authentication errors.

### DIFF
--- a/geonode/geoserver/helpers.py
+++ b/geonode/geoserver/helpers.py
@@ -799,6 +799,54 @@ def get_stores(store_type=None):
     return store_list
 
 
+def get_attribute_map_from_postgis(layer):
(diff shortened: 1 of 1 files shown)


afabiani:
Hi @capooti so, I have taken a look quickly to this issue, see my comments below:

1. The request for the attributes map to GeoServer fails because user's credentials should be passes to the "http_client.get" method, as the "proxy" method does.

2. We need a test case for this also. Integration tests currently do setup a Postgis DB. We can test the method through them.

I'll try to send a patch for the point (1) today.

afabiani:
Closing this for now.

PR #5045, opened 2019-10-15
[Fixes #1784][Fixes #2161][Fixes #2164][Fixes #5052]

### LINKED ISSUE #1784: Private group showing info without being logged in.
--Without being logged in, i can see all the info concerning the group but i created it as Private, and i think it shouldn't.--

info without log in

### LINKED ISSUE #2161: Private groups (and members) still visible
The new functionalities of the groups in Geonode allow to set a group as "private":
_Private: Registered users cannot see any details about the group, including membership. Only invited users can join_
My understanding is that a non registered user should not see whether a private group is present nor who its members are.
But the current behaviour is that, even setting the access as "private", the groups and their members are still visible for unregistered users.

### LINKED ISSUE #2164: Admin roles inside a private group doesn't handle correctly the visibility of members
This is related to issues #1784 and #2161 and splits the first into this bug for version 2.4 and moves the suggested fine-grained authorization model proposed by @state-hiu into the 2.5 milestone:

roles: notMember <----{inherits} (trivial though)--- Member <----{inherits}--- Manager <----{inherits}--- Superuser
- **Member** cannot see others with the same role (automatically the more privileged roles), _working yet_
- **Manager** has to see all members (_not working_) and can see all managers (_not implemented_)
- **Superuser** has to see any role of the private groups (_not working_)

### LINKED ISSUE #5052: "Registered Members" Group: Auto assign users to group(s) after activation
Currently there's no concept of "loggedin users" within the GeoNode permission system. Registered users without a group assigned are treated like anonymous users.

Assigning sepcific permissions to loged in users can be overcome creating a "loggedin users" group and assign active users to the group, but this must be done manually (and worst it must be done one by one).

The proposal here is to implement a post_save signal hook in GeoNode to auto-assign users to one or more groups when a new user is registered.

**Describe alternatives you've considered**
### OPTION 1
We could have this new settings:

`AUTO_ASSIGN_USERS_TO_GROUPS = ['group1', 'group2', etc.]`

This is a bit unflexible since when group names change this settings must be updated and the application reloaded.

### OPTION 2
Create a new "**UserAutoAssignedGroups**":

```
class UserAutoAssignedGroups(models.Model):
     groups =  models.ForeignKey(django.contrib.auth.models.Group)
```

which will be queried within the post_save hook.

afabiani:
* Fixes #1784
* Fixes #2161
* Fixes #2164
* Fixes #5052

This PR adds the following improvements to GeoNode Groups:

 1. Hide "private" groups to "non-members"
    The groups with access of type "private" won't be listed/visible to non-members at all.

 2. A "superuser" can always "manage" a Group, despite it is a member or not
   It won't be possible to exclude a "superuser" from a Group anymore.

 3. "Registered Members" Group concept
   By default, GeoNode will create a group called "Registered Members" allowing a resource owner to provide access to all members.
   At login, GeoNode will check for active membership to this default group.
   This group won't be accessible nor visible to anonymous users.

### DIFF
--- a/docs/basic/settings/index.rst
+++ b/docs/basic/settings/index.rst
@@ -216,10 +216,18 @@ AUTH_EXEMPT_URLS
     ``AUTH_EXEMPT_URLS = ('/maps',)`` will allow unauthenticated users to
     browse maps.
 
+AUTO_ASSIGN_REGISTERED_MEMBERS_TO_REGISTERED_MEMBERS_GROUP_NAME
+---------------------------------------------------------------
+
+    | Default: ``False``
+    | Env: ``AUTO_ASSIGN_REGISTERED_MEMBERS_TO_REGISTERED_MEMBERS_GROUP_NAME``
+
+    Auto assign users to a default "Registered Users" private group after activation.
+
 AUTO_GENERATE_AVATAR_SIZES
 --------------------------
 
-    Default: ``20, 30, 32, 40, 50, 65, 70, 80, 100, 140, 200, 240``
+    | Default: ``20, 30, 32, 40, 50, 65, 70, 80, 100, 140, 200, 240``
 
     An iterable of integers representing the sizes of avatars to generate on upload. This can save rendering time later on if you pre-generate the resized versions.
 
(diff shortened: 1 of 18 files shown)


davekennewell:
All three of these objectives would be super-useful. Thankyou for the work Alessio.

gannebamm:
One question about the "Active Members" Group concept:

What does tag an user to be active? It reads like everyone registered is labeled as 'active'. Which is ok, but I would change the name to 'registered users' or whatever, since active sounds like they participate in the portal somehow.

afabiani:
@gannebamm yes, the idea is to include all the "registered" members.

PR #5292, opened 2019-11-20
Added kikislater to contributors

t-book:
@afabiani cannot push directly here. let me know if we need an issue here and I will open one and add checklist ...


kikislater:
Thanks !

PR #5884, opened 2020-03-16
add abstract and purpose to filtering setup on reasource base

petrus7:


petrus7:
should not be create against geonode master

gannebamm:
@petrus7 @giohappy nonetheless this would be a nice feature. Please let me know if you could think of this as upstream feature and if you would need help.

giohappy:
@gannebamm yes it will be merged upstream but the PR was created too early.
It will be submitted again soon.

PR #6083, opened 2020-05-13
remove empty result dropdown

petrus7:
remove empty results from autocompletes


petrus7:
close made agains wrong branch

PR #6425, opened 2020-09-09
[Issue #6414] File system operations do not adhere to Django file storage API

### LINKED ISSUE #6414: File system operations do not adhere to Django file storage API
## Expected Behavior
All file system operations correctly utilise [Django's file storage API](https://docs.djangoproject.com/en/2.2/ref/files/storage/) allowing third-party Django plugins such as [django-storage-swift](https://github.com/dennisv/django-storage-swift) to function correctly if they implement Django's documented functions.

## Actual Behavior
Not all file system operations adhere to the Django file storage API causing issues when third-party extensions are used. For example, [line 1319 of `base/models.py`](https://github.com/GeoNode/geonode/blob/master/geonode/base/models.py#L1319) attempts a glob at the specified thumbnail path in the filesystem. If this filesystem is remotely hosted using [django-storage-swift](https://github.com/dennisv/django-storage-swift) or similar, an exception is thrown and the thumbnail fails to generate.

  1. Activate your GeoNode virtual environment and run `pip install django-storage-swift`
  2. Edit your `settings.py` and add parameters listed in the [django-storage-swift documentation](https://github.com/dennisv/django-storage-swift/blob/master/README.rst). You do not need to specify static file storage parameters.
  3. Load GeoNode and add a layer
  4. Observe failure to generate the thumbnail resulting in the default being used (subsequent 'Set Thumbnail' operations will also fail)

## Specifications

  - GeoNode version: 3.0
  - Installation method (manual, GeoNode Docker, SPCGeoNode Docker): manual
  - Platform: Ubuntu 18.04 LTS
  - Additional details:

mtnorthcott:
Fixes #6414 

PR to replace file system operations performed using Python's `os`, `shutil` etc. library functions with the more generic [Django file storage API](https://docs.djangoproject.com/en/2.2/ref/files/storage/) function calls with the goal of supporting Django storage plugins such as [django-storage-swift](https://github.com/dennisv/django-storage-swift).

## Checklist

For all pull requests:

- [x] Commits changing the **settings**, **UI**, **existing user workflows**, or adding **new functionality**, need to include documentation updates

### DIFF
--- a/geonode/base/models.py
+++ b/geonode/base/models.py
@@ -1307,6 +1307,10 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin, ItemBase):
     # that indexing (or other listeners) are notified
     def save_thumbnail(self, filename, image):
         upload_path = os.path.join('thumbs/', filename)
+        local_thumbs = os.path.join(settings.MEDIA_ROOT, "thumbs")
+        if not os.path.exists(local_thumbs):
+            os.makedirs(local_thumbs)
+
         try:
             # Check that the image is valid
             from PIL import Image
--- a/geonode/base/models.py
+++ b/geonode/base/models.py
@@ -74,6 +74,9 @@ from geonode.notifications_helper import (
     send_notification,
     get_notification_recipients)
 from geonode.people.enumerations import ROLE_VALUES
+from geonode.base.thumb_utils import (
+    thumb_path,
+    remove_thumbs)
 
 from pyproj import transform, Proj
 
@@ -1306,10 +1309,7 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin, ItemBase):
     # Note - you should probably broadcast layer#post_save() events to ensure
     # that indexing (or other listeners) are notified
     def save_thumbnail(self, filename, image):
-        upload_path = os.path.join('thumbs/', filename)
-        local_thumbs = os.path.join(settings.MEDIA_ROOT, "thumbs")
-        if not os.path.exists(local_thumbs):
-            os.makedirs(local_thumbs)
+        upload_path = thumb_path(filename)
 
         try:
             # Check that the image is valid
@@ -1319,18 +1319,14 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin, ItemBase):
             im = Image.open(content_data)
             im.verify()  # verify that it is, in fact an image
 
-            thumbnail_name, ext = os.path.splitext(filename)
-            _, _thumbs = storage.listdir("thumbs")
-            for _thumb in _thumbs:
-                if _thumb.startswith(thumbnail_name):
-                    storage.delete(os.path.join("thumbs", _thumb))
-                    logger.debug("Deleted existing thumbnail: " + _thumb)
+            name, ext = os.path.splitext(filename)
+            remove_thumbs(name)
 
             if upload_path and image:
                 actual_name = storage.save(upload_path, ContentFile(image))
                 url = storage.url(actual_name)
                 _url = urlparse(url)
-                _upload_path = os.path.join('thumbs/', os.path.basename(_url.path))
+                _upload_path = thumb_path(os.path.basename(_url.path))
                 if upload_path != _upload_path:
                     if storage.exists(_upload_path):
                         storage.delete(_upload_path)
--- a/geonode/base/utils.py
+++ b/geonode/base/utils.py
@@ -30,7 +30,6 @@ from datetime import datetime, timedelta
 # Django functionality
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.core.files.storage import FileSystemStorage
 from django.core.files.storage import default_storage as storage
 
 # Geonode functionality
@@ -55,23 +54,22 @@ def delete_orphaned_thumbs():
     """
     Deletes orphaned thumbnails.
     """
-    if isinstance(storage, FileSystemStorage):
-        documents_path = os.path.join(settings.MEDIA_ROOT, 'thumbs')
-    else:
-        documents_path = os.path.join(settings.STATIC_ROOT, 'thumbs')
-    if os.path.exists(documents_path):
-        for filename in os.listdir(documents_path):
-            fn = os.path.join(documents_path, filename)
-            model = filename.split('-')[0]
-            uuid = filename.replace(model, '').replace('-thumb.png', '')[1:]
-            if ResourceBase.objects.filter(uuid=uuid).count() == 0:
-                print('Removing orphan thumb %s' % fn)
-                logger.debug('Removing orphan thumb %s' % fn)
-                try:
-                    os.remove(fn)
-                except OSError:
-                    print('Could not delete file %s' % fn)
-                    logger.error('Could not delete file %s' % fn)
+    deleted = []
+    _, files = storage.listdir("thumbs")
+
+    for filename in files:
+        model = filename.split('-')[0]
+        uuid = filename.replace(model, '').replace('-thumb.png', '')[1:]
+        if ResourceBase.objects.filter(uuid=uuid).count() == 0:
+            logger.debug("Deleting orphaned thumbnail " + filename)
+            try:
+                storage.delete(os.path.join("thumbs", filename))
+                deleted.append(filename)
+            except NotImplementedError as e:
+                logger.error(
+                    "Failed to delete orphaned thumbnail '{}': {}".format(filename, e))
+
+    return deleted
 
 
 def remove_duplicate_links(resource):
--- a/geonode/base/utils.py
+++ b/geonode/base/utils.py
@@ -22,7 +22,6 @@
 """
 
 # Standard Modules
-import os
 import logging
 from dateutil.parser import isoparse
 from datetime import datetime, timedelta
@@ -30,7 +29,6 @@ from datetime import datetime, timedelta
 # Django functionality
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.core.files.storage import default_storage as storage
 
 # Geonode functionality
 from guardian.shortcuts import get_perms, remove_perm, assign_perm
@@ -41,6 +39,9 @@ from geonode.base.models import ResourceBase, Link, Configuration
 from geonode.geoserver.helpers import ogc_server_settings
 from geonode.maps.models import Map
 from geonode.services.models import Service
+from geonode.base.thumb_utils import (
+    get_thumbs,
+    remove_thumb)
 
 logger = logging.getLogger('geonode.base.utils')
 
@@ -55,7 +56,7 @@ def delete_orphaned_thumbs():
     Deletes orphaned thumbnails.
     """
     deleted = []
-    _, files = storage.listdir("thumbs")
+    files = get_thumbs()
 
     for filename in files:
         model = filename.split('-')[0]
@@ -63,7 +64,7 @@ def delete_orphaned_thumbs():
         if ResourceBase.objects.filter(uuid=uuid).count() == 0:
             logger.debug("Deleting orphaned thumbnail " + filename)
             try:
-                storage.delete(os.path.join("thumbs", filename))
+                remove_thumb(filename)
                 deleted.append(filename)
             except NotImplementedError as e:
                 logger.error(
--- a/geonode/base/utils.py
+++ b/geonode/base/utils.py
@@ -22,6 +22,7 @@
 """
 
 # Standard Modules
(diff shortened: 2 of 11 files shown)


mtnorthcott:
> Hi @mtnorthcott it's been a bit difficult to understand when a PR is ready for the review or not.
> 
> I'm kindly asking to either create just one PR for each issue, with no further commits, or put the PRs in DRAFT until they are still a work in progress. Thanks for your understanding.

@afabiani I had intended to tag the description with 'Closes #6414' once complete and make a comment but I did not communicate this so my apologies for the confusion. I have tagged this PR as a draft for the time being.

mtnorthcott:
I have done some testing and, although there are likely instances of local storage operations that I've missed, this PR is now in a state in which feature-breaking bugs are fixed for third-party plugins as described. This predominantly includes creation of documents and document thumbnails using `unoconv`, layer thumbnail generation and orphaned thumbnail deletion. I am happy for @afabiani to review as time permits. Please let me know of any changes you would like made.

afabiani:
@mtnorthcott for the time being I see some tests failing on Travis like this 

```
======================================================================
ERROR: test_create_document_url_view (geonode.documents.tests.DocumentsTest)
----------------------------------------------------------------------
  File "/home/travis/build/GeoNode/geonode/geonode/documents/tests.py", line 140, in test_create_document_url_view
    response = self.client.post(reverse('document_upload'), data=form_data)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/test/client.py", line 543, in post
    response = super().post(path, data=data, content_type=content_type, secure=secure, **extra)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/test/client.py", line 357, in post
    secure=secure, **extra)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/test/client.py", line 422, in generic
    return self.request(**r)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/test/client.py", line 503, in request
    raise exc_value
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/views/generic/base.py", line 97, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/views/generic/edit.py", line 172, in post
    return super().post(request, *args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/views/generic/edit.py", line 142, in post
    return self.form_valid(form)
  File "/home/travis/build/GeoNode/geonode/geonode/documents/views.py", line 240, in form_valid
    self.object.save(notify=True)
  File "/home/travis/build/GeoNode/geonode/geonode/base/models.py", line 913, in save
    super(ResourceBase, self).save(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/polymorphic/models.py", line 91, in save
    return super(PolymorphicModel, self).save(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/db/models/base.py", line 741, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/db/models/base.py", line 790, in save_base
    update_fields=update_fields, raw=raw, using=using,
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/dispatch/dispatcher.py", line 175, in send
    for receiver in self._live_receivers(sender)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/django/dispatch/dispatcher.py", line 175, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "/home/travis/build/GeoNode/geonode/geonode/documents/models.py", line 220, in create_thumbnail
    result = create_document_thumbnail.delay(object_id=instance.id)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/celery/app/task.py", line 426, in delay
    return self.apply_async(args, kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/celery/app/task.py", line 564, in apply_async
    link=link, link_error=link_error, **options)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/celery/app/task.py", line 776, in apply
    ret = tracer(task_id, args, kwargs, request)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/celery/app/trace.py", line 429, in trace_task
    I, R, state, retval = on_error(task_request, exc, uuid)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/celery/app/trace.py", line 412, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/travis/build/GeoNode/geonode/geonode/documents/tasks.py", line 81, in create_document_thumbnail
    image_file.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

geonode/base/models.py
```diff
@@ -1307,6 +1307,10 @@ def has_thumbnail(self):
     # that indexing (or other listeners) are notified
     def save_thumbnail(self, filename, image):
         upload_path = os.path.join('thumbs/', filename)
+        local_thumbs = os.path.join(settings.MEDIA_ROOT, "thumbs")
+        if not os.path.exists(local_thumbs):
+            os.makedirs(local_thumbs)
+
```
GeoTob, line 1313: This needs fixing as discussed. If we want to be storage independent, all file and path changes should go through Django's storage system.

GeoTob (changes requested):

geonode/base/utils.py
```diff
@@ -55,23 +55,22 @@ def delete_orphaned_thumbs():
...
-                except OSError:
-                    print('Could not delete file %s' % fn)
-                    logger.error('Could not delete file %s' % fn)
+    deleted = []
+    files = get_thumbs()
+
+    for filename in files:
+        model = filename.split('-')[0]
+        uuid = filename.replace(model, '').replace('-thumb.png', '')[1:]
+        if ResourceBase.objects.filter(uuid=uuid).count() == 0:
```
GeoTob, line 64: This will make one DB request per file, which can significantly slow things down. Since you're already re-writing this method I suggest something like below:

```python
def get_uuid(filename):
    # You may want to think about error handling here, ie what if the
    # `filename` argument doesn't follow the convention you expect
    # (if that's a situation that your code can feasibly run into)?
    # Maybe regex is better to validate correct format and extract the bit you need?
    model = filename.split('-')[0]
    return filename.replace(model, '').replace('-thumb.png', '')[1:]

...
# Get a set of UUIDs (to ensure uniqueness) 
uuids = {get_uuid(name) for name in get_thumbs()}

# Match uuids against the DB in a single query. This is likely better than executing multiple queries,
# however note that a) depending on which UUID version is used such DB  fields can be slow to look up,
# b) if the list of uuids is very long the `IN` queries constructed by Django may not be the fastest.
db_uuids = ResourceBase.objects.filter(uuid__in=uuids).values_list('uuid', flat=True)

# Iterate over all orphaned uuids (without a DB match):
for uuid in uuids - set(db_uuids):
    try:
        remove_thumb(uuid)
        deleted.append(uuid)
    except NotImplementedError as e:
        logger.error("Failed to delete orphaned thumbnail '{}': {}".format(filename, e))

return deleted
```

You could go even further by making your `remove_thumb` handle & log errors internally and return either the removed file's uuid or `None` on failure. In that case the last few lines of this method would like below. However, I'm not sure whether that's worth the extra effort.

```python
orphaned_uuids = uuids - set(db_uuids)
deleted = (remove_thumb(uuid) for uuid in orphaned_uuids)
return [uuid for uuid in deleted if uuid is not None]
```

geonode/documents/renderers.py
```diff
@@ -58,28 +58,23 @@ def render_document(document_path, extension="png"):
...
+    if extension != "pdf" and guess_mimetype(document_path) != 'application/pdf':
+        document_path = render_document(document_path, extension="pdf")
+        dispose_input = True
 
     # spawn subprocess and render the document
-    output = NamedTemporaryFile(suffix='.{}'.format(extension))
+    output_path = None
     if settings.UNOCONV_ENABLE:
         timeout = None
+        _, output_path = tempfile.mkstemp(suffix=".{}".format(extension))
```
GeoTob, line 70: Why use `mkstemp` instead of `NamedTemporaryFile` as in the original code? I don't see the improvement?

geonode/documents/renderers.py
```diff
@@ -58,28 +58,23 @@ def render_document(document_path, extension="png"):
...
-                return process.kill()
-
             unoconv = subprocess.Popen(
                 [settings.UNOCONV_EXECUTABLE, "-v", "-e", "PageRange=1-2",
-                    "-f", extension, "-o", output.name, temp_path],
+                    "-f", extension, "-o", output_path, document_path],
                 stdout=subprocess.PIPE, stderr=subprocess.PIPE
             )
-            timeout = Timer(settings.UNOCONV_TIMEOUT, kill, [unoconv])
+            timeout = Timer(settings.UNOCONV_TIMEOUT, lambda p: p.kill(), [unoconv])
```
GeoTob, line 77: Isn't this equivalent to the following (I'm not actually sure):

```python
timeout = Timer(settings.UNOCONV_TIMEOUT, unoconv.kill)
```

geonode/qgis_server/tests/test_helpers.py
```diff
@@ -251,22 +252,21 @@ def test_delete_orphan(self):
         # register file list
         layer_path = settings.QGIS_SERVER_CONFIG['layer_directory']
         tiles_path = settings.QGIS_SERVER_CONFIG['tiles_directory']
-        geonode_layer_path = os.path.join(settings.MEDIA_ROOT, 'layers')
 
-        qgis_layer_list = set(os.listdir(layer_path))
-        tile_cache_list = set(os.listdir(tiles_path))
-        geonode_layer_list = set(os.listdir(geonode_layer_path))
+        qgis_layers = set(os.listdir(layer_path))
```
GeoTob, line 256: I wouldn't usually go as far as renaming existing variables in a large project like this just for the sake of semantics. 

Keep it for now given that this is in a test file and not used anywhere, but I think the improvements are rather minimal :)

geonode/qgis_server/tests/test_helpers.py
```diff
@@ -251,22 +252,21 @@ def test_delete_orphan(self):
...
+        tile_caches = set(os.listdir(tiles_path))
+        _, geonode_layers = map(set, storage.listdir("layers"))
```
GeoTob, line 258: I think a comment would be useful explaining why a `listdir` operations requires `map(set`.

In fact, since you're discarding the set resulting from the first list returned by `storage.listdir` why not just use something like:
```python
# Django's storage.listdir always returns a tuple of two lists, the latter of which
# contains the file names we need.
geonode_layers = set(storage.listdir("layers")[1])
```

Note the same applies below.

GeoTob (commented):
There are still a couple of changes that should be made before merging this code. Apologies @mtnorthcott for not getting to review this earlier.

geonode/documents/renderers.py
mtnorthcott: `mkstemp` provides more control of when the file gets cleaned up. This approach was the only way I could get the large temporary image generated by `unoconv` to persist long enough to generate thumbnail content. These calls are equivalent so long as the file gets deleted at the end.

GeoTob (approved):

GeoTob:
@mtnorthcott apologies this took a while to review, but looking good to me now.

PR #6711, opened 2020-12-09
[Fixes #6704] UUID will not be used on XML upload

### LINKED ISSUE #6704: UUID will not be used on XML upload
## Expected Behavior
If I upload a shapefile with iso metadata xml the uuid will be used.

## Actual Behavior

the uuid will not be used.

  1. download and inspect a shapefile with iso xml file (eg: https://github.com/gannebamm/geonode-workshop/tree/main/00_getStarted/data)
  2. upload a shapefile with XML data 
  3. compare the UUIDs

## Specifications

  - GeoNode version: master
  - Installation method (manual, GeoNode Docker, SPCGeoNode Docker): manual
  - Platform: ubuntu
  - Additional details: not plattform dependend

gannebamm:
This should enable XML UUIDs to get used even if metadata preservation is not used.

## Checklist

> This is not well tested currently, therefore it is just a draft PR. It will use the UUIDs but I have not checked what other complications could arise. Maybe there should be Tests covering this behaviour.

For all pull requests:


gannebamm:
I was busy with other things. I have tested the UUID usage locally and it is still working fine. I just changed one line, so this is nothing big. If the devs do like the concept of using the UUID from a metadata file on upload I think this can be merged. The test suites are still processing, though.

gannebamm:
I see the CircleCI ones fail often? Are those working properly? @t-book @afabiani

afabiani:
@gannebamm few things:

1. The fix is ok, but it will work only when using `file_upload` upload. Is that intentional? We should probably consider to align the behavior [here](https://github.com/GeoNode/geonode/blob/master/geonode/geoserver/tasks.py#L284) too.
2. Just overriding the `UUID` looks a bit dangerous to me. In general this whole piece of code is not robust at all. E.g.: what happens if I upload a XML containing a wrong UUID maybe from another resource? We will end up having two resources with the same UUID. We should probably perform some sort of sanity check here and throw an error accordingly. What do you think?

gannebamm:
@afabiani I added a raise to the code if the uuid is already in use (see https://github.com/GeoNode/geonode/pull/6711/commits/46b503fb485085ced61b674648e4a9e7c8fa7380). This will prevent the uploading of two datasets with the same uuid, but the displayed error does not show the actual error message but a internal server error. Is my approach wrong?

Furthermore, I do not see when exaclty the tasks.py will be executed. Will this be used in async mode?

afabiani:
@gannebamm I'll review and try this out locally as soon as possible and push few changes accordingly (if needed)

gannebamm:
Hi @afabiani I have tested this further and see that using a docker-composition instead of a paver dev setup will use a different upload mechanism, at least it seems like that. I think that is bound to async uploading and what you meant by changing these lines of code here: https://github.com/GeoNode/geonode/blob/master/geonode/geoserver/tasks.py#L284

On paver setup the stated code will abort the upload and the layer will not get created. The docker-compose method will:
1) not use the uuid (not even if metadata preserve is used)
2) not fail if two layers with same uuid will be uploaded.

Furthermore, I developed XML file parsing for documents, too. This way an xml documents metadata (if iso,dc,...) is read and used for the resourceBase fields (just like layers with xml). My logic over there does work with docker and async, but will not provide meaningful error messages for the user. It just fails with the default error page.

afabiani:
@gannebamm well noted. I'll try to give it a shot today.

Would you like to describe in more details some use case so that I can teste them and create test cases accordingly?

gannebamm:
> 
> 
> @gannebamm well noted. I'll try to give it a shot today.
> 
> Would you like to describe in more details some use case so that I can teste them and create test cases accordingly?

The main idea is to enable our scientists to use their own metadata editor, if they so choose, and be able to ingest foreign datasets with metadata in our system. Hereby a persistent and unique identifier is important to keep track of those datasets and fulfil our regulatory laws to keep and archive science products (like datasets). To ensure there is only one dataset the check should decline the uploading of already present datasets (uuid wise). This shall hint the scientist to take a look for the original dataset and check if his/her is identical or an update (which should get a new uuid).

Here is a simple dataset that shares metadata and uuid:
https://drive.google.com/open?id=1fK4LRiMmOcJCIpdsFKmx1GcQNQTJ4B93

uploading the 'a' dataset should work and the uuid 'e68061d5-88e2-48c8-960a-8b89066d17d6' should be used for the dataset. If later dataset 'b' is uploaded the upload shall fail with the console.error message. Maybe we also could incorporate the dataset title of the duplicate uuid resourceBase which is already present in the system.

afabiani:
@gannebamm still have to finish wrinting the tests, but in the meantime you can give it a shot.

By the way, translations for the error messages should be generated; could you take care of those ones?

image

afabiani:
@gannebamm tests added too, let's see if they all pass

gannebamm:
Many thanks @afabiani ! I will test this PR now.

gannebamm:
My tests show it is working with docker and displaying the message properly. But while working on the translations I noticed that the `python manage.py makemessages` command will comment out some of the needed translations. This was the cause of #6984 . Some translations are part of fixtures, eg. https://github.com/GeoNode/geonode/blob/master/geonode/base/fixtures/initial_data.json#L40

As much as I discovered up to now, there is no easy way to get fixtures included in the .po files through `makemessages`. One idea is to put those somewhere in the code to get them picked up by `makemasseges` - which I do not like at all. Any idea how to get this more resilient in the future @t-book @afabiani? It will break every time someone will trigger the `makemessage` command, which is handy and need.

afabiani:
Hi @gannebamm currently nothing comes into my mind, except maybe try out one of the options described [here](https://stackoverflow.com/questions/32749854/django-makemessages-decides-to-comment-already-existing-translations) and update the docs accordingly.

Can you quickly try those options?

gannebamm:
@afabiani those wont work:
`-i` will ignore a module. We want to keep previously translated msgid which are not marked as gettext items anymore.
`--no-obsolete` will delete our translations instead of marking them as needed. See https://stackoverflow.com/questions/49339298/django-admin-makemessages-no-obsolete-doesnt-seem-to-be-working and documentation.

`python manage.py makemessages --help` does not show anything which would be able to make our translations working.

The only thing I can think of is to create a python file with the relevant fixture string and load gettext there. This will trigger makemessages and therefore keep them in the translation.

gannebamm:
re: translations

https://gist.github.com/gannebamm/10cb356b93f97b47e869d394dd883e29

placing this file somewhere in the GeoNode codebase will create the iso texts from the fixtures as msgids. I think this is the only way to keep those active. Therefore I would vote to get this PR merged and in a second PR revise the current translation files.

My Idea: I would like to change the makemessage usage to not write out locations, since those will change due to code changes (linenumbers) and this will create loads of noise in the commit messages. This renders it hard to spot new translations and errors (like the #6984 regression). Instead, I would propose to use the above shown custom_message.py file, which will activate the already translated strings. In a second step, I would strip the .po files from the locations ( lines like this `#: geonode/base/__init__.py:29`) and all obsolete old strings (like worldmap stuff). By this we will have a clean and future proof makemessage output.

gannebamm:
See https://github.com/GeoNode/geonode/issues/7015

PR #6839, opened 2021-01-19
[Issue #6843] Remove six library

burner761:
remove all uses of the six library in the project.


afabiani (approved):

afabiani:
@burner761 can you please take a look at the new LGTM issues introduced with this PR?

geonode/monitoring/collector.py
```diff
@@ -452,7 +451,7 @@ def _key(v):
                                   'label': label,
                                   'samples_count': samples,
                                   'value_raw': value or 0,
-                                  'value_num': value if isinstance(value, integer_types + (float, Decimal,)) else None})
+                                  'value_num': value if isinstance(value, int + (float, Decimal,)) else None})
```
mtnorthcott, line 454: `integer_types` is a collection/sequence so `integer_types + (float, Decimal)` extends the sequence. `int` is not a sequence and does not support the `+` operation on the `tuple` type. Please fix this to simply contain `int` within the `tuple`.

mtnorthcott (changes requested):
`six` can still be removed from the following:
* geonode/social/tests.py
* geonode/geoapps/api/serializers.py
* geonode/geoserver/tasks.py
* requirements.txt

afabiani (changes requested):
There are some conflicts, moreover see if it is possible to solve the new LGTM issues.

afabiani (approved):

Issue #6843, opened 2021-01-19
Improve code quality


mtnorthcott:
*Reserved for Catalyst Academy 2021*

We would like to analyse and find potential improvements in code quality throughout the GeoNode codebase. The following improvements have been proposed and should be addressed by this epic:

* Replace body of Layer.display_type with a conditional
* Move constants to top of file
* Fetch dictionary keys with dict.get(...) where appropriate to provide default value
* Remove redundant method overrides
* Prefer sets for in-line lookups
* Refactor long conditional statements
* Prefer generators as input to compatible functions
* Remove redundant enumerate calls
* Remove duplicate function calls
* Remove or replace dependencies
* Replace instances of print with equivalent logger statements
* Add docstrings to functions where it is missing
* Investigate @cached_property and where it could be used
* Remove python_resize_image library if we can do ops in PIL
* Simplify conditional logic
* Replace magic numbers and calculations where possible
* Add helper function to perform encode-decode operations
* Catch specific exceptions
* Remove Python 2/3 cross-compatible code
* Implement GET and POST handler functions
* Close open file descriptors
* Replace existing string formatting with f-strings

Issue #6809, opened 2021-01-11
Refactor long conditional statements


mtnorthcott:
*Reserved for Catalyst Academy 2021*

### Description
There is at least one occurrence where a lengthy conditional could be refactored to be much shorter.

### Example
*geonode/layers/forms.py:124*
```python
elif ext.lower() == '.dbf':
   dbf_file = filename
elif ext.lower() == '.shx':
   shx_file = filename
elif ext.lower() == '.prj':
   prj_file = filename
elif ext.lower() == '.xml':
   xml_file = filename
elif ext.lower() == '.sld':
   sld_file = filename
```
Can be refactored to dictionary
```python
files = {}
...
for filename in filenames:
    ...
    files[ext] = filename
```
and is similar to the existing `cleaned` dictionary in the same method.

### Tasks
* [ ] Refactor the expression in *geonode/layers/forms.py:124*
* [ ] Find similar occurrences of lengthy conditional statements
* [ ] Refactor any conditionals found
* [ ] Verify behaviour is functionally identical

gannebamm:
Hi,
You probably could make those lines https://github.com/GeoNode/geonode/blob/master/geonode/layers/forms.py#L124-L133 better to read, faster, etc. pp.

But AFAIK the proposed replacement will not work properly. Eg a `my_file.dbf` will be set as sld_file which is wrong.

mtnorthcott:
@gannebamm Hi, thanks for the heads up. I overlooked the variable name there and have amended the issue to be as you suggested.

PR #6846, opened 2021-01-20
[Issue #6843] Simplify long conditionals

burner761:
simplifyed some long conditionals.


mtnorthcott (changes requested):
Please check the associated issue #6809 as this has specific information on what this PR should involve. Instances of `if variable is None:` are not considered long conditionals and should be preferred as `if not variable:` is not specific. [Explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/).

afabiani (approved):

PR #6998, opened 2021-02-26
Assorted minor performance issues

afabiani:
This PR includes a small series of minor improvements and fixes as described here below:

 - Make `is_monochromatic_image` able to fetch local thumbs instead of querying URL anytime: this is a performance improvement; if the method recognizes that the thumbnail url is pointing to a local file, it will fetch it from the local storage instead of getting the stream from the URL. This also resolves a potential issue with NATted instances.
- Create "tempdirs" under "settings.STATIC_ROOT" in order to make sure containers share the same folder: when running docker containers in `ASYNC` mode, if the `/tmp` folder is not shared between `django` and `celery` containers, the latter won't be able to retrieve the uploaded auxiliary files. By creating the `/tmp` folder under the `STATIC_FOLDER` fixes the problem, since this is shared by default.
- Make sure the SRID is always set also in the case it could not be found into the SRID string: this is a minor potential issue whenever the regex on the `base.models` is not able to fetch the SRID correctly.

### DIFF
--- a/geonode/base/models.py
+++ b/geonode/base/models.py
@@ -1056,7 +1056,7 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin, ItemBase):
         if self.bbox_polygon:
             bbox = self.bbox_polygon
             match = re.match(r'^(EPSG:)?(?P<srid>\d{4,6})$', self.srid)
-            srid = int(match.group('srid'))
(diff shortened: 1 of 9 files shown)


t-book (approved):

afabiani:
Going to merge... usual test_upload timeout (still to be solved)

PR #7270, opened 2021-04-11
[Fixes  #7269] Reduce information returned to users to only what is strictly required and accessible

### LINKED ISSUE #7269: Users information leak
## Expected Behavior

Only administrators must be able to retrieve a full list of users including their personal information.

## Actual Behavior

GeoNode's /owners API endpoint returns the full list of users, including their personal information (!!!!).

  1. Go to layers page
  2. On inspect, In the network tab, Check response for owners request. The list returned includes all attributes on a user.
  3. Also check on the user's profile, more information is provided.

## Specifications

  - GeoNode version: master
  - Platform:
  - Additional details:

marthamareal:
Protects user's information by returning limited attributes on their profile to anonymous and non-admin users.

## Checklist

For all pull requests:

### DIFF
--- a/geonode/api/api.py
+++ b/geonode/api/api.py
@@ -538,6 +538,26 @@ class ProfileResource(TypeFilteredResource):
                     bundle.obj).pk,
                 'object_id': bundle.obj.pk})
 
+    def dehydrate(self, bundle):
+        """
+        Protects user's personal information from non staff
+        """
+        is_owner = bundle.request.user == bundle.obj
+        is_admin = bundle.request.user.is_staff or bundle.request.user.is_superuser
+        if not (is_owner or is_admin):
+            bundle.data = dict(
+                id=bundle.data.get('id', ''),
+                username=bundle.data.get('username', ''),
+                first_name=bundle.data.get('first_name', ''),
+                last_name=bundle.data.get('last_name', ''),
+                avatar_100=bundle.data.get('avatar_100', ''),
+                profile_detail_url=bundle.data.get('profile_detail_url', ''),
+                documents_count=bundle.data.get('documents_count', 0),
+                maps_count=bundle.data.get('maps_count', 0),
+                layers_count=bundle.data.get('layers_count', 0),
+                )
+        return bundle
+
     def prepend_urls(self):
         if settings.HAYSTACK_SEARCH:
             return [
@@ -583,6 +603,16 @@ class OwnersResource(TypeFilteredResource):
         if bundle.request.user.is_superuser:
             email = bundle.obj.email
         return email
+
+    def dehydrate(self, bundle):
+        """
+        Protects user's personal information from non staff
+        """
+        is_owner = bundle.request.user == bundle.obj
+        is_admin = bundle.request.user.is_staff or bundle.request.user.is_superuser
+        if not (is_owner or is_admin):
+            bundle.data = dict(id=bundle.obj.id, username=bundle.obj)
+        return bundle
 
     def serialize(self, request, data, format, options=None):
         if options is None:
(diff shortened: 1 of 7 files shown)


afabiani (changes requested):
When connecting with a standard user, trying to fetch the list of people, I see this

image

marthamareal:
> When connecting with a standard user, trying to fetch the list of people, I see this
> 
> image

@afabiani  This is because only admins can access all information on other users, For logged-in and anonymous users, they only have access to (usernames) for other users.

giohappy:
@marthamareal isn't the username enough in that template? Why is "afabiani" showing up there and not the other two usernames?

marthamareal:
> @marthamareal isn't the username enough in that template? Why is "afabiani" showing up there and not the other two usernames?

Because the template is checking for empty first_name [here](https://github.com/GeoNode/geonode/blob/master/geonode/people/templates/people/_profile_list_item.html#L12).

giohappy:
@marthamareal I think it's fine to also return first name and last name. We can also return the avatar if the template requires it, but we could remove it from the API response, can't we?

marthamareal:
The template uses the data that is already modified from the api.

afabiani:
@marthamareal we can probably remove the whole card if the username is not present?

@giohappy I would probably add an option to the users in order to allow them to make the profile public or not. This is because IMHO most of the use cases need the people list to be visibile publicly (or at least most of it).

t-book:
>  I would probably add an option to the users in order to allow them to make the profile public or not.

👍

giohappy:
> @giohappy I would probably add an option to the users in order to allow them to make the profile public or not. This is because IMHO most of the use cases need the people list to be visibile publicly (or at least most of it).

In general clients ask for a "private by default" policy regarding resources and users information. I'm fine with adding an option to make one's profile public, but I would do it in a separate issue.

We have to take a decision on this. What we are implementing here is:

1 - return restricted users info to non-admin users
2 - only return users which have published at least a resource which is visible to the current user

If we add the option to let the user to make his profile public we should override both these restrictions, i.e. show the public profile and independently from having published a resource or not. Is this what you expect @afabiani and @t-book?

giohappy:
Ok, we can move this topic to a next phase and keep returning all the users. However I have a question: what's the difference between `owners/` and `profiles/` endpoints, except some different field in the response?

afabiani:
The owners should return the facets for the counters on the resources detail (if I recall this correctly)

giohappy:
> The owners should return the facets for the counters on the resources detail (if I recall this correctly)

profiles/ details return the facets.
owners/ detaild seems broken, it always returns null for count.

giohappy:
@marthamareal this topic requires some more discussion. Moreover this API will be replaced by the newer API v2, so I wouldn't spend more time on this.

Please remove the counts==0 check, so we revert to returning all the users (with reduced information)

geonode/api/api.py
```diff
@@ -126,7 +127,14 @@ def to_json(self, data, options=None):
         counts = self.get_resources_counts(options)
         if 'objects' in data:
             for item in data['objects']:
-                item['count'] = counts.get(item['id'], 0)
+                count = counts.get(item['id'], 0)
+                if count == 0 and 'owners' in item.get('resource_uri', ''):
```
marthamareal, line 131: @giohappy I had provided this check to consider the count of resources on only owners resource, what do you think?

giohappy: To be honest I don't like that hardcoded "owners" check, especially inside a generic method. Let's revert to eh previous behaviour. This PR will only reduce the returned fields but we'll keep returning users as before. We will review this with new API v2

marthamareal:
I have made only (`avatar`, `first_name`, `last_name`, and `username`) public fields on getting owners and this also applies when getting profiles.

giohappy:
what about the organization @marthamareal? Isn't it needed by the template?

afabiani (approved):

giohappy (approved):

PR #7299, opened 2021-04-14
[Fixes #7292] Thumbnail failure when data bbox extends beyond the are a of use of EPSG:3857

### LINKED ISSUE #7292: Thumbnail failure when data bbox extends beyond the area of use of EPSG:3857
The new thumbnail code employes the 4326 to 3857 transformation (through pyproj's Trnasformer) in several places. The trasnform fails silently since the `errcheck` parameter is False (by default). The returned wrong values are then used downstream leading to failures.

Thumbnails assume EPSG:3857 backgrounds and overlays to generate the output image. This means that nothing beyond the EPSG:3857 area of use can be rendered (the same as with map view's themselves). This menas that we can safely clamp any bbox calculation to the EPSG:3857 area of use:

```
>>> from pyproj.crs import CRS
>>> crs_3857 = CRS.from_user_input(3857)
>>> crs_3857.area_of_use()
AreaOfUse(west=-180.0, south=-85.06, east=180.0, north=85.06, name='World between 85.06°S and **85.06°N.')
```

Moreiver the transform should use `errcheck` and manage the exception (with some logging).

bieganowski:
Refers: ISSUE #7292

Implemented solution
- for generating a default thumbnail (thumbnail generated in `EPSG:3857`; BBOX retrieved from `layer.bbox` of an instance), cropping the BBOX to the `EPSG:3857`'s area of use was introduced to ensure correct CRS transformations
- for the default thumbnail's CRS (`EPSG:3857`), the flow for BBOXes provided in `EPSG:3857` was preserved
- for the default thumbnail's CRS (`EPSG:3857`), for BBOXes provided in CRSs different from `EPSG:3857` an additional check was introduced, forcing the crop of the overflowing coordinates to `EPSG:3857`'s area of use
- for forced thumbnail's CRS (different from `EPSG:3857`; by default used in layers from external services), and additional check was introduced, turning off thumbnail's ratio preservation, in case of receiving a BBOX in CRS different from `EPSG:3857`
## Checklist

For all pull requests:

### DIFF
--- a/geonode/thumbs/utils.py
+++ b/geonode/thumbs/utils.py
@@ -22,7 +22,7 @@ import time
 import base64
 import logging
 
-from pyproj import Transformer
+from pyproj import Transformer, CRS
 from typing import List, Tuple, Callable, Union
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


geonode/thumbs/utils.py
```diff
@@ -261,3 +261,63 @@ def fetch_wms(url: str, max_retries: int = 3, retry_delay: int = 1):
             break
 
     return image
+
+
+def epsg_3857_area_of_use():
```
giohappy, line 266: I would set the returned values as constants (i.e. populate the constants from this method at import time), without calculating them each time.

giohappy (commented):
@bieganowski in case the aspect preservation can't be performed, what if we render the final image with the thumbnail aspect ratio and we just center the layer image on it without ditorting it? The final thumbnail will have either two horizontal or two vertical blank stripes, but it will be better then a distorted image.

PR #7317, opened 2021-04-15
Add initial support for harvesting

ricardogsilva:
## THIS IS A WORK_IN_PROGRESS - NOT READY YET!

- [ ] - code
- [ ]  - tests
- [ ] - documentation
- [ ] - translations

This PR is the first in an ongoing work that aims to add harvesting capabilities to GeoNode. It adds a new `geonode.harvesting` app which in time shall grow to include functionalities for enabling harvesting of other services and bringing their metadata records (and data too) onto GeoNode.

A brief overview of the intended workflow is:

1. Privileged user creates a new `Harvester`. Among its properties, are the URL for the remote service to be harvested, the type of harvester worker and a harvesting periodicity.
2. In due time, celery beat schedules a new harvesting session for execution. This entails instantiating the harvesting worker and persisting a `HarvestingSession` record in the GeoNode DB
3. During the harvesting session, the harvesting worker probes the remote service and discovers its resources. From each resource the worker builds a `ResourceDescription`, which holds relevant metadata about it.
4. The harvesting worker sends the generated resource descriptions to the GeoNode `resource manager` (something which does not exist yet, but is also being worked on (@afabiani)
5. The resource manager takes care of creating or updating GeoNode resources from the information that has been harvested

## Implementation details

The proposed `Harvester` model has the following attributes:

- `remote_url` - base URL of the remote service to be harvested
- `enabled` - whether the harvester shall be scheduled periodically or not
- `update_frequency` - how often (in minutes) should new harvesting sessions be scheduled
- `default_owner` - which GeoNode user shall become the owner of the harvested resources
- `default_access_permissions` - What should be the default access permissions of the harvested resources
- `harvester_type` - the type of harvester that is to be used. This can grow over time to support different remote services. The current PR adds a sketch for supporting harvesting from remote GeoNodes. It shall be continued in subsequent PRs
- `harvester_type_specific_configuration` - any specific configuration that may be needed by concrete harvester types. This is stored as a `JSONField` in order to be flexible.

When a new `Harvester` is created, we also create a new a `django_celery_beat.models.PeriodicTask` to serve it. This periodic task has a scheduling interval and it also is directed to execute our `geonode.harvesting.tasks.harvesting_session_dispatcher` task. The celery beat scheduler takes care of scheduling this task every `x` minutes. The `harvesting_session_dispatcher` task spawns a harvesting worker of the relevant type, as configured in the harvester (i.e a `geonode.harvesting.harvesters.geonode.GeonodeHarvester`) and calls its `perform_metadata_harvesting()` method.

Each concrete harvesting worker executes its own logic in the `perform_metadata_harvesting()` method. The proposed implementation adds a `GeonodeHarvester`, which is able to harvest from remote GeoNode services. Harvesting is done by using the remote GeoNode's CSW API. It is broken down into batches of ten resources, according with the size of each CSW `GetRecords` response. Each batch is processed in parallel by spawning additional celery tasks. In the end there is a synchronization step - this uses celery's `chord()` primitive.

As mentioned above, this PR does not actually implement creating new GeoNode resources. The provided `GeonodeHarvester` probes a remote GeoNode and extracts information about existing records into a list of `ResourceDesciption` objects. Further work to come shall take these and create the actual resources on the local GeoNode instance

Also included in this PR is an initial implementation of new API endpoints for harvesting.


t-book:
@ricardogsilva you're writing the harvester will ingest data as well. would it be conceivable to consume MQTT data as well? – or is it's planned to keep it generic so that users can add a custom harvest service for their needs?

giohappy:
@t-book the harvester is meant to support pluggable clients. The first use case will be the [GeonodeHarvester](https://github.com/GeoNode/geonode/pull/7317/files#diff-8a21614f9433d70c957381f608eec8a9f406d48cd76b58e9cf4d21e54a0bf20fR40) implementation, but it's deisgned to support any custom harvester that can be activeted through settings.
Each harvester instance can also have distinct scheduling settings.

t-book:
wow, sounds awesome! thanks for the insights @giohappy

afabiani:
Going to close this one as it will be part of [GNIP 89: Architecture Design - Resource and Storage Manager Modules](#7670)

PR #7364, opened 2021-04-21
[Fixes  #7355] Introduce the adoption of groups to define users permissions, and define add_resource permission for users

### LINKED ISSUE #7355: Introduce the adoption of groups to define users permissions, and define add_resource permission for users
We want to improve the way permissions are managed in GeoNode, to make room for more advanced combinations of roles and permissions. The idea is to increase the usage of permissions to granularly drive the functionalities both in the backend and the client, sintead of simply relying on the "AnonymousUser" and the admin concepts.

We start by introducing the concpet of a "contributors" role (Django Group) which is assigned the base_addresourcebase. Every user will be assigned to the group by default, so that the current assumption that any registerd user can create resources is maintained.

The first usage of this permissions will be:

 - the client uses it to decide if the "Create ..." and "Save as..." should be visible for the user
 - replace the READ_ONLY checks in the backend to decide if resources can be created (the only case where they can't is when the read only mode is active)

More use cases to come, and more permissions will be added to the group and security checks will be done on top of these permissions.

## Expected Behavior
-  A `contributors` Django group is created and the `base_addresourcebase` permission is assigned to it. The creation of this group could follow the same approach that we're using to create (and assign) a user to the `anonymous` group, i.e. it's created the very first time we need it.
- Every user will be assigned, by default, to the `contributors` group.(except anonymous and admins)
- A new `People.perms` property will return all the set (unique list) of permissions from the user's groups, i.e. permissions will be the combination of all the group's (non-repeated) permissions the user is a member of. 

The returned permissions are a mapping from Django permissions to a new set of GeoNode constants that will be declared inside the `security` module.
The first constant will be:

`base_addresourcebase -> add_resource`

The same will be done along the way to convert all the GeoNode code to use constanst instead of the original Django / Guardian permission strings.

## Actual Behavior
- There is no contributor group and `base_addresorcebase`

  1. New addition

## Specifications

  - GeoNode version: master
  - Platform:
  - Additional details:

marthamareal:
References: #7355

- Creates a new user group `contributors` with permission `base_addresourcebase`.
- Updates existing users to belong to a new contributors group.
- Includes user's group perms to `get_user_perms` function

## Checklist

For all pull requests:


giohappy:
@marthamareal what's the purpose of adding the base_addresourcebase permission to the Profile model?

marthamareal:
> @marthamareal what's the purpose of adding the base_addresourcebase permission to the Profile model?

This was before we introduced using contributors group, let me change and create the permission on the group

geonode/security/constants.py
```diff
@@ -0,0 +1,9 @@
+# Permissions constants
+
+GENERAL_PERMISSIONS = {
```
giohappy, line 3: @maabdelghaffar I would rename this to `PERMISSIONS`, simply

geonode/security/models.py
```diff
@@ -396,6 +399,10 @@ def get_user_perms(self, user):
         """
         Returns a list of permissions a user has on a given resource
         """
+        # To avoid circular import
+        from geonode.base.models import Configuration
+
+        config = Configuration.load()
         ctype = ContentType.objects.get_for_model(self)
         PERMISSIONS_TO_FETCH = VIEW_PERMISSIONS + ADMIN_PERMISSIONS + LAYER_ADMIN_PERMISSIONS
```
giohappy, line 407: @marthamareal I would also move VIEW_PERMISSIONS, ADMIN_PERMISSIONS, LAYER_ADMIN_PERMISSIONS into the permissions module.

marthamareal: Done

giohappy (dismissed):

giohappy (changes requested):
@marthamareal instead of using a command to assign the contributors group to users it would be better to use a migration. This way it runs automatically and only once.

Something like this would be fine:

```
# Assign the contributors group to users according to #7364

from django.db import migrations, models
from django.db.migrations.operations import RunPython

def forwards_func(apps, schema_editor):
    # assign contributors group to users
    pass

def reverse_func(apps, schema_editor):
    # remove contributors group from users
    pass

class Migration(migrations.Migration):

    dependencies = [
        ('people', '0031_merge_20210205_0824'),
    ]

    operations = [
        RunPython(forwards_func, reverse_func)
    ]

```

marthamareal:
> @marthamareal instead of using a command to assign the contributors group to users it would be better to use a migration. This way it runs automatically and only once.
> 
> Something like this would be fine:
> 
> ```
> # Assign the contributors group to users according to #7364
> 
> from django.db import migrations, models
> from django.db.migrations.operations import RunPython
> 
> def forwards_func(apps, schema_editor):
>     # assign contributors group to users
>     pass
> 
> def reverse_func(apps, schema_editor):
>     # remove contributors group from users
>     pass
> 
> class Migration(migrations.Migration):
> 
>     dependencies = [
>         ('people', '0031_merge_20210205_0824'),
>     ]
> 
>     operations = [
>         RunPython(forwards_func, reverse_func)
>     ]
> ```

updated this

afabiani:
@marthamareal LGTM complains about a new unused import introduced with this PR, please, take a look. Thanks.

giohappy (approved):

giohappy (changes requested):
@marthamareal let's use lowercase for GeoNode permission names, so please change "ADD_RESOURCE" to "add_resource".

PR #7720, opened 2021-06-28
[Fixes #7717] Temporary files created under static root are never rem…

### LINKED ISSUE #7717: Temporary files created under static root are never removed
There are many situations where temp files are created under static root (prepare zip for download, unzip folders, etc.).
None of these methods appear to correctly remove these temp files, which makes the static root grow indefintely.

Ex.
 - resource download: https://github.com/GeoNode/geonode/blob/347a2053d179721b3dec53410d67757a563114d1/geonode/proxy/views.py#L283
 - unzip files: https://github.com/GeoNode/geonode/blob/347a2053d179721b3dec53410d67757a563114d1/geonode/utils.py#L119

We must check all the places where this is done and correctly handle the revoal of temp files, including exceptions.

afabiani:
…oved

## Checklist

For all pull requests:

### DIFF
--- a/geonode/geoserver/tests/test_server.py
+++ b/geonode/geoserver/tests/test_server.py
@@ -646,7 +646,7 @@ class LayerTests(GeoNodeBaseTestSupport):
             self.assertEqual(line_sld_name, 'line 3')
         finally:
             if d is not None:
-                shutil.rmtree(d)
+                shutil.rmtree(d, ignore_errors=True)
 
     @on_ogc_backend(geoserver.BACKEND_PACKAGE)
     def test_layer_acls(self):
--- a/geonode/geoserver/upload.py
+++ b/geonode/geoserver/upload.py
@@ -18,6 +18,7 @@
 #
 #########################################################################
 import uuid
+import shutil
 import logging
 import geoserver
 
@@ -116,7 +117,7 @@ def geoserver_upload(
     logger.debug('>>> Step 4. Starting upload of [%s] to GeoServer...', name)
 
     # Get the helper files if they exist
-    files = get_files(base_file)
+    files, _tmpdir = get_files(base_file)
     data = files
     if 'shp' not in files:
         data = base_file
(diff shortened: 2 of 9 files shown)


mattiagiupponi (changes requested):
I'll consider adding a check to verify that the temp folder exists. Because (for any reason) if the folder has been deleted during the operations, on the `shutil.rmtree(_tmpdir)` this will raise an error. Maybe something like this could be enough to prevent it:
```python
if os.path.exists(tempdir) and tempdiris not None
    shutil.rmtree(tempdir)
```

afabiani:
> I'll consider adding a check to verify that the temp folder exists. Because (for any reason) if the folder has been deleted during the operations, on the `shutil.rmtree(_tmpdir)` this will raise an error. Maybe something like this could be enough to prevent it:
> 
> ```python
> if os.path.exists(tempdir) and tempdiris not None
>     shutil.rmtree(tempdir)
> ```

I would prefer to avoid existence checks. Instead I would use the following options of the `shutil.rmtree` method eventually

`def rmtree(path, ignore_errors=False, onerror=None)`

mattiagiupponi:
> I would prefer to avoid existence checks. Instead, I would use the following options of the `shutil.rmtree` method eventually
>
> `def rmtree(path, ignore_errors=False, onerror=None)`

Maybe using `ignore_errors=True` could not be the best approach since it just silently ignore errors.
from the doc:

>  If ignore_errors is true, errors resulting from failed removals will be ignored; if false or omitted, such errors are handled by calling a handler specified by onerror or, if that is omitted, they raise an exception.

For example, if the error raised is a `NotFound` is fine that we can ignore it, but if it raises an error regarding the User / App permissions on the filesystem/storage system, is correct that an `Exception` is raised, but with `ignore_errors=True`, you are just ignoring a potential issue.
I think that just checking the existence of the directory could the safest approach

afabiani:
> > I would prefer to avoid existence checks. Instead, I would use the following options of the `shutil.rmtree` method eventually
> > `def rmtree(path, ignore_errors=False, onerror=None)`
> 
> Maybe using `ignore_errors=True` could not be the best approach since it just silently ignore errors.
> from the doc:
> 
> > If ignore_errors is true, errors resulting from failed removals will be ignored; if false or omitted, such errors are handled by calling a handler specified by onerror or, if that is omitted, they raise an exception.
> 
> For example, if the error raised is a `NotFound` is fine that we can ignore it, but if it raises an error regarding the User / App permissions on the filesystem/storage system, is correct that an `Exception` is raised, but with `ignore_errors=True`, you are just ignoring a potential issue.
> I think that just checking the existence of the directory could the safest approach

The `onerror` callback is supposed to catch this... however in this specific case we don't care about catching the exception... we are just trying to delete a temp folder.
If we need a hard stop we will need to manage the specific case.
The existence check is FS specific and we are migrating to the storage manager.

PR #7774, opened 2021-07-09
[Fixes #7750] GeoNode shows incorrect keywords

### LINKED ISSUE #7750: GeoNode shows incorrect keywords
## Expected Behavior
With hierarchical keywords we want to correctly see the assigned keywords to a resource also if we assign to the resource just a leaf keyword.
Example of 3 levels keywords:
image 
If we assign just the keyword KC to a resource we want to see it correctly assigned in the **keywords** section of the layers page.

## Actual Behavior
When we assign just the keyword KC and depending of the creation order of the keywords we cannot see the assigned keyword to the resource:
image

  1. Create a 3 levels keywords (eg KA, KB, KC)  
image 
  and assign just the keyword KC to a resource (R1) 
  2. Open the "Explore Layers" page and take a look on the **keywords** section.
  3. Depending on the order of creation of the keywords and the nodes hierarchy you cannot see the keyword

## Specifications

  - GeoNode version: 3.2.0
  - Installation method (manual, GeoNode Docker, SPCGeoNode Docker): manual, GeoNode Docker
  - Platform: Ubuntu 18.04
  - Additional details:

asgharsir:
References: #7750

## Checklist

For all pull requests:

- [X] Confirm you have read the [contribution guidelines](https://github.com/GeoNode/geonode/blob/master/CONTRIBUTING.md) 
- [X] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in the documentation)
- [X] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

- [X] There is a ticket in https://github.com/GeoNode/geonode/issues describing the issue/improvement/feature (a notable exemption is, changes not visible to end-users)
- [X] Commit message must be in the form "[Fixes #<issue_number>] Title of the Issue"
- [X] New unit tests have been added covering the changes, unless there is an explanation on why the tests are not necessary/implemented
- [X] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [X] This PR passes the QA checks: flake8 geonode

Seeing the issues in the previous testing I approached for using far lighter `APITestCase` and avoided the `GeoNodeBaseTestSupport` mixin and created `factories` in the base app, which I have not only used for this test but it also worked for the earlier test case which failed for the 'exotic encoding text'. In future we can use the factories to avoid, wherever intended, the `GeoNodeBaseTestSupport` which is far heavier, and `create_layer_data()`, which I see is a 'hard-coded' data, so also obscure the data for readability of the tests/code.

Also added in the base.utils the `make_public` function as cleaning the 'dirty_state' and making public of a resource, goes hand in hand here, and have many use cases.


giohappy:
@asgharsir I see two issue:

I have a layer "Castore ..." which has assigned two keywords, "cartografia storica", "a" and "b"

the keywords hierarchy is:

image

(the parent of "cartografia storica" is not assigned to the layer

### Keywords tree: 
  - I expect to obtain the tree for "storico", even if "storico" itself is not assigned to the layer. Of course the keyword won't have any badge
  - Why "a" and "b" are not represented as a tree?

image


### Search page: if I search for "Castore" I expect to obatin only the keywords (tree) for the search result, instead all the existing keywords are returned (not as a tree).

image

giohappy (changes requested):
@asgharsir I see two issue.

I have a layer "Castore ..." which has assigned two keywords, "cartografia storica", "a" and "b"

the keywords hierarchy is:

image

(the parent of "cartografia storica" is not assigned to the layer

### 1. Keywords tree: 
  - I expect to obtain the tree for "storico", even if "storico" itself is not assigned to the layer. Of course the keyword won't have any badge
  - Why "a" and "b" are not represented as a tree?

image


### 2. Search page: if I search for "Castore" I expect to obatin only the keywords (tree) for the search result, instead all the existing keywords are returned (not as a tree).

image

mattiagiupponi (changes requested):
@asgharsir few notes:
1) At this moment the Factories are not required. We already have classes to support the test cases as you mentioned before. The `GeoNodeBaseTestSupport` which is a bit heavy, but contains all the stuff required to run the tests, or the following methods under `geonode/base/populate_test_data.py`:
- `create_single_layer()`
- `create_single_doc()`
- `create_single_map()`
- `create_models()`

which are far lighter weight than the `GeoNodeBaseTestSupport` class. If you need to add the keywords for example by using the `create_single_layer()`, extend the method instead of creating a factory. the `make_public` is not required since the above methods will take care of it.
Can you please convert your tests by using the above methods?

2) For the users you can simply use:
`get_user_model()` to get the models and then create the users or call directly the user with `get_user_model().objects.get(username='admin')` if you are using `GeoNodeBaseTestSupport`
3) looks like that in the method `dump_bulk_tree` the concept of `depth` is no longer present. Please note that we can have a situation like this one:

image

And the tree should be represented accordingly

4) I'll take in consideration to add a tests with a simple assert of the expected payload and the one returned by the API, something like (pseudo-code):

```python
expected = {"keyword": "dict"}
actual = self.client.get(reverse(self.url))
self.assertDictEqual(expected, actual)

**Comment by asgharsir** [comment]
In your example @giohappy  did you mean, A is parent of B, Cartografia Storica is parent of A and Storica is the root? And in the Keywords rendering, even if Storica is not assigned to any layer being searched, they should appear like:
Storica
	Cartografia Storica
		A
			B

Secondly, I want to confirm that here search means searching a resource, not searching resources by keyword?

So, how are we passing that resource? Is that some tastypie url that I need to know? Because the url in the initial description is only expecting a type input. 

I think with the initial description, I failed to assume payload endpoint/call and the payload/arguments all.

**Comment by asgharsir** [comment]
Tree things is already achieved but I need feedback on which endpoint is calling to the subject method. Only one view is using that method dump_bulk_tree. And the method and the view both are only accepting resource type not the resource itself as the input. So, I must be missing some knowledge here.
image

**Comment by giohappy** [comment]
>In your example @giohappy did you mean, A is parent of B, Cartografia Storica is parent of A and Storica is the root? And in the Keywords rendering, even if Storica is not assigned to any layer being searched

If you look to the first image in my previous comment you will see the following herarchy:

- `a` (ssigned to the resource)
  - `b` (assigned to the resource)
- `storico` (NOT assigned)
  - `cartografia storica` (assgiend)

I expect to see the same tree inside the keywords filter, including `storico` which is not assigned to the resource (so the bade count will not include the resource).

The search page is the one for this URL: `https://stable.demo.geonode.org/search/`, which is triggered either from the top bar search box or the text box at the top of the left sidebar. I admit I'm surprised to see that no query parameter is sento the the `h_keyword_api` endpoint from within the search box. Apparently the only filtering applies to this endpoint is by resource type (layer, document, etc.), and it's only used from the related list pages.
Forget about this for the moment and just focus on the fix of the tree structure returned by the endpoint.

**Comment by asgharsir** [comment]
@giohappy This was done on Friday already:
Keywords Tree Working

I am going for the tests.

**Comment by asgharsir** [comment]
As to factories @mattiagiupponi, Firstly populate_test_data methods are not lightweight. Please use create_single_layer etc in such particular tests in places of LayerFactory you will see the difference. The difference is because these methods are making a complete resource and apparently their purpose is to "populate" these resource models to test these resources themselves. While in our tests, we may use a layer not for testing a layer, like here we are testing returned keywords and not the layer.

Secondly these methods as I said are not meant for general testing. For example, they do make_public() things built-in. Why I did not encapsulate make_public() things in the factories. Because it is against Interface Segregation Principle to do so. It simply means it will not leave my factories useful for general testing but for a specific purpose only. What if you are applying create_single_layer() method in a test where you need unclean or raw layers. Or how many lines of codes a new contributor has to read to know this fact. Or how many codes a test reader would need to read to know all such facts.

And mainly populate_test_data methods are not perceivable helpers for testing. A new contributor would not know whether they exist or not. A new contributor would know the project is using factory-boy and the first thing he would look for such testing is what factories are already existing.

I see that factories are already used in our repo. Following may be helpful to see why factories would be a right step to continue with. 

https://realpython.com/django-pytest-fixtures/#using-a-factory
https://stackoverflow.com/questions/11543117/what-are-the-best-practices-for-testing-different-layers-in-django

Test suggestion is nice. I would add some of this pattern too. They can help less-code and readability if done wisely. Thanks Mattia.

File: geonode/base/utils.py [hunk 1]
```diff
@@ -218,3 +218,11 @@ def _restore_owner_permissions(self):
                     assign_perm(perm, self.resource.owner, self.resource)
                 elif perm not in {'change_resourcebase_permissions', 'publish_resourcebase'}:
                     assign_perm(perm, self.resource.owner, self.resource)
+
+
+def add_keywords_to_resource(resource, keywords):
+    for keyword in keywords:
+        resource.keywords.add(keyword)
+
+    resource.save()
```
afabiani, line 227: This method seems to be used only for testing purposes.
Please do not mix utility methods for testing with the code on the core.
Remove it from `base` and place it on some other specific utility module for testing.

Other than this, `resource.save()` is not needed here.

geonode/base/models.py
```diff
@@ -354,54 +354,85 @@ def dump_bulk_tree(cls, user, parent=None, keep_ids=True, type=None):
...
+            tags = TaggedContentItem.objects.filter(
+                    content_object__in=resources,
+                    tag=hkw
+                    )
 
-                if 'id' in fields:
-                    # this happens immediately after a load_bulk
-                    del fields['id']
+            if tags.exists():
+                tags_count = tags.count()
```
mattiagiupponi, line 384: You can use directly `tags.count()` since you are using a `filter` for the queryset. This means returning an empty queryset `<MultilingualPolymorphicQuerySet []>` so the count will always be 0.

geonode/base/models.py
```diff
@@ -337,76 +338,101 @@ def get_queryset(self):
...
 
     @classmethod
-    def dump_bulk_tree(cls, user, parent=None, keep_ids=True, type=None):
-        """Dumps a tree branch to a python data structure."""
+    def resource_keywords_tree(cls, user, parent=None, resource_type=None, resource_name=None):
+        """ Returns resource keywords tree as a dict object. """
         user = user or get_anonymous_user()
-        ctype_filter = [type, ] if type else ['layer', 'map', 'document']
-        qset = cls._get_serializable_model().get_tree(parent)
+        resource_types = [resource_type] if resource_type else ['layer', 'map', 'document']
```
afabiani, line 347: We need to add here also the `GeoApp` subtypes in the case the `resource_type` was not valorized.

Add the `apps subtypes` other than `['layer', 'map', 'document']` to the default `ctypes`

afabiani (changes requested):

geonode/utils.py
```diff
@@ -2121,3 +2121,10 @@ def verify_image(stream):
     except Exception as e:
         logger.debug(e)
         return False
+
+
+def find_by_attr(lst, val, attr="id"):
+    """ Returns an object if the id matches in any list of objects """
+    for item in lst:
+        if item[attr] == val:
```
mattiagiupponi, line 2129: Did you evaluate using the `.get` method of the dictionary? Because in this case if `attr` does not exist in the` item` dict, it will raise an exception.
Maybe something like can avoid it:

```python
def find_by_attr(lst, val, attr="id"):
    """ Returns an object if the id matches in any list of objects """
    for item in lst:
        if item.get(attr, None):
            return item
```

geonode/base/utils.py
```diff
@@ -220,6 +220,9 @@ def _restore_owner_permissions(self):
                     assign_perm(perm, self.resource.owner, self.resource)
 
 
-def make_public(resource):
-    resource.set_default_permissions()
-    resource.clear_dirty_state()
+def add_keywords_to_resource(resource, keywords):
+    for keyword in keywords:
```
mattiagiupponi, line 224: I'll keep this utility only for tests purposes because, for a normal flow, the `KeywordHandler` has the responsibility to assign the keywords to the resource. 
It converts the keyword to a specific format in order to let the object understand if is normal keywords or thesaurus once. If you are interested, here are the test cases of the keyword handler -> [link](https://github.com/GeoNode/geonode/blob/0c8b6b83afb4271109041fb4d3ca87ed53577eed/geonode/upload/tests/test_utils.py#L69)

geonode/base/utils.py
asgharsir: This can be a utility method for `ResourceBase`, as we are passing in Resource to the method, and the functionality has forseeable usage. But if agreed to not use for that, then we can move it `populate_test_data` itself.

geonode/base/models.py
asgharsir: I just continued it with the existing code input. As you guys suggest.

mattiagiupponi (changes requested):
Other than the other comments, there is a small UI issue.
The `+` and the `-`  for the tree object are not working. If I press the `-` as a user, I'm expecting that the tree is folded, but is not working.
Please take some time to check if is an easy fix (timebox yourself for about 15/20 mins) if the fix requires more time, according to @giohappy please just remove the `+` and the `-`  from the UI

geonode/base/models.py
asgharsir: You are right @mattiagiupponi. I wrongly intuited that on null query it will raise a count() on none. Thanks

geonode/base/utils.py
afabiani: Yes, please remove it.

Keep in mind that in GeoNode modifying the models directly is always highly discouraged and moreover invoking a `save()` on them.

This is something that can be done during the test phase, but not in the core. Why? Because the `'save()` methods in GeoNode are linked to a bunch of signals that interact with the GIS backend. They are not only heavy and time-consuming, but also intrinsically dangerous.

This is why we should not have utilities on the core that invoke a `save()` call.

geonode/base/models.py
afabiani: If possible and not too hard, it would be good to improve this part.

geonode/utils.py
asgharsir: We can put a check on that to return None in that case. Actually it was a last minute change before PR commit when I found the same code is repeated in both tests and model. It was initially only for id only in tests. I will put a check.

geonode/base/utils.py
asgharsir: Our `keywords` are in fact ManyToMany field behind the scene by Taggit. As taggit [document admits it](https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py#59) that if it is not managed by their manager we can do with it all Django goodness. But as I said earlier, if most feel this method to be for test purpose only, I should move it to `populate_test_data` itself.

mattiagiupponi (approved):

PR #7827, opened 2021-07-19
[Fixes #7826] Implement tests for harvesters

meomancer:
<Include a few sentences describing the overall goals for this Pull Request>

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

For all pull requests:

- [x] Confirm you have read the [contribution guidelines](https://github.com/GeoNode/geonode/blob/master/CONTRIBUTING.md) 
- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in the documentation)
- [x] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

The following are required only for core and extension modules (they are welcomed, but not required, for contrib modules):
- [ ] There is a ticket in https://github.com/GeoNode/geonode/issues describing the issue/improvement/feature (a notable exemption is, changes not visible to end-users)
- [ ] The issue connected to the PR must have Labels and Milestone assigned
- [ ] PR for bug fixes and small new features are presented as a single commit
- [ ] Commit message must be in the form "[Fixes #<issue_number>] Title of the Issue"
- [ ] New unit tests have been added covering the changes, unless there is an explanation on why the tests are not necessary/implemented
- [ ] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [ ] This PR passes the QA checks: flake8 geonode
- [ ] Commits changing the **settings**, **UI**, **existing user workflows**, or adding **new functionality**, need to include documentation updates
- [ ] Commits adding **new texts** do use gettext and have updated .po / .mo files (without location infos)

**Submitting the PR does not require you to check all items, but by the time it gets merged, they should be either satisfied or inapplicable.**

PR #7843, opened 2021-07-21
Allow copying remote documents to local GeoNode when performing harvesting

ricardogsilva:
This PR adds initial support for copying remote resources to the local GeoNode.

It is a resubmission of the older geosolutions-it/geonode#864

The original PR description follows:

This PR adds support for copying remote documents to the local GeoNode, when performing harvesting.

The workflow suggested by the proposed implementation is roughly:

```python
harvested_info = worker.get_resource(harvestable_resource)
if worker.should_copy(harvestable_resource):
    copied_name = worker.copy_resource(harvestable_resource, harvested_info)
    harvested_info.copied_resources.append(copied_name)
worker.update_geonode_resource(harvested_info, harvestable_resource)
```

This workflow is implemented in the `geonode.harvesting.tasks._harvest_resource()` function.

Copying a remote resource is thus a feature that is optional - individual harvesting workers may implement it or not.

In the case of the `GeonodeLegacyHarvester` worker, this PR includes support for copying remote resources of type **document**. This can be configured by specifying `copy_documents=true` as an `additional_harvester_type_specific_configuration` property (it defaults to false). 

Support for copying remote layers is not included in this PR, as it seems that creation of new layers via the resource_manager is not implemented yet.


giohappy (changes requested):
fix conflicts

ricardogsilva:
@afabiani @giohappy 

We have another active PR for adding tests to the harvesting app (#7827) so I think the codecov fail is not too relevant here.

PR #7853, opened 2021-07-22
[Fixes #7832] Delete incomplete upload sessions after a configurable amount of time

### LINKED ISSUE #7832: Delete incomplete upload sessions after a configurable amount of time
Incomplete sessions (either in error state or waiting for inputs) accumulate indefinitely, with a list of bad side effects.
We should configure a deadline after which the session and any connected resource (file, models) are deleted.
I would set a default retention time of 24 hours, to be configured through Django settings.

asgharsir:

## Checklist

For all pull requests:

- [X] Confirm you have read the [contribution guidelines](https://github.com/GeoNode/geonode/blob/master/CONTRIBUTING.md) 
- [X] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in the documentation)
- [X] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

### DIFF
--- a/geonode/tasks/tasks.py
+++ b/geonode/tasks/tasks.py
@@ -204,9 +204,7 @@ def set_permissions(self, permissions_names, resources_names,
                 verbose=True
             )
 
-app.conf.beat_schedule = {
-    "delete-incomplete-session-resources": {
-        "task": "geonode.upload.tasks.delete_incomplete_session_uploads",
-        "schedule": crontab(hour='*/12'),
-    }
+settings.CELERY_BEAT_SCHEDULE["delete-incomplete-session-resources"] = {
+    "task": "geonode.upload.tasks.delete_incomplete_session_uploads",
+    "schedule": crontab(hour='*/12'),
 }
--- a/geonode/tasks/tasks.py
+++ b/geonode/tasks/tasks.py
@@ -19,7 +19,6 @@
 #########################################################################
 import celery
 from celery.utils.log import get_task_logger
-from celery.schedules import crontab
 
 from django.conf import settings
 from django.core.mail import send_mail
@@ -203,9 +202,3 @@ def set_permissions(self, permissions_names, resources_names,
                 delete_flag,
                 verbose=True
             )
-
-
-settings.CELERY_BEAT_SCHEDULE["delete-incomplete-session-resources"] = {
-    "task": "geonode.upload.tasks.delete_incomplete_session_uploads",
-    "schedule": crontab(hour='*/12'),
-}
--- a/geonode/tasks/tasks.py
+++ b/geonode/tasks/tasks.py
@@ -19,15 +19,14 @@
 #########################################################################
 import celery
 from celery.utils.log import get_task_logger
+from celery.schedules import crontab
 
 from django.conf import settings
 from django.core.mail import send_mail
-
-from django.db import (
-    connections,
-    transaction)
+from django.db import (connections, transaction)
 
 from geonode.celery_app import app
+
 
 try:
     import pylibmc
@@ -204,3 +203,10 @@ def set_permissions(self, permissions_names, resources_names,
                 delete_flag,
                 verbose=True
             )
+
+app.conf.beat_schedule = {
+    "delete-incomplete-session-resources": {
+        "task": "geonode.upload.tasks.delete_incomplete_session_uploads",
+        "schedule": crontab(hour='*/12'),
+    }
+}
--- a/geonode/tasks/tasks.py
+++ b/geonode/tasks/tasks.py
@@ -204,6 +204,7 @@ def set_permissions(self, permissions_names, resources_names,
                 verbose=True
             )
 
+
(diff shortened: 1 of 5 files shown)


afabiani (changes requested):
The attribute to check is `create_date` not `date`.
Please adapt the code accordingly.

geonode/tasks/tasks.py
```diff
@@ -204,3 +203,9 @@ def set_permissions(self, permissions_names, resources_names,
                 delete_flag,
                 verbose=True
             )
+
+
+settings.CELERY_BEAT_SCHEDULE["delete-incomplete-session-resources"] = {
```
afabiani, line 208: It does not make much sense to put this setting here.

Either:
1. We consider the `geonode.upload` module as a plugin, in that case the task should be defined on the `geonode.upload` module itself
2. We consider `geonode.upload` module as part of the cose, in that case let's harmonize with the others and place this into `geonode.settings`

afabiani (changes requested):

asgharsir:
> The attribute to check is `create_date` not `date`.
> Please adapt the code accordingly.

I am afraid `create_date` won't work for us. I thought `date` is the `update date`. If it is not `create_date` shall not be accurate. Let's see a scenario. A user created an upload 23 hours ago. But right now, in the last hour, he started making changes. So, he is in an active session still.

What is `date` field here BTW. Not update date?

geonode/tasks/tasks.py
asgharsir: I myself think, as suggested before, all the schedules should be at one place. But our `settings.py` is already over 2100 line. My humble suggestion is, bring even the other two settings in the standard place, `tasks`.

- First and foremost that is the expected place for this.
- We may not be able to adjust all future schedules in settings.py without more mess.
- Thirdly, even in `settings.py` they will not be at one place, if we see the way harvester is brought in. It has its own conditional place.
- Another option is not use `CELERY_BEAT_SCHEDULE` at all in the settings, rather use Celery's `app.config.beat_schedule`, so that they are not scattered. And we don't need to override settings at all. At this point, these are only a couple and we can do it easily.

afabiani:
> > The attribute to check is `create_date` not `date`.
> > Please adapt the code accordingly.
> 
> I am afraid `create_date` won't work for us. I thought `date` is the `update date`. If it is not, `create_date` shall not be accurate. Let's see a scenario. A user created an upload 23 hours ago. But right now, in the last hour, he started making changes. So even if he in the 25th or 28th hour, he is in an active session still.
> 
> What is the `date` field here BTW. Not update date?

`date` is updated every time the API reads the list of active upload sessions. I a session is stucked for some reason, the `date` will be updated anyway and it will never been removed.

geonode/tasks/tasks.py
afabiani: Nope, this will create confusion to people already using `settings` and also they would need to update their instances.

I prefer moving just this one to `geonode.upload.tasks`.

asgharsir: Let's go ahead for this one then.

asgharsir:
> > > The attribute to check is `create_date` not `date`.
> > > Please adapt the code accordingly.
> > 
> > 
> > I am afraid `create_date` won't work for us. I thought `date` is the `update date`. If it is not, `create_date` shall not be accurate. Let's see a scenario. A user created an upload 23 hours ago. But right now, in the last hour, he started making changes. So even if he in the 25th or 28th hour, he is in an active session still.
> > What is the `date` field here BTW. Not update date?
> 
> `date` is updated every time the API reads the list of active upload sessions. If a session is stucked for some reason, the `date` will be updated anyway and the upload session will never been removed.

So, we don't have an update date time. That would be tricky then. Would you like me to add one standard `updated` field with `auto_now`? That would be no bigger overhead then other solutions.

afabiani:
> > > > The attribute to check is `create_date` not `date`.
> > > > Please adapt the code accordingly.
> > > 
> > > 
> > > I am afraid `create_date` won't work for us. I thought `date` is the `update date`. If it is not, `create_date` shall not be accurate. Let's see a scenario. A user created an upload 23 hours ago. But right now, in the last hour, he started making changes. So even if he in the 25th or 28th hour, he is in an active session still.
> > > What is the `date` field here BTW. Not update date?
> > 
> > 
> > `date` is updated every time the API reads the list of active upload sessions. If a session is stucked for some reason, the `date` will be updated anyway and the upload session will never been removed.
> 
> So, we don't have an update date time. That would be tricky then. Would you like me to add one standard `updated` field with `auto_now`? That would be no bigger overhead then other solutions.

So I just had a deeper look at the code, and it looks like `date` is updated when the `update_from_session` method is invoked. So, that could be considered an _update date_. We can stick with that. Sorry for the noise.

asgharsir:
Thanks @afabiani , it rather gave me an opportunity to look into it further.

afabiani:
Superseded by https://github.com/GeoNode/geonode/pull/7867

PR #7856, opened 2021-07-22
Custom harvestable resource deletion

ricardogsilva:
This PR adds a `finalize_resource_deletion()` method to the `BaseHarvesterWorker` class.

This method is called whenever a harvestable resource is deleted and it allows a harvesting worker subclass to perform further actions just before the harvestable resource is deleted from the django DB. This is useful, for example, to ensure any custom resources have a chance to be deleted, and thus correctly implement the base class' `delete_orphan_resources_automatically` functionality.

NOTE: This is a re-submission of the now obsolete geosolutions-it/geonode#869


ricardogsilva:
@afabiani 

We have another active PR for adding tests to the harvesting app (#7827) so I think the codecov fail is not too relevant here.

giohappy:
@ricardogsilva new conflict to fix

ricardogsilva:
@giohappy conflict fixed. Can we merge this soon, so that I don't need to keep re-updating this PR?

PR #7859, opened 2021-07-23
[Fixes #7857] Separate upload execution from client actions and polling

### LINKED ISSUE #7857: Separate upload execution from client actions and polling
The current upload procedure depends on the execution steps invoked by the client. The async nature of the upload API is only partial, meaning that without the active collaboration of the upload page it won't be finalized. Moreover, status polling from the client operates on the same execution endpoints, which leads to interference that can make [race conditions arise](https://github.com/GeoNode/geonode/blob/3.3.x/geonode/upload/models.py#L224-L231).

The [upcoming Resource Service API](https://github.com/GeoNode/geonode/pull/7829) will solve these issues, but for the time being we need to keep using the current upload mechanism (client and backend), also considering that the new RS is only available on master branch, and we want to fix these issue also on 3.3.x.

The key aspects to address are:

 - make the upload procedure independent from client requests, except when user input is required to proceed (e.g. when a dataaset is eligible to be configured as a time series). For this task a [Celery chain](https://docs.celeryproject.org/en/latest/userguide/canvas.html#chains) might be a good candidate
 - separate status polling code from execution code

This should also solve https://github.com/GeoNode/geonode/issues/7836

afabiani:
References: #7857

## Checklist

For all pull requests:


geonode/maps/admin.py
```diff
@@ -50,6 +50,14 @@ class MapAdmin(TabbedTranslationAdmin):
     form = MapAdminForm
     actions = [metadata_batch_edit]
 
+    def delete_queryset(self, request, queryset):
+        """
+        We need to invoke the 'ResourceBase.delete' method even when deleting
+        through the admin batch action
+        """
+        for obj in queryset:
+            obj.delete()
```
giohappy, line 59: @afabiani isn't [queryset,delete()](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#delete) better here (and the other cases)?

geonode/upload/views.py
```diff
@@ -707,15 +707,25 @@ def view(req, step=None):
...
 
         resp = _steps[step](req, upload_session)
+
+        try:
+            content = resp.content
+            if isinstance(content, bytes):
+                content = content.decode('UTF-8')
+            resp_js = json.loads(content)
+
+            if not settings.ASYNC_SIGNALS:
```
giohappy, line 717: @afabiani does this mean that when async signals are activated we will rely on the Celery PeriodicTask to finalize the upload? If yes, this means that even quick uploads will take at least the time interval configured for the PeriodiTask to complete, right?
Moreover isn't the PeriodiTask going to concentrate a spike of work on Geoserver and the DB?

geonode/upload/views.py
```diff
@@ -707,15 +707,25 @@ def view(req, step=None):
...
+                if 'upload/final' in resp_js.get('redirect_to', ''):
```
giohappy, line 718: This looks like a hack. Don't we have a better solution rather then relying on the "redrect_to" content of the response? I know this is a quick solution to move the final step out of the client, but it's quite ugly to me...

geonode/maps/admin.py
afabiani: @giohappy this is intentional, if we are not going to call `delete()` on each single object, the bulk delete from the admin dashboard will completely bypass the overridden `delete` methods and signals... Django misteries

geonode/upload/views.py
afabiani: @giohappy correct and yes there will be spikes of work on GeoServer and DB because I'm going to parallelize the uploads through the `celery chord` directive.
However each upload will start after the files have been sent to the server and the Importer session has been created.
If the server is small, I strongly suggest to lower down the number of parallel workers.

geonode/upload/views.py
afabiani: Well actually this is the check done before by the upload workflow and, as far as I know, it is also the most reliable one.
The response from the final view, in the case everything was good and no further action is needed, will be always a JSON containing the `redirect_to` element.
In other words this is how the current `geoserver.upload` workflow has been implemented. I can see if we can check other variables, but I really don't think that in practice we will benefit from that except maybe code readability.

geonode/upload/views.py
giohappy: Honestly I don't like very much the idea of a PeriodicTask for this specific purpose. I guess you have chosen this solution for simplicity, given the current upload architecture, but I think that a Celery chain fits it better. Why do we need to make this periodic check when we have a well defined chain of states? 
Moreover, as from my previous comment, it means that we will have to wait multiples of 25 seconds for the completion of an upload, which most of the time will be a higher time compared to what expected for the average size of an upload.

On the other hand a PeriodicTask is the right tool for the check of stale sessions.

geonode/maps/admin.py
giohappy: I see... Not nice, but I undersatnd.

geonode/upload/views.py
giohappy: np, not nice code but I see your point.

geonode/upload/views.py
afabiani: @giohappy let's do a step back. How the current `geonode.upload` workflow is designed, having a celery `chain` is unuseful if not wrong.

For standard uploads we have only two steps:
1. Upload the file and create the Importer session accordingly, which is an atomic action
2. Finalize the Import and configure GeoNode

In the case during the step 1 GeoNode recognizes something missing (SRS or temporal dimension) the Importer Session is already created but GeoNode puts the upload into a `WAITING` state allowing the user to provide some further details.

After that it falls back to point 2.

The "heavy" part of the whole process is the point 2 and this one is currently composed by several configuration steps that are executed asynchronously by the periodic task.

Given that, architecturally speaking, I do not see any particular advantage on having Celery `chain` in the middle. Considering also that means re-factor completely the actual upload workflow logic.

It is true that there are 25 secs slots. We can avoid that by issuing the periodic tasks as soon as the Import session is ready eventually. We probably loose a bit of parallelism, but that is feasable.

giohappy: The point is why we really need a PeriodTask to run point 2? Maybe I miss some Celery background but I though that a simple task to be triggered when the the import sessions is ready was enough, without loosing parallelism. What is the advantage of a PeriodicTask on this aspect @afabiani?

afabiani: Two advantages mainly:

1. When uploading more than one dataset simultaneously, you don't need to wait the previous one to be finalized before starting processing the next one. Once you uploaded all the files, you can continue working with GeoNode waiting for the uploads to be finalized.
2. The last operation, wich means prepare the dataset on GeoNode, create the styles, the geofence rules, the thumbnail and so on, is the most time consuming operation. All this part is executed asynchronously, and the user is not blocked on the upload page waiting for all this stuff to be finalized.

giohappy: >1. When uploading more than one dataset simultaneously, you don't need to wait the previous one to be finalized before starting processing the next one. Once you uploaded all the files, you can continue working with GeoNode waiting for the uploads to be finalized.

Are you saying that in case of multiple uploads the whole upload procedure is managed as a single workflow? I thought distinct chains of tasks were spawned for each dataset....

>2. The last operation, wich means prepare the dataset on GeoNode, create the styles, the geofence rules, the thumbnail and so on, is the most time consuming operation. All this part is executed asynchronously, and the user is not blocked on the upload page waiting for all this stuff to be finalized.

A Celery task doesn't need an active browser session, so here I don't see where the PeriodTask has advantages over a simple task. I mean, when the `'upload/final'` response is received we can just run the finalization in a simple Celery task, isn't it?

afabiani: About point 1) not as a single workflow but sequentially. The next upload does not start until the previous one has finished. Moreover, how it is currently designed, the upload does not proceed until someone triggers the `final_view`. That means that if you leave the page early, the upload won't proceed.

About point 2) yes, that is feasible, it is basically what it already does if `ASYNC_SIGNALS=False`. The advantage of having a PeriodicTask is that if for some reason the process stops (server down or unresponsive, or whatever reason), the upload finalization can be resumed on a later stage.

As I suggested few points above, we can have both, start the async task suddenly when the import session is ready, and still run the periodic task which will simply skip the session already processing and remove the stale ones.

giohappy: Ok, let's go for the last option mentioned from you. Trigger the async task immediately and let the PeriodicTask to the housekeeping.

PR #7929, opened 2021-08-06
[FIXES #7976] Developing Commands App for Management Command Jobs via API

### LINKED ISSUE #7976: Create app for management commands
Refer Epic #909:
Issue refers to a proposal:
> "to implement a mechanism inside GeoNode to generate an HTTP API for any management command. The idea is to expose management over a common HTTP endpoint, granting of course full control of which commands we want to be exposed and under which security restrictions."

## Proposed Solution
* [x] Explore the [django-http-commands](https://github.com/geosolutions-it/django-http-commands) library to undestand the proposal.
* [x] Create an app called `mngmt_commands_http`.
* [x] Create a decorator `expose_command_over_http` to indicate that a specific command should be exposed or not.
* [x] Implement method to list the exposed commands.
    * Note: Using `django.core.management.get_command` to list the commands and a decorator on the commands that we want to be exposed.
    * Note: Only the commands with the decorator `@expose_command_over_http` are going to be exposed.
* [x] Create a Model for the ManagementCommandJob:
    * Fields: command, app_name, user, args, kwargs, output_message, start_time, end_time, celerytaskresult (FK).
    * Status options: Created, Queued, Running, Success, Failed.
    * Method: `start_task` --> Change status from `created` to `queued` and adds to the celery queue.
* [x] Create a method to handle the job running (`run_management_command`):
    * Redirects stdout and stderr to the `output_message` field.
    * Updated the status to --> `running` (on enter) and `success` or `fail` (on exit). 
* [x] Create a single celery async task: Call the `run_management_command` method.
* [x] Create an APIView, providing the following endpoints:
    * [x]  `GET` `/api/v2/management/` : **List** exposed commands
    * [x] `GET` `/api/v2/management/<cmd_name>`: Get the **help message** of a given cmd_name (if it's exposed)
    * [x] `POST` `/api/v2/management/<cmd_name>`: **Create** and **run** an async job with the given command. 
* [x] Tests and validation

##  Alternative solutions considered.
On the discussion of the issue #920, some alternative solutions were considered.
* ~Using `django-http-commands` as external library~ --> The approach to add an app alongside the geonode apps was chosen.
* Not creating a model for the ManagementCommands -->  Following this approach because all the management commands can be listed by the `django.core.management.get_command` django built-in function.
* ~Not creating a model for the ManagementCommandJobs~ --> A model will be created, although celery's TaskResult model is already storing some data related to the task, it's better to store these requests separately and to have control over the celery execution. Also, following this approach, we can show/control the results in the django-admin page and use a ModelViewSet (See pictures below).

## Additional context
* Admin view (list) for ManagementCommandJobs
image
* Admin view (details)
image

* Response of `POST` `/api/v2/management/<cmd_name>`
```json
{
    "success": true,
    "error": null,
    "data": {
        "id": 129,
        "command": "sleep_some_minutes",
        "app_name": "geonode.base",
        "status": "QUEUED",
        "args": "[0.1]",
        "kwargs": "{}"
    }
}
```

asgharsir:

## Checklist

For all pull requests:


asgharsir:
This is just a WIP, in haste, on @afabiani request. Test are yet to be completed and mess is yet to be cleaned (as not even flake is done yet). @mattiagiupponi please.

geonode/commands/tests/test_views.py
```diff
@@ -0,0 +1,53 @@
...
+        self.standard_user = get_user_model().objects.create(
+            username="test", email="test@test.com"
+        )
+        self.admin_user, _ = get_user_model().objects.get_or_create(
+            username="super", is_staff=True, is_superuser=True
+        )
+
+
+    def test_permissions(self):
+        cases = (
```
mattiagiupponi, line 42: I'll avoid adding more tests cases in a single test.
Potentially here you have 3 different tests. Merging them all together will be a mess if you have to debug, plus is easier to read them even with the verbose. Potentially here you have:

```python
def test_anonymous_users_is_unauthorized(self):
    '''
    Anonymous users cannot access to the API
    '''
    pass
def test_normal_user_is_unauthorized(self):
    '''
    Standard users cannot access to the API
    '''
    pass
def test_admin_user_can_see_the_api(self):
    '''
    Admin user can access to the API
    '''
    pass
```

geonode/commands/urls.py
```diff
@@ -0,0 +1,27 @@
...
+#
+#########################################################################
+
+from django.urls import path
+from .views import CommandView
+
+app_name = 'geonode.commands'
+
+urlpatterns = [
+    path("enque/", CommandView.as_view(), name="enque"),
```
mattiagiupponi, line 26: Typo: `enque` -> `enqueue`

geonode/commands/views.py
```diff
@@ -0,0 +1,86 @@
...
+        "cmd": "test_command",
+        "args": ["delta"],
+        "kwargs": {
+                "cpair": [10, 40],
+                "ppair": ["a", "b"]
+            }
+        }
+        """
+        
+        args = request.data.get("args", [])
```
mattiagiupponi, line 59: We can think about having this part into a `serializer` with his own `validation` method

mattiagiupponi (changes requested):

geonode/commands/tests/test_views.py
asgharsir: Hmm. We are using `subtest` here. Which handles this parametrization for us for each case. So, for debugging when a test fails, test specify the case where it failed. This also helps readability also. We read far less code.
https://docs.python.org/3/library/unittest.html#subtests

geonode/commands/views.py
asgharsir: But we are not using any model. That's why we are settled to `APIView` also.
Even otherwise, the `BaseCommand` class does the error handling, and that validation has specialization of such Python CLI args.

mattiagiupponi: Can't we use an abstract model to handle this part? Is just a matter of making it a bit more robust :)

asgharsir: Again why we need to make abstract model when handling is already handled by `BaseCommand`.
And if it such necessary, best option is the model itself, and you have serializer also in the package.

mattiagiupponi (changes requested):
Before continue, the following point should be fixed:
- The title is not in the standard format `"[Fixes #<issue_number>] Title of the Issue"`
- The issue in the GeoNode master project is missing
- The build is failing, please give a check to the tests and fix them
- Convert the issue to ready-to-review

asgharsir:
> Before continue, the following point should be fixed:
> 
> * The title is not in the standard format `"[Fixes #<issue_number>] Title of the Issue"`

As the Epic was placed in a wrong repo, so the title was to be fixed after consultation of the guys returning from vacation.
> * The issue in the GeoNode master project is missing

As above. Now created. But before attaching the new issue here, I would need suggestion if we better create new branch, as the name `issue-921` would be misleading.

For the tests, I could not reproduce the errors, with even Celery + Redis running locally. I will now make a test commit to verify some doubts.

mattiagiupponi:
> As above. Now created. But before attaching the new issue here, I would need suggestion if we better create new branch, as the name `issue-921` would be misleading.

Yes, rename the branch once you have created the issue on geonode

> For the tests, I could not reproduce the errors, with even Celery + Redis running locally. I will now make a test commit to verify some doubts.

It looks like a temporary error since the last build on master (1 hour ago) is all green.

PR #7947, opened 2021-08-13
[Fixes #7945] Ingest harvested layer data to geonode

### LINKED ISSUE #7945: [Harvesting - GeoNode Client] Harvesting of layers data
Based on https://github.com/geosolutions-it/nexus-geonode/issues/178

meomancer:
References: #7945

This PR implements copying of remote layers and documents in the `GeonodeLegacyHarvester`. It also includes a number of bug fixes.

Copying remote data is configured by means of editing a `Harvester`'s `harvester type specific configuration` parameter. The following additional configuration parameters are now available:

- `copy_documents`
- `copy_datasets`

They both default to `False`, which means that by default a harvester will not copy remote data. Setting them to `True` will cause GeoNode to copy the remote data when performing harvesting.

The video below shows the proposed implementation in action:

https://user-images.githubusercontent.com/732010/134215867-fcc81dc6-d51e-4521-bcf1-c00d359e9771.mp4

In the video I am demonstrating harvesting with two harvesters simultaneously. One of the harvesters (`copys pcrafi`) performs a copy of the remote data onto the local GeoNode, while the other (`links pcrafi`) merely copies the metadata. The video also demonstrates an error with the copying of a remote raster (the error is intended here) and how this is recorded in the corresponding harvesting session. The harvested resources appear on the GeoNode frontend.

#### Improvements

- The `HarvestingSession` model is now more useful, as it records the progress of harvesting operations as they are on-going
- The harvesting section of the django admin now includes a number of additional fields in order to ease navigation between harvesters, harvesting sessions and harvestable resources.
- It is now possible to harvest individual harvestable resources directly from the admin section without needing to go to the harvester page. This is mainly useful for testing.

#### Fixes

Metadata fields for resource's thumbnail, URL extension (for documents) and subtype (for datasets) are now correctly provided.

When harvesting is merely referencing the remote resources (_i.e._ we are not copying the actual data, just the metadata):

- [x] Only metadata is copied over to the local GeoNode
- [x] Remote thumbnail is used (fixes geosolutions/nexus-geonode#450)
- [x] Remote resource URL is used

#### Notes

In the above video we can see that for remote documents, the GeoNode frontend offers a download of the file in the main page. This seems like a bug of the frontend, so I'm leaving it out of the current PR's scope (which is already quite big).

fixes #7945


meomancer:
This fix https://github.com/GeoNode/geonode/issues/7945

giohappy:
@ricardogsilva can you please review the comment from @meomancer?

afabiani (changes requested):
please check out the warnings from LGTM

https://lgtm.com/projects/g/GeoNode/geonode/rev/pr-0b2e9cfca6d7c63b7098557e39710a3a1b5e27ca

ricardogsilva (dismissed):
Looks good to me, but I'm biased, since I took over from @meomancer and implemented most of the changes :stuck_out_tongue:

afabiani:
@ricardogsilva @meomancer any chance to have a look at the LGTM issues [here](https://lgtm.com/projects/g/GeoNode/geonode/rev/pr-7b6bc6466cbf90340208aa36d5e88a5b4f21dddc)?

ricardogsilva:
> @ricardogsilva @meomancer any chance to have a look at the LGTM issues [here](https://lgtm.com/projects/g/GeoNode/geonode/rev/pr-7b6bc6466cbf90340208aa36d5e88a5b4f21dddc)?

I've checked them and they look kind of dumb tbh

- Most warnings are about preventing a circular import which will never happen because the file being imported (_i.e._ `harvesting.harvesters.base`) is 100% required in the `harvesting.harvesters.geonode` file (as all harvesters must derive from the base one). The other way around - importing `geonode` harvester from `base` is absurd and will also never happen. I think maybe `lgtm` is getting confused by the name of the file, which is simply `geonode`. I will rename it to `geonodeharvester` to see if that is helpful;
- The other warnings are about logging of a layer's `uuid`, which I can understand could be interpreted as a security warning, but it also provides relevant feedback in this case.

ricardogsilva:
@afabiani @giohappy my latest push introduces a rename of the file where the `GeonodeLegacyHarvester` resides from `geonode` to `geonodeharvester` - I'm trying to make lgtm happy with this.

I'll be pushing a django data migration shortly that will deal with previously exisitng harvester configs in the DB.

afabiani (changes requested):
So far it almost works fine... I noticed that in the case I leave the option

## ISSUE 1

```json
{
   "harvest_datasets": true,
   "copy_datasets": false
}
```

The layer won't be populated not accessible.

E.g.: by using the demo instance https://development.demo.geonode.org/ as a source I get the following

image

image

The dataset is empty.

## ISSUE 2
Trying to harvest a Map fails with an exception, the session and the harvester are then blocker forever.

## ISSUE 3
Whenever an harvest session fails, there's no way to recover the harvester, which remains busy forever.

giohappy:
Ragarding ISSUE 3 this is a long standing request that was already raised in the past @ricardogsilva. If a job breaks there's no way to recover it but going into the DB and manually changing the status back to `ready`. If we need a dedicated issue for this please open it @ricardogsilva.

ricardogsilva:
@afabiani I will make the necessary adjustments. Some clarification wrt the issues you raise:

for **Issue 1**, the current `GeonodeLegacyHarvester` has been designed for harvesting older GeoNode (read pre-v2 API) instances. The fact that it does not work on the development demo of GeoNode, while unfortunate, is not too surprising. The plan is to create another worker that is able to serve modern GeoNode instances - the code will be a lot cleaner, since this legacy harvester has some additional complexity to compensate for the fact that the GeoNode API used to be a bit rough. I think the ideal solution will be to have a single harvester that is able to distinguish between older and newer GeoNode deployments. We ar e not there yet though.

For *issue 2**, that is a clear bug, which I will fix ASAP. The current harvester worker does not claim to support maps, so it should just fail in an orderly fashion.

For **issue 3**, that is a by-product of the bug described above probably. Generally speaking, the code is trying to catch known exceptions in the code so that it avoids entering that invalid state. The fact that it is still happening is a sign of some error in the code which is not being properly caught.

giohappy:
@ricardogsilva regarding ISSUE 1, I expect the legacy harvester to work even on 3.3.x (the development demo) and master, since nothing from the old API and CSW has been dropped so far.

afabiani:
btw, any chance to fix also this one while looking for the other issues?

```
======================================================================
FAIL: test_perform_harvesting (geonode.harvesting.tests.test_admin.HarvesterAdminTestCase)
----------------------------------------------------------------------
  File "/usr/local/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "/usr/src/geonode/geonode/harvesting/tests/test_admin.py", line 73, in test_perform_harvesting
    self.assertEqual(self.harvester.status, models.Harvester.STATUS_PERFORMING_HARVESTING)
AssertionError: 'ready' != 'harvesting-resources'
- ready
+ harvesting-resources
```

afabiani:
## ISSUE 4
Raster datasets seem to be not harvested correctly even when the `copy_dataeset` option has been set to `True`

image

ricardogsilva:
@afabiani 

My latest commit should fix the issues you reported.

I fixed some missing metadata fields not being correctly passed to the resource manager.

I also disabled harvesting of maps for the GeoNode harvester. The reason for this is that API does not really provide enough information for us to be able to replicate a map locally - there is no info about the map's layers, etc.

To the best of my knowledge documents and datasets are now correctly harvested, regardless of being of type `vector` or `raster` and regardless of them being copied or just referenced from the remote.

I cannot guarantee 100% that this works for all GeoNode remote deployments, as I'm not sure how stable the old API had been between releases. Seems to work OK with those deployments that I am using as a test bed.

As for the failing test, I don't seem to get it locally. I'm waiting to let the CI build be done to check it output.

afabiani:
@ricardogsilva there still something not fully working with the remote resources and in particular with the WMS harvester, but I'll postpone those fixes to https://github.com/GeoNode/geonode/pull/8149 which will include this PR too.

afabiani (approved):

PR #8019, opened 2021-08-30
[Fixes #8018] FeatureInfoTemplate from REST API v2 not properly working

### LINKED ISSUE #8018: FeatureInfoTemplate from REST API v2 not properly working
## Expected Behavior
 - If FeatureInfoTemplate is enabled I'd expect API return the custom HTML template
 - If FeatureInfoTemplate is disabled or None, I'd expect the API return an HTML template formatted accordingly to the attributes set and the selected "labels" and "types"

## Actual Behavior
 - If a FeatureInfoTemplate is present, the API return it despite it is enabled or not
 - If FeatureInfoTemplate is disabled the API completely ignore the attributes set settings by the user

  1. Edit a vectorial dataset metadata and provide some labels and types to the attributes
  2. Click on WMS GetFeatureInfo and notice that nothing changes
  3. Edit the metadata again and create a custom HTML template
  4. From now on the custom template will be applied despite it is enabled or not

## Specifications

  - GeoNode version: 4.0.0
  - Installation method (manual, GeoNode Docker, SPCGeoNode Docker): any
  - Platform: any
  - Additional details:

afabiani:
References: #8018

## Checklist

For all pull requests:

### DIFF
--- a/geonode/layers/api/serializers.py
+++ b/geonode/layers/api/serializers.py
@@ -93,31 +93,31 @@ class FeatureInfoTemplateField(DynamicComputedField):
                     elif _field.featureinfo_type == Attribute.TYPE_IMAGE:
                         _template += '<div class="col-xs-12" align="center" style="font-weight: bold; word-wrap: break-word;"> \
                             <a href="${properties.%s}" target="_new"><img width="100%%" height="auto" src="${properties.%s}" title="%s" alt="%s"/></a></div>' % \
-                            (_field, _field, _label, _label)
+                            (_field.attribute, _field.attribute, _label, _label)
                     elif _field.featureinfo_type in (
                             Attribute.TYPE_VIDEO_3GP, Attribute.TYPE_VIDEO_FLV, Attribute.TYPE_VIDEO_MP4,
                             Attribute.TYPE_VIDEO_OGG, Attribute.TYPE_VIDEO_WEBM, Attribute.TYPE_VIDEO_YOUTUBE):
                         if 'youtube' in _field.featureinfo_type:
                             _template += '<div class="col-xs-12" align="center" style="font-weight: bold; word-wrap: break-word;"> \
                                 <iframe src="${properties.%s}" width="100%%" height="360" frameborder="0" allowfullscreen></iframe></div>' % \
-                                (_field)
+                                (_field.attribute)
                         else:
                             _type = f"video/{_field.featureinfo_type[11:]}"
                             _template += '<div class="col-xs-12" align="center" style="font-weight: bold; word-wrap: break-word;"> \
                                 <video width="100%%" height="360" controls><source src="${properties.%s}" type="%s">Your browser does not support the video tag.</video></div>' % \
-                                (_field, _type)
+                                (_field.attribute, _type)
                     elif _field.featureinfo_type == Attribute.TYPE_AUDIO:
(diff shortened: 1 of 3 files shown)


giohappy (approved):

geonode/layers/api/serializers.py
```diff
@@ -71,6 +71,59 @@ class Meta:
...
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def get_attribute(self, instance):
+        if instance.use_featureinfo_custom_template and instance.featureinfo_custom_template:
+            return instance.featureinfo_custom_template
+        else:
+            _attributes = instance.attributes.filter(visible=True).order_by('display_order')
+            if _attributes.count():
```
mattiagiupponi, line 84: it would be better using the `.exists()` instead of the `.count()`

geonode/layers/api/serializers.py
```diff
@@ -71,6 +71,59 @@ class Meta:
...
+                _template = '<div>'
+                for _field in _attributes:
+                    _label = _field.attribute_label or _field.attribute
+                    _template += '<div class="row">'
+                    if _field.featureinfo_type == Attribute.TYPE_HREF:
+                        _template += '<div class="col-xs-6" style="font-weight: bold; word-wrap: break-word;">%s:</div> \
```
mattiagiupponi, line 90: I would consider putting in common the `<div>` tag since only the values are changing and use the `format` string to replace the values. Maybe something like:
```python
_div_structure = "<div class="{}" style="font-weight: bold; word-wrap: break-word;">%s:</div>"
...... other code ....
_template += _div_structure.format("col-xs-6") \
```
Handling HTML templates in the code is always a mess, let's try to make in common as many things as possible!

geonode/layers/api/serializers.py
afabiani: wrong, we want to print out a template only if there's at least one attribute.

geonode/layers/api/serializers.py
afabiani: I don't think the `div` is always the same one... however I will double check.

geonode/layers/api/tests.py
```diff
@@ -61,13 +61,59 @@ def test_datasets(self):
...
         # Test detail response has attribute_set
-        url = urljoin(f"{reverse('datasets-list')}/", f"{Dataset.objects.first().pk}")
+        url = urljoin(f"{reverse('datasets-list')}/", f"{_dataset.pk}")
         response = self.client.get(url, format='json')
         self.assertIsNotNone(response.data['dataset'].get('ptype'))
         self.assertIsNotNone(response.data['dataset'].get('subtype'))
         self.assertIsNotNone(response.data['dataset'].get('attribute_set'))
 
+        # Test "featureinfo_custom_template"
+        _attribute, _ = Attribute.objects.get_or_create(dataset=_dataset, attribute='name')
```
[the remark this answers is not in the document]
afabiani: I prefer leave it like this and avoid an additional call to the `setupClass`. We are already testing here the returns from the datasets apis.

geonode/layers/api/serializers.py
afabiani: updated due to https://stackoverflow.com/a/23585057/7597536

geonode/layers/api/serializers.py
afabiani: the divs are all different, for the time being I'll leave like this.

PR #8052, opened 2021-09-06
[Fixes GeoNode#7976] Create app for management commands

italogsfernandes:
## What was done

* An app for running management commands was created.

## How to test

* Django Testing
```
./manage.py test geonode.management_commands_http --keepdb
```

* cURL:
```
curl --location --request POST '{{BASE_URL}}/api/v2/management/ping_mngmt_commands_http/' \
--header 'Authorization: Basic <INSERT_HERE_YOUR_AUTH_TOKEN>' \
--data-raw '{
    "args": ["--sleep", "30"],
    "kwargs": {},
    "autostart": true
}'
```
* You can see a list o jobs and details at: `/admin/mngmt_commands_http/managementcommandjob/`

## Checklist

For all pull requests:


geonode/mngmt_commands_http/admin.py
```diff
@@ -0,0 +1,82 @@
...
+        "celery_traceback",
+    )
+    actions = ["execute", "stop"]
+
+    def execute(self, request, queryset):
+        for job in queryset.iterator():
+            job.start_task()
+
+    def stop(self, request, queryset):
+        for job in queryset.iterator():
```
mattiagiupponi, line 69: The queryset is already an iterable, should be enough to use: `for job in queryset`

geonode/mngmt_commands_http/decorators.py
```diff
@@ -0,0 +1,31 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+#########################################################################
+class expose_command_over_http:
```
mattiagiupponi, line 19: I don't know if the idea was to have them exposed by default or not. Better have the @giohappy opinions here

geonode/mngmt_commands_http/management/commands/ping_mngmt_commands_http.py
```diff
@@ -0,0 +1,42 @@
...
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+#########################################################################
+from time import sleep
+from django.core.management.base import BaseCommand
+from geonode.mngmt_commands_http.decorators import expose_command_over_http
+
+
+@expose_command_over_http
+class Command(BaseCommand):
```
mattiagiupponi, line 25: This looks like more a "test" command than a real one. I'd consider moving this one in the tests package then keeping it as an exposed command. This one is very useful in a "smoke test"

geonode/mngmt_commands_http/models.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    def __str__(self):
+        return (
+            f"ManagementCommandJob"
+            f"({self.id}, {self.command}, {self.user}, {self.created_at})"
+        )
+
+    def start_task(self):
+        from .tasks import run_management_command_async
```
mattiagiupponi, line 68: Consider moving those local imports at the top. Having them as `local` means that every time the function is called, the module is reimported. This can cause high memory usage (the same for the `stop_task` and `celery_task_meta` def)

geonode/mngmt_commands_http/tests.py
```diff
@@ -0,0 +1,135 @@
...
+        resource_url = f"/api/v2/management/{cmd_name}/"
+        response = self.client.get(resource_url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        response_json = response.json()
+        self.assertTrue(response_json['success'])
+        self.assertIn(cmd_name, response_json['data'])
+
+    def test_management_commands_detail_not_found(self):
+        cmd_name = "some_unavaliable_command"
+        resource_url = f"/api/v2/management/{cmd_name}/"
```
mattiagiupponi, line 68: If you want to make the resource_url dynamic, you can declare it in the `setUp` as -> `self.resource_url = "/api/v2/management/{}` and then in each function use the `format` to compone it example: `self.client.get(resource_url.format(cmd_name ))`

mattiagiupponi (changes requested):

geonode/mngmt_commands_http/admin.py
```diff
@@ -0,0 +1,82 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+#########################################################################
+from django.contrib import admin
+
+from .models import ManagementCommandJob
```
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/mngmt_commands_http/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+#########################################################################
+from django.apps import AppConfig
+
+
+class MngmtCommandsHttpAppConfig(AppConfig):
```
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/mngmt_commands_http/admin.py
italogsfernandes: Updated ✅

geonode/mngmt_commands_http/models.py
italogsfernandes: Updated ✅

geonode/management_commands_http/utils/job_runner.py
```diff
@@ -0,0 +1,82 @@
...
+        self.job.save()
+
+
+def run_management_command(job_id, async_result_id: str = ""):
+    """
+    Loads the job model from database and run it using `call_command` inside a
+    context responsible to updating the status and redirecting stdout and
+    stderr.
+    """
+    job = ManagementCommandJob.objects.get(id=job_id)
```
mattiagiupponi, line 77: In this case, if the `job_id` does not exist it will raise an `Exception`. Consider adding a check, something like:
```python
job = ManagementCommandJob.objects.filter(id=job_id)
if not job.exists():
    return controlled exception or return false
 .....
other code
 .....
    with JobRunner(job.get(), output, async_result_id):
```

mattiagiupponi (changes requested):

mattiagiupponi:
@italogsfernandes 
- Looks like that is not possible to create a new management command job from the Admin Interface because the `has_add_permission` is set to false. This will be implemented later?
- with a normal geonode configuration the job is working as expected, but after activating the async workflow in geonode, the management command job is stuck at `QUEUED` status. Can you give me a check on it?
(set `ASYNC_SIGNALS=True`, run `memcache`, `rabbitmq` and run celery with `celery_dev.sh`)

geonode/mngmt_commands_http/management/commands/ping_mngmt_commands_http.py
italogsfernandes: @mattiagiupponi I didn't get the idea here.
* Do you mean move it to another app or inside a folder `tests` of this app?
* Django does not support management commands inside the `tests` folder, so we would not be able to use or expose it.

italogsfernandes:
> @italogsfernandes
> 
>     * Looks like that is not possible to create a new management command job from the Admin Interface because the `has_add_permission` is set to false. This will be implemented later?
> 
>     * with a normal geonode configuration the job is working as expected, but after activating the async workflow in geonode, the management command job is stuck at `QUEUED` status. Can you give me a check on it?
>       (set `ASYNC_SIGNALS=True`, run `memcache`, `rabbitmq` and run celery with `celery_dev.sh`)

* Create from the Admin Interface: Yes, this will be implemented later.
Issue 950 of the Epic (geosolutions-it fork), https://github.com/geosolutions-it/geonode/issues/950

* job stuck when activating async workflow: Ok, checking...

geonode/mngmt_commands_http/management/commands/ping_mngmt_commands_http.py
mattiagiupponi: Yes, the idea was to move it under the test package, but if Django does not support it just keep it as is

geonode/management_commands_http/utils/job_runner.py
italogsfernandes: Updated ✅

* Remarks:
In this case, as an approach to explicit the `Exception` case, it's better to handle the `Object.DoesNotExist` exception. 
Using `queryset.exists` and then `queryset.get`, two SQL queries will be sent to the database (and others django ORM differences). 
```
    try:
        job = ManagementCommandJob.objects.get(id=job_id)
    except ManagementCommandJob.DoesNotExist:
        raise exceptions.NotFound(
            f"ManagementCommandJob with id {job_id} was not found."
        )
```

geonode/management_commands_http/tests.py
```diff
@@ -0,0 +1,143 @@
...
+        resource_url = f"/api/v2/management/{cmd_name}/"
+        response = self.client.post(resource_url)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        response_json = response.json()
+        self.assertTrue(response_json['success'])
+        self.assertEqual(response_json['data']["command"], cmd_name)
+        self.assertEqual(
+            response_json['data']["status"], ManagementCommandJob.QUEUED
+        )
+        mocked_async_task.delay.assert_called_once()
```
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/mngmt_commands_http/tests.py
italogsfernandes: Updated ✅

geonode/management_commands_http/utils/job_runner.py
mattiagiupponi: Yes, the exists is placed for performance reasons, it will hit the DB but as Django doc says `This tries to perform the query in the simplest and fastest way possible`
Mode details here: https://stackoverflow.com/questions/6868939/in-django-what-is-the-most-efficient-way-to-check-for-an-empty-query-set/23585057#23585057
Anyway, in this case if fine to have the Exception since we want that the exception is raised

geonode/mngmt_commands_http/decorators.py
giohappy: The best would be a configuration inside the settings, where we whitelist the commands to be exposed.

On this regard @italogsfernandes I haven't gone through the implementation but I suppose you have configured the authentication and authorizations required to call the commands. Can you tell me how it is managed?

italogsfernandes: Only admin users can call the commands (`permission_classes = [permissions.IsAdminUser]` - [Test Code](https://github.com/italogsfernandes/geonode/blob/cbc33020cabb8d1ebeca11fd3f542bfbd6be433e/geonode/management_commands_http/tests.py#L55:L58)).

giohappy: ok, thx @italogsfernandes. Do you think we can implement the configuration of exposed commands through settings?

italogsfernandes: Yes, I'm working on it.  👨‍💻

Something like the following config inside `settings.py`: 
```python
MANAGEMENT_COMMANDS_EXPOSED_OVER_HTTP = [
    "ping_mngmt_commands_http",
    "importlayers",  # for example...
]
```

Besides "ping_mngmt_commands_http" which other commands would be interesting to be on this list by default? 

[commands.md](https://github.com/GeoNode/geonode/files/7178528/commands.md)

giohappy: I'd say the following:

- `sync_geonode_datasets`
- `sync_geonode_maps`
- `importlayers`
- `set_all_datasets_metadata`

italogsfernandes:
> @italogsfernandes
> 
>     * Looks like that is not possible to create a new management command job from the Admin Interface because the `has_add_permission` is set to false. This will be implemented later?
> 
>     * with a normal geonode configuration the job is working as expected, but after activating the async workflow in geonode, the management command job is stuck at `QUEUED` status. Can you give me a check on it?
>       (set `ASYNC_SIGNALS=True`, run `memcache`, `rabbitmq` and run celery with `celery_dev.sh`)

@mattiagiupponi  I just updated the task. Hope it's working fine now. :rocket: 
Could you try again and verify if it's solved?

geonode/mngmt_commands_http/decorators.py
italogsfernandes: @giohappy changed on bb13b3b856a48083045f36ff2d5a8e4ae568f9f0

geonode/settings.py
```diff
@@ -2144,3 +2146,14 @@ def get_geonode_catalogue_service():
...
+# If a command name is listed here, the command will be available to admins over http
+# This list is used by the management_commands_http app
+MANAGEMENT_COMMANDS_EXPOSED_OVER_HTTP = [
+    "ping_mngmt_commands_http",
+    "updatelayers",
+    "sync_geonode_layers",
+    "sync_geonode_maps",
+    "importlayers",
+    "set_all_layers_metadata",
+]
```
mattiagiupponi, line 2159: I'm thinking out loud, is it worth having this as an environment variable? it will be something like (by removing duplicates):

`MANAGEMENT_COMMANDS_EXPOSED_OVER_HTTP = defaults + os.get_env('LIST_OF_COMMANDS_FROM_ENV', [])`

mattiagiupponi:
> @mattiagiupponi I just updated the task. Hope it's working fine now. 🚀
> Could you try again and verify if it's solved?

@italogsfernandes I tried to run the `importlayers` management command, but the job stuck in `PENDING` status (the `ping_mngmt_commands_http` for example is working fine).

image

Two strange behaviours:
- The JSON response return this as `args` value -> ` "args": "\"/home/mattia/gis_data/single_point/\"",`
- The Celery ID, is not available in the Celery Task admin page

Note: by running from the command line the command is working as expected
```python
python manage.py importlayers /home/mattia/gis_data/single_point/
**geonode logs**

Finished processing 1 layers in 15.96 seconds.

1 Created layers
0 Updated layers
0 Skipped layers
0 Failed layers
15.959701 seconds per layer
```

mattiagiupponi:
@italogsfernandes
I tested it in a normal flow (async flow false) and looks like that it's taking in the wrong way the input path.
The import layers loop over all the files available in the given path and upload the layers into geonode. So the options['path'] is expecting something like `["/home/mattia/gis_gisdata/singlepoint/"]` while it's looping over this: `['/', 'h', 'o', 'm', 'e', '/', 'm', 'a', 't', 't', 'i', 'a', '/', 'g', ...]`
There is probably something in the command parsing, this is why is not working properly

italogsfernandes:
@mattiagiupponi The new checks were added.
Also:
* [x] Setting `MANAGEMENT_COMMANDS_EXPOSED_OVER_HTTP` loaded from env. 
* [x] When using `ASYNC_SIGNALS = False` the job were not up-to-date in the return. Fixed.
* [x] Checks added for the serializer.
* [x] Possibility to search in admin by the `output_result`.

Hope I didn't forget anything. :)

mattiagiupponi (approved):

giohappy (approved):

PR #8128, opened 2021-09-16
Revert "[master][Fixes #8112] Dump and restore Geofence rules in case of requests or …"

### LINKED ISSUE #8112: Dump and restore Geofence rules in case of requests or transaction failures
## Expected Behavior
When updating/changing the layer permissions rules, it should communicate with Geofence in the following way:
* Purge all Geofence permission rules.
* Create the new rules one by one.
* Restore rules in case of failure.

## Actual Behavior
* Purge all Geofence permission rules.
* Create the new rules one by one.
* If something goes wrong: 
    * The Geofence rules are let in an inconsistent state. For example: the new rules are not created after the purge.

  1. Create a Layer.
  2. Change Layer Permissions.
  3. During the update, add some intentional error:
      * For example: Edit the code.
  4. The rule at Geofence will be in an inconsistent state.

## Specifications

  - GeoNode version: 3.3.x, 3.2.x and also master
  - Installation method (manual, GeoNode Docker, SPCGeoNode Docker): GeoNode Docker
  - Platform:
  - Additional details:

afabiani:
Reverts GeoNode/geonode#8123

### DIFF
--- a/geonode/geoserver/manager.py
+++ b/geonode/geoserver/manager.py
@@ -67,7 +67,6 @@ from .helpers import (
     set_attributes_from_geoserver,
     create_geoserver_db_featurestore)
 from .security import (
-    GeofenceLayerRulesUnitOfWork,
     _get_gf_services,
     _get_gwc_filters_and_formats,
     get_user_geolimits,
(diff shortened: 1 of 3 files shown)


afabiani:
wrongly merged

PR #8176, opened 2021-09-27
Allow aborting of on-going harvesting tasks

### LINKED ISSUE #7955: Implement Django admin action to stop running tasks
Based on https://github.com/geosolutions-it/nexus-geonode/issues/351

## Overview
Admin must be able to stop harvestable resources retrieval and ingestion.
In this moment the only option is either to use the session API from an API client, or kill the task abruptly and restore the harvester status manually inside the DB, We need a Django admin action to do it.

ricardogsilva:
This PR implements aborting of on-going harvesting-related asynchronous tasks.

It improves the existing concept of a harvesting session by implementing the `AsynchronousHarvestingSession` model. This model is now used for the two types of async operations supported by a harvester:

- Refreshing the list of harvestable resources that are available for being harvested on the remote;
- Performing the actual harvesting of the selected remote resources.

Each session also stores some useful properties and can be used to get a sense of progress. For now, some of these properties are also being reflected on the `Harvester` model. In the future we may remove them from the harvester and use the session exclusively, as it seems like a more natural place to store them.

Despite introducing better support for harvesting-related sessions, the `HarvesterAdmin` continues to be the main place to kickoff/abort new sessions. As such, the admin section for `harvesting.harvesters` gains new actions for aborting on-going sessions. An aborted session is signaled as `aborted` by means of its `status` field. On the API side this means that the harvesting sessions continue to be read-only and that the `harvesters/{harvester-id}` continues to be the endpoint to reach for whenever starting/aborting a new session - this is done via sending HTTP PATCH requests that modify the harvester's `status` field and was not changed.

Another notable change introduced by this PR is that when a harvester is created via the django admin it no longer automatically updates its underlying harvestable resources list. The user must now explicitly ask for a refresh of the harvestable resources. This change was introduced to both a) make the code cleaner and b) give the user one more chance to review the harvester parameters after creation, without forcing him to wait for a potentially long time before the harvester becomes ready again.
 

fixes #7955


ricardogsilva:
@afabiani the drop in coverage seems to be unrelated to this PR. Perhaps due to other parts of the codebase changing? Or maybe I am reading things wrong?

afabiani:
@ricardogsilva no worries, I really don't think it is depending by this PR

afabiani:
@ricardogsilva should we worried about that 
https://lgtm.com/projects/g/GeoNode/geonode/rev/pr-188ed5bc3da3aca1edeeb3a2355e430771ca961b

ricardogsilva:
@afabiani yeah let me fix that before you move on

ricardogsilva:
@afabiani should be fixed now

ricardogsilva:
@afabiani I accidentally left a commented line in `settings.py` (I'm using the default test runner) - let me quickly fix that and then I think it would be good to merge.

afabiani (approved):

PR #8197, opened 2021-10-04
Harvester geonode current

ricardogsilva:
This PR adds additional harvesters for being able to process current GeoNode stable deployments. More details to be added soon...


giohappy:
@ricardogsilva we can skip pulling remote maps (as wels as geostories and dashboards) for the moment. Anyway FYI resources backed my MapStore (maps, geostories and dashboards) can be rebuilt from the `blob` data that is returned by the API when the `include[]=data` parameter.

PR #8314, opened 2021-11-03
[Fixes #8312] Thumbnails filenames or URLs should always change when generetad to avoid browser caching

### LINKED ISSUE #8312: Thumbnails filenames or URLs should always change when generetad to avoid browser caching
After removing the cache busting for media files we're now facing a problem when generating new thumbnails for a resource, since the name remains the same and the browser cached the images.

The proposal is to always generate a different filename or a different `Link` url when a new thumbnail is generated.

marthamareal:
References: #8312

## Checklist

For all pull requests:


afabiani:
@marthamareal can you please remove this log line here or switch to `debug`?

image

mattiagiupponi (changes requested):
@marthamareal I don't think that this solution will work for the master branch.
The main difference between the master branch and 3.3.x versions is the `StorageManager`.

In the master branch, the storage could be remote so it can differ from the default `FileSystemStorage` provided by Django.

Let's take as an example `DropboxStorageManager`.
After the upload of a Dataset/Doc/Map, the storage manager uses `Django-storages` library to handle the storage.
In this case, the URL returned by the storage manager looks like this:
```
https://uc80c3027ff34c0a1b19e214cc23.dl.dropboxusercontent.com/cd/0/get/BZZ3o5IL0AuSkxvMMqry_5Bra6nBBca_GB9BiD6jI9jc739bLXbIpG6vi_LuQz_IgVJg5c0PuCEALiGoLZIQ_r-0jQM_SbMPRAIoMIWgNN5gfc0hOL7QUS18gZv892COmitETaW-0HvRzWJsUCeXjNUA/file
```
So the next time the thumbnail is updated the URL change, but we are not able to delete the old thumbnail since the UUID of the resource is not present in the URL. This behaviour in a long term can take all the space available in the storage.

My suggestion is to evaluate if we can store the image path by using `storage_manager.path(resource_name)` in the [ResourceBase.file](https://github.com/GeoNode/geonode/blob/6a0793510b8c59324745de41f21f4991722e0969/geonode/base/models.py#L1008), retrieve the value when is required to upload the thumb, delete the remote resource using `storage_manager.delete(filename)` and re-upload the thumbnail. (same behaviour as for the Document replace)

@afabiani @giohappy any thought on this?

afabiani:
@mattiagiupponi correct, we need to rely on the `StorageManager` for `master` branch otherwise this won't work with remote ones.
Not sure how the thumbnail path is store now on the `StorageManager`, however the approach @mattiagiupponi proposed is the correct one.

afabiani (changes requested):
@marthamareal please make sure we don't rely on `FileStytem` here but on the `StorageManager` functions.
The `thumb_path` also should probably be updated accordingly.

afabiani:
@mattiagiupponi I would not use `ResourceBase.files` since that one contains the paths of the source geospatial files. That could cause some issues when dealing with the resource download and/or replace append.

It is probably better to have another attribute on the model and modify the `*_thumbnail()` accordingly logic accordingly, which now relies on the thumbnail `Link`.

mattiagiupponi:
> @mattiagiupponi I would not use `ResourceBase.files` since that one contains the paths of the source geospatial files. That could cause some issues when dealing with the resource download and/or replace append.
> 

Technically, the `files` attribute does not contain only the geospatial files, since it also contains the Document's path (which could be a simple `png`). 
From my point of view, it is a generic `file` field, so it can contain all the file paths related to the resource, not only the geospatial ones. In order to avoid issues with the replace/append, the `file` could be changed from a `LIST` of files to a JSON with as a key (the resource type), the representation of the file list, for example:

```JSON
{
    "dataset": [
        "/tmp0350ztqs/san_andres_y_providencia_location.dbf",
        "/tmp0350ztqs/san_andres_y_providencia_location.prj",
        "/tmp0350ztqs/san_andres_y_providencia_location.shx",
        "/tmp0350ztqs/san_andres_y_providencia_location.shp"
    ],
    "thumb": [
        "/path/to/thumb.jpg"
    ]
}
```
or for Documents;

```JSON
{
    "document": [
        "/tmp0350ztqs/my_doc.pdf"
    ],
    "thumb": [
        "/path/to/doc_thumb.jpg"
    ]
}
```

> It is probably better to have another attribute on the model and modify the `*_thumbnail()` accordingly logic accordingly, which now relies on the thumbnail `Link`.

The `ResourceBase` is already a quite wide model. Creating new fields can be counterproductive in the long term

afabiani:
@mattiagiupponi true but in this case I really don't think it is worth the effort. Changing the structure how you suggest means reviewing the whole core modules `resources.managers` and `upload`. Moreover we should always check for files associated to some specific keys.

mattiagiupponi (changes requested):
@marthamareal I just tested this in a **vanilla** installation of GeoNode.

The deletion of the image is performed 2 times:
- Inside the function `get_unique_upload_path` and with the non-unique name here: https://github.com/GeoNode/geonode/blob/032a4e7ea3085d95524b57dfe2861ed85f97505d/geonode/base/models.py#L1669-L1671 
Should be better to avoid double calls on this. Be sure to have a unique ad explicit deletion (inside the function is a bit hidden)
- On dataset upload it raise also this error (probably a matter of missing config), can you give a check on it?:
```python
[Errno 2] No such file or directory: '/opt/contrib/geonode/geonode/uploaded/thumbs/dataset-9ee9986c-43d2-11ec-a960-00155d8733b5-thumb-0ee9db1b-a754-45f9-bc8a-0bd912a43409.png'
  File "/opt/contrib/geonode/geonode/base/models.py", line 1694, in save_thumbnail
    cover.save(tmp_location, format='PNG')
  File "/home/mattia/.virtualenvs/test/lib/python3.8/site-packages/PIL/Image.py", line 2237, in save
    fp = builtins.open(filename, "w+b")
FileNotFoundError: [Errno 2] No such file or directory: '/opt/contrib/geonode/geonode/uploaded/thumbs/dataset-9ee9986c-43d2-11ec-a960-00155d8733b5-thumb-0ee9db1b-a754-45f9-bc8a-0bd912a43409.png'
```
- The deletion actually is not working. The `remove_thumb` should be called with the `filename + file_extension` to let the `storage_manager.exists(path)` return `True` and then delete the old Thumbs

afabiani:
@mattiagiupponi @marthamareal I'm on it

mattiagiupponi (approved):

PR #8343, opened 2021-11-08
[Fixes #8342] Extend the Map API

### LINKED ISSUE #8342: Extend the Map API
### Rationale
The current model for `Map` and `MapLayer` is intricate, convoluted and unnecessarily complex. It builds on previous models (GPX*) and now we are managing an overly complex bridge between the MapStore config model, the Map and MapLayer models, and the GPX classes.

At the opposite we have the GeoApp model which is an abstract model which just provides a "type" information and access to the `BaseResource.blob` field which is used by MapStore to do its own things. The `blob` field (`data` as called by the API) is completely opaque to GeoNode, and this is how things should be,

On the other side the MapStore config held inside the blob field must be inspected and recreated by GeoNode and the MapStore adapter to extract / provide the additional information needed by the current Map model, and here we enter the grey area of parsing, and building the `blob` content. 

However, the MapStore client in master doesn't need this, and we don't have legacy views / templates that need the information that we're currently storing inside the `Map` and `MapLayer` (and GXP*) models, so most of the bridging performed by the MapStore adapter and the underlying GeoNode methods is probably useless.
Moreover, previous GeoNode functionalities that used this information are not exposed anymore (download of a map, exposing WMS services for a map, etc.).

### Proposal
The proposal is to simplify the Map and MapLayer models to only retain the minimal information required for the functioning of the maps catalog. Any information targeting the client / viewer functionality will be dropped. Any client related information can be stored inside the `ResourceBase.blob` field (exposed as `data` from the APIs) and clients can use for their own purposes. 

**Map API GET**
The objects even simpler than what I said in my previous comment. And this is needed only for existing maps (GET):

```
{
    <what is already returned by the `/api/v2/map/{id}?include[]=data`>,
    "maplayers": [
        {
            id: maplayer.id,
            extra_params: (seem below)
            dataset: {
                (seem below)
            },
            current_style: "workspace:name",
            styles: ["workspace:name"]
        }
    ]
}
```

- **dataset**: only a subset of fields of the Dataset model are required here, i.e. those required by the client as can be seen here https://github.com/GeoNode/geonode-mapstore-client/blob/master/geonode_mapstore_client/client/js/utils/ResourceUtils.js#L53-L64
- **extra_params**: an opaque JSONField where the client can put useful information about the maplayer. For the moment, the only extra information will be the "msid", which is set by the client to match the maplayer with the layer inside the mapconfig blob
- **current_style**: a name referencing a (Geoserver) style currently applied to this MapLayer
- **styles**: an array of names referencing (Geoserver) styles available for this MapLayer. Notice that such styles might only exist in the context of the `MapLayer` and not the underlying `Dataset`.

The "defaults" that I included before are not needed, since the client already obtain them from the context inside the [_geonode_config.html](https://github.com/GeoNode/geonode-mapstore-client/blob/master/geonode_mapstore_client/templates/geonode-mapstore-client/_geonode_config.html#L100-L112) 

Notice that at this moment, the map's maplayers can be retrieved from `/api/v2/map/{id}/datasets`. The name of the endpoint is wrong (it should be `/api/v2/map/{id}/maplayers`), but in any case it's not currently used by the client. The new MapLayer objects will be returned (and set) inlined with the map, so this endpoint can be removed in the future.

**Map API POST/PUT/PATCH**
The client is using the /api/v2/map/{id} also to save and update maps. The API currently use the MapStore serializer to read into the map config and perform the downstream serialization. Once it will be the client to provide the needed information, we can remove the [hooksets ](https://github.com/GeoNode/geonode/blob/7a8f750030a20bc9b8a62cd13c63303163ae6339/geonode/maps/api/views.py#L74-L83) and any [reference to the MapStore serializer](https://github.com/GeoNode/geonode/blob/7a8f750030a20bc9b8a62cd13c63303163ae6339/geonode/maps/api/hook.py#L107-L112) from the maps module 

On save the client will send a partial object as the one returned from the API. The following case depicts the request body for a new map (POST):

```
{
    <these fields are already set by the client when posting, putting or patching a map>
    "maplayers": [
        {
            extra_params: (seem below)
            style: (seem below)
        },
        current_style: "workspace:name",
        styles: ["workspace:name"]
    ]
}
```

In the case of PUT/PATCH the object might contain additional fields for preexisting data.

italogsfernandes:
references: https://github.com/GeoNode/geonode/issues/8342

## Extend the Map API

Adds the following info to the `maps` endpoint (`/api/v2/maps/<id>/?include[]=data`):

```json
"maplayers": [
    {
        "pk": 9999,
        "extra_params": {},
        "styles": [""],
        "current_style": "",
        "dataset": {
            "alternate": "",
            "featureinfo_custom_template": "",
            "title": "",
            "perms": [],
            "pk": 9999,
            "has_time": false,
            "default_style": {
                "pk": 9999,
                "name": "",
                "workspace": "",
                "sld_title": "",
                "sld_url": ""
            },
            "styles": [
                {
                    "pk": 9999,
                    "name": "",
                    "workspace": "",
                    "sld_title": "",
                    "sld_url": ""
                }
            ],
            "ptype": "",
            "link": "",
            "links": [
                {
                    "extension": "",
                    "link_type": "",
                    "name": "",
                    "mime": "",
                    "url": ""
                }
            ]
        }
    }
]
```

## How to test

* Create a map
* Get the info: *GET* `/api/v2/maps/<id>/?include[]=data`
* Verify the presence of the `maplayers` field and its data. 

## Checklist

For all pull requests:


afabiani:
@italogsfernandes is this ready for review?

italogsfernandes:
> @italogsfernandes is this ready for review?

Almost, I'll let you know when ready.

italogsfernandes:
@afabiani PR ready for review

afabiani (changes requested):
@italogsfernandes it looks good, however I would add to the api map tests also the write operations in order to actually demonstrate the possibility to change the values of the new attributes.

Also we should probably wrap the `set_current_style_value` into a single transaction in order to avoid inconsistent status and optimize the commits to the DB.

Another small hint, whenever we do cycles like this where we will simply iterate over a list of objects but we don't need to store them into memory, let's use the `iterator()` instead of `all()`

e.g.:

  `MapLayer.objects.iterator()`

italogsfernandes:
@afabiani I added the changes:
* [x] Api map tests for the write operations
* [x] Use of `.iterator()` instead of `.all()`

> probably wrap the set_current_style_value into a single transaction

Considering that the `styles` column is stored as a string, it is a little tricky for me to parse the string to a list and then get the first element of the list or null and, finally, store that value in `current_style` in a *single sql transaction*. If that change is needed, I'll need help with that part.

giohappy:
@afabiani the set_current_style_value method is executed for each layer inside the migration.  A migration is atomic by default, unless you explicitely set it differently.

I'm not sure what yoy by saying to wrap the calla in a single transaction.

afabiani:
@giohappy sure, my fault, migrations are atomic of course unless explicitly declared.

afabiani (approved):

PR #8395, opened 2021-11-15
[Fixes #8394] Move Maps API responsibilities from geonode-mapstore-cl…

### LINKED ISSUE #8394: Move Maps API responsibilities from  geonode-mapstore-client to GeoNode
# Move Maps API responsibilities from  geonode-mapstore-client to GeoNode

## Overview

* To create/update a map by API, the geonode api calls the mapstore-client [perform_create](https://github.com/GeoNode/geonode-mapstore-client/blob/master/mapstore2_adapter/plugins/serializers.py#L228) and [perform_update](https://github.com/GeoNode/geonode-mapstore-client/blob/master/mapstore2_adapter/plugins/serializers.py#L243), so the client handles the creation and calls different parts of geonode in order to do it.
* They are highly coupled (in a complicated mix of responsibilities).
* Because of that, some legacy code from `GPX*` classes and others are still being called, blocking the drop of this code.

italogsfernandes:
…ient to GeoNode
references: #8394

## What was done

* The create and update action of the Maps API were modified. Instead of using the hookset they do:
    * Saves the values directly to the database, almost ignoring the `blob` data.
    * After saving, registers an event, call the resource manager and sets the thumbnail.
    * After changing, the same, but also takes cares of the `map_changed_signal`.
    * Note: There were a lot of code getting data from the blob, doing some processing and changing the way it was stored. After the changes written here, the changes outside the blob should be explicit send by the client. 
* minor changes:
    * `maps/utils.py` changed to `maps/utils/`
    * `maps/contants.py` created, so the string constants can be easily imported by different files.
        * string constants moves from `maps/views.py` to the new file. 

## Checklist

For all pull requests:


geonode/maps/api/views.py
```diff
@@ -86,12 +92,122 @@ def local_datasets(self, request, pk=None):
...
+        # Initialize maplayers serializer
+        maplayers_data = serializer.validated_data.pop("maplayers")
+        maplayers_serializer = MapLayerSerializer(data=maplayers_data, many=True)
+        maplayers_serializer.is_valid(raise_exception=True)
+
+        # Create new maplayers, map relation will be added by serializer.save()
+        maplayers = maplayers_serializer.save()
+        return maplayers
+
+    def _get_maplayers_data_from_blob(self, blob_map_data: dict):
```
giohappy, line 156: @italogsfernandes why are you extracting this information from the blob? We said we don't want GeoNode to put its nose into the blob. And, again, I don't see why we need to extract this data. Where are they required?

geonode/maps/api/serializers.py
```diff
@@ -76,15 +80,24 @@ class Meta:
             "current_style",
             "styles",
             "dataset",
+            "stack_order",
```
giohappy, line 83: Why are you setting the following fields? This information is inside the blob, and is only required by the client. Why do we need to be aware of these attributes?

giohappy (changes requested):

geonode/maps/api/views.py
italogsfernandes: @giohappy The blob operations are inside `_get_data_from_blob` and `_get_maplayers_data_from_blob`.
The current version of Mapstore does not send the the `maplayers` data outside the blob. Without this information the relation between Maps and Layers won't be created at geonode.

This needs to change when mapstore will be up to date and that part is here, for the moment, exists only to ensure that the maplayers relation will be created when the `maplayers` field is not provided bu the POST/PATCH.

geonode/maps/api/serializers.py
italogsfernandes: Changed

geonode/maps/api/views.py
giohappy: @italogsfernandes I'd prefer to remove this from the beginning. GeoNode currently doesn't rely on this information, since we're not showing the map / maplayer / dataset relationships now. If we keep them we risk to hide potential issues.

italogsfernandes: So, it's removed. I started the `test_suite` to see if it will broke something not seem by us yet.

giohappy (approved):

PR #8501, opened 2021-12-16
Update django.po for el

kalxas:
Minor fix in Greek translation


afabiani:
@kalxas could you send the ".mo" too?

kalxas:
Thanks @afabiani I will reproduce and send again :)

PR #8512, opened 2021-12-17
[Fixes #8346] Implement upload size limiting mechanism for datasets

### LINKED ISSUE #8346: Implement upload size limiting mechanism for datasets
GeoNode currently doesn't provide an explicit and controlled mechanism to limit the upload size for datasets (and documents).
In the past custom configurations (mainly at the level of Nginx) have been adopted to enforce such a limit, but suche solutions lacl an integration with the client (leading to a poor UX) and require to set fixed configurations that cannot be easily changed at runtime.

In view of introducing additional controls over upload funtionality (e.g. rate limiting, quota limits, etc.) we want, as a first step, introduce a size limiting mechanism inside GeoNode core. 

It will be controlled by a new global configuration (to be placed inside GeoNode settings module). A numerical value will set the file size limit for the fiels that a user can upload for a single session.

The implementation will:

- provide the means to let frontend clients cooperate (expose the limits to the client through configurations). For example, client might block the upload operation from the beginning by inspecting files size
- integrate with the Upload API to support the required messaging between the frontend and the upload backend (e.g. for reporting limits, errors, etc.)
- provide an internal API that can be used by GeoNode modules to query for the configured limits
- offer a high level interface to be integrated inside the Upload module (APIs / views), with an underlying concrete implementation that could be adapted for master (the target of this proposal) and 3.3.x (optional backporting). 

An analysis of the interactions between the Upload module and the Resource Manager +  Storage Manager services will be required to define the right placement and architecture of this size limiting component.

italogsfernandes:

## What was done
* During the form field validation, it will retrieve the max_size data from the `UploadSizeLimit` model on the database.
* The values stored on `UploadSizeLimit` can be modified and read using an api `api/v2/upload-size-limits` or the django-admin interface.
* Update `django.po` with new texts.
* Implement custom `FILE_UPLOAD_HANDLER`.
    * [FILE_UPLOAD_HANDLERS](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-FILE_UPLOAD_HANDLERS)
    * [custom-upload-handlers](https://docs.djangoproject.com/en/3.2/ref/files/uploads/#custom-upload-handlers)

## How to test

* Setting up the size limits:
     * Using django Admin: 
          * Go to `admin`-> `Upload size limits`: http://localhost:8000/en/admin/upload/uploadsizelimit/ 
          * Set a file limit to a lower max_size value. for example (1 byte).
     * Using the API: (remember to replace `http://localhost:8000` by your base_url, and `Authorization: Basic YWRtaW46YWRtaW4=` to your auth token (here I'm using admin:admin).
         * Setting a new limit: 
        ```cURL
        curl --location --request PATCH 'http://localhost:8000/api/v2/upload-size-limits/base_file' \
        --header 'Accept: application/json' \
        --header 'Authorization: Basic YWRtaW46YWRtaW4=' \
        --data-raw '{
            "max_size": 1024
        }'
        ```
        * Checking the list of limits using the API:  
         ```cURL
        curl --location --request GET 'http://localhost:8000/api/v2/upload-size-limits/' \
        --header 'Accept: application/json' \
        --header 'Authorization: Basic YWRtaW46YWRtaW4='
        ```

* Now, you can try to upload a file with a size greater than the limit: http://localhost:8000/upload/
* You will get an error message, similar to the one here below:
image

## Additional Resources

Upload Size Limit Api - postman collection: [Size Limit Api.postman_collection.zip](https://github.com/GeoNode/geonode/files/7737957/Size.Limit.Api.postman_collection.zip)

## Checklist

For all pull requests:

- [x] Commits adding **new texts* do use gettext and have updated .po / .mo files (without location infos)


italogsfernandes:
Converted to draft while CI tests are running...

italogsfernandes:
Converted to draft again, changes requested by reviewer.

## ToDo

* Add a check for sum of file sizes.
    * Instead of validating each file (`.shp`, `.dbf`, `.prj`, `.shx`...), only the sum of the sizes will be validated.
    * Obs: `.dbf` and `.shp` file are the biggest in general (for shape files).  geotiff and geopkg are saved in `base_file`.

giohappy:
@italogsfernandes what's the benefit of adopting FILE_UPLOAD_HANDLERS or a custom upload handler, over the solution that we have now? I mean, beyond having a more "djangoist" solution, do you see relevant limitations to the current solution?

italogsfernandes:
> @italogsfernandes what's the benefit of adopting FILE_UPLOAD_HANDLERS or a custom upload handler, over the solution that we have now? I mean, beyond having a more "djangoist" solution, do you see relevant limitations to the current solution?

@giohappy The solution we have now will still be needed. The `FILE_UPLOAD_HANDLER` is called during the parsing of the `request` object and is responsible for copying the uploaded file to a temporary location.
If we just interrupt this upload process without validating the form (solution we have here in this PR), the connection will be lost and the user will not receive a descriptive error message.

The main advantage is that we can avoid sending large files to the server before they arrive in the server's temp folder. See the following case, for example: 
* Upload size limit is 100 MB.
* The user tries to upload 1 GB of files.
* What happens now:
    * Each file will be copied [chunk by chunk](https://github.com/django/django/blob/main/django/http/multipartparser.py#L240:L267) (64 kB) to a [temp file](https://github.com/django/django/blob/main/django/core/files/uploadhandler.py#L149:L150).
    * Later, [our form validation](https://github.com/italogsfernandes/geonode/blob/ISSUE_8346/geonode/upload/forms.py#L95-L101) will verify the file size and return an error message.
* What can be changed/improved:
    * Before the file upload to the server, we can verify the size of the request content.
    * If it is larger than 500 MB (for example) we can skip the creation of temporary files, so instead of copying the file to the server we go directly to form validation with the error message. 
    * The [loop to upload the file](https://github.com/django/django/blob/main/django/http/multipartparser.py#L240) won't be necessary, instead we can create an empty [UploadedFile](https://github.com/django/django/blob/main/django/core/files/uploadedfile.py#L17) obj with its size property set to 1 GB.
    * The request will follow its way until the form validation, where the error will be thrown: "Total upload size exceeds 100 MB."

giohappy:
@italogsfernandes how long would it take to add this imprevement?

italogsfernandes:
I think it would take from 1 day and half to 2 days.

giohappy:
ok @italogsfernandes let's do it within the same PR.

italogsfernandes:
> ok @italogsfernandes let's do it within the same PR.

It's done

geonode/upload/uploadhandler.py
```diff
@@ -0,0 +1,215 @@
...
+                if file_obj:
+                    # If it returns a file object, then set the files dict.
+                    _files.appendlist(force_str(old_field_name, encoding, errors="replace"), file_obj)
+                old_field_name = None
+
+            try:
+                disposition = meta_data["content-disposition"][1]
+                field_name = disposition["name"].strip()
+            except (KeyError, IndexError, AttributeError):
+                continue
```
mattiagiupponi, line 87: can we log the exception?

geonode/upload/uploadhandler.py
```diff
@@ -0,0 +1,215 @@
...
+        file_name = html.unescape(file_name)
+        file_name = file_name.rsplit("/")[-1]
+        file_name = file_name.rsplit("\\")[-1]
+
+        if file_name in {"", ".", ".."}:
+            return None
+        return file_name
+
+    def _get_max_size(self):
+        max_size_db_obj, created = UploadSizeLimit.objects.get_or_create(
```
mattiagiupponi, line 176: From my point of view, will be better to create by default the UploadSizeLimit with the default value during the migration phase.
Also here we are hiding a possible object creation inside a get method which could not be a good pattern.
A possible solution could be using the exists method of the queryset and in the case of Not exists object:

1. raise an exception
2. call creation method (maybe a method of the model like UploadSizeLimit.objects.create_default_limit())

mattiagiupponi (changes requested):

mattiagiupponi:
@italogsfernandes the issues say that this functionality should be available for datasets and documents.

> GeoNode currently doesn't provide an explicit and controlled mechanism to limit the upload size for datasets (and **documents**).

I tried to upload a document with the following settings:
file_upload_handler -> 1
total_upload_size_sum -> 1
The size of my image is `133,4 kB (133.375 bytes)`, I'm expecting to have an error, but the upload goes smoothly, can you give a check please?

geonode/upload/forms.py
```diff
@@ -86,6 +95,24 @@ def clean(self):
         cleaned["valid_extensions"] = valid_extensions
         return cleaned
 
+    def validate_files_sum_of_sizes(self):
+        max_size = self._get_uploads_max_size()
+        total_size = sum([django_file.size for django_file in self.files.values()])
```
mattiagiupponi, line 100: @italogsfernandes this could raise some issues with zip files.
I was trying to upload a dataset, but something is not as expected, the Multidict value has X keyword as are the input files.
For example, the .zip file will have these two keywords:
```
'zip_file':<SimpleUploadedFile: ProvCM01012019_g.zip (application/zip)>
'base_file':<SimpleUploadedFile: ProvCM01012019_g.zip (application/zip)>
```
so with the some, the total size (assuming that the file is 1 byte) will be 2 bytes, while the file is only 1 byte.
My suggestion here could be ignore the "base_file" in the size sum in order to have a real and valid sum

geonode/base/api/permissions.py
```diff
@@ -136,6 +136,26 @@ def has_object_permission(self, request, view, obj):
         return IsOwnerOrAdmin.has_object_permission(self, request, view, obj)
 
 
+class IsAdminOrReadOnly(permissions.BasePermission):
```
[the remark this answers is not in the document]
italogsfernandes: Using `IsSelfOrAdminOrReadOnly` creates a security breach because it would be possible for users to edit objects if the object `id` is the same of the user id, for some coincidence.
https://github.com/GeoNode/geonode/blob/master/geonode/base/api/permissions.py#L46

geonode/upload/uploadhandler.py
italogsfernandes: Usually this is not logged by the default django managers. We can log, but should we log in this case?

https://github.com/django/django/blob/main/django/http/multipartparser.py#L166-L169

geonode/base/api/permissions.py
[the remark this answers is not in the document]
giohappy: Rather then creating additional combinations IMHO the IsSelf method should be fixed to enforce a check on the obj model before doing the comparison.

geonode/settings.py
```diff
@@ -2067,3 +2067,10 @@ def get_geonode_catalogue_service():
     "importlayers",
     "set_all_datasets_metadata",
 ] + ast.literal_eval(os.getenv('MANAGEMENT_COMMANDS_EXPOSED_OVER_HTTP ', '[]')))
+
+
+FILE_UPLOAD_HANDLERS = [
+    'geonode.upload.uploadhandler.SizeRestrictedFileUploadHandler',
```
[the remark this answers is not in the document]
italogsfernandes: Issue created: https://github.com/GeoNode/documentation/issues/209
A PR with the documentation updates will be opened soon.

geonode/base/api/permissions.py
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/upload/api/tests.py
```diff
@@ -478,3 +480,242 @@ def test_rest_uploads(self):
...
+
+    def test_list_size_limits(self):
+        url = reverse('upload-size-limits-list')
+
+        # List as an admin user
+        self.client.force_authenticate(user=self.admin)
+        response = self.client.get(url)
+
+        # List as an Anonymous user
+        self.client.force_authenticate(user=None)
```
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/upload/migrations/0034_uploadsizelimit.py
```diff
@@ -0,0 +1,51 @@
+# Generated by Django 3.2.7 on 2021-12-16 16:36
+
+import django.core.validators
+from django.db import migrations, models
+
+DEFAULT_MAX_UPLOAD_SIZE = 104857600  # 100 MB
```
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/upload/models.py
```diff
@@ -36,6 +40,8 @@
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_MAX_UPLOAD_SIZE = 104857600  # 100 MB
```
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/upload/uploadhandler.py
```diff
@@ -0,0 +1,215 @@
...
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.core.files.uploadhandler import FileUploadHandler
+from django.http import QueryDict
+from django.http.multipartparser import FIELD, FILE, ChunkIter, LazyStream, Parser, exhaust
+from django.utils.datastructures import MultiValueDict
+from django.utils.encoding import force_str
+
+from geonode.upload.models import UploadSizeLimit
+
+DEFAULT_MAX_BEFORE_UPLOAD_SIZE = 209715200  # 200 MB
```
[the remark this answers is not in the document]
italogsfernandes: Updated ✅

geonode/upload/uploadhandler.py
italogsfernandes: Updated ✅ link: https://github.com/GeoNode/geonode/pull/8512/files#diff-7d1f963cfdf3df10fcd4c9edd023d0e095745fff6477219eba74ff40889185b2R173-R178

geonode/upload/forms.py
giohappy: @italogsfernandes what about this?

italogsfernandes: To do:
* [ ] Apply validation to `documents` upload page and add tests.
* [ ] Ignore `zip_file`, (`base_file` is used by other layer formats).

geonode/upload/forms.py
```diff
@@ -86,11 +95,36 @@ def clean(self):
...
+
     def _get_uploaded_files(self):
         """Return a list with all of the uploaded files"""
         return [django_file for field_name, django_file in self.files.items()
                 if field_name != "base_file"]
 
+    def _get_uploaded_files_total_size(self):
+        """Return a list with all of the uploaded files"""
+        excluded_files = ("zip_file", )
+        uploaded_files_sizes = [
```
mattiagiupponi, line 121: @italogsfernandes the issue looks resolved for the zip files, but is still present for the scenario where u want to upload the files without a zip (providing shp, shx, etc).
The base_file key is present in the same file (usually) as the shp key, so in this case, the total sum of the files is not correct.
Those are the available files in the `self.files.items()`, as u can see the shp file is duplicated:
```python
0:<SimpleUploadedFile: ProvCM01012019_g_WGS84.shp (application/x-qgis)>
1:<SimpleUploadedFile: ProvCM01012019_g_WGS84.dbf (application/x-dbf)>
2:<SimpleUploadedFile: ProvCM01012019_g_WGS84.prj (application/octet-stream)>
3:<SimpleUploadedFile: ProvCM01012019_g_WGS84.shp (application/x-qgis)>
4:<SimpleUploadedFile: ProvCM01012019_g_WGS84.shx (application/x-qgis)>
```
Can we add few test on this part?

geonode/upload/forms.py
mattiagiupponi: @italogsfernandes i tried to upload again a document, but no exception is raised for the limit size (I set 1 byte with an image of 133.375 bytes) From UI it doesn't raise any exception, but it upload an empty resource and log this warn in console:

`Document has no field named 'name' - {'name': 'space.jpg', 'subtype': 'image', 'doc_url': None, 'extension': 'jpg', 'abstract': 'No abstract provided', 'date': datetime.datetime(2022, 1, 13, 9, 1, 7, 259632, tzinfo=<UTC>), 'date_type': 'Creation', 'bbox_polygon': None}`

can you give another check on it, please?

italogsfernandes: It seems that your Upload size limit with slug=`document_upload_size` was set to the default value (100 MB) and your `file_upload_handler` size limit set to some small value.

The in the description of the `file_upload_handler` it says that the value should be greater than the other limits, but it may not be enough to prevent this kind of misconfiguration.

I gonna add a couple of validation methods to prevent this.

mattiagiupponi (dismissed):

giohappy:
please @italogsfernandes double check the failing tests. I suspect they're are related to upload limitinh.

afabiani (approved):

PR #8581, opened 2022-01-10
[Fixes #8561] Refactor the API for the update of the thumbnails

### LINKED ISSUE #8561: Refactor the API for the update of the thumbnails
## Expected Behavior

Thumbnails for all resources should be saved correctly.

## Actual Behavior

In this moment the update of thumbnails from the client work "by accident". Except for maps, the base64 string sent from the client is saved as it is, along the rest of the resource data, when a PATCH request is submitted. When the client loads the resource it's capable to manage the base64 string returned inside the `thumbnail_url` property and show it. 
By the way this is not the way things should work. The thumbnail should always be saved as an image and the `thumbnail_url` should always contain a staic url for the image.

The fix requires the following changes:

- [ ] The ResorceManager.set_thumbnail method must be extended to 
  - [ ] accept a file-like object representing the thumbnail 
  - [ ] call any specific intermediate method according to the resoruce type. In the case of documents the final `document.save_thumbnail` should be removed
  - [ ] call the `resource.save_thumbnail` method in the end
- [ ] A new set_thumbnail API must be exposed under the resources/{pk} endpoint. This endpoint will validate the POST data, perform any base64 decoding ( we can reuse some logic from [handle_map_thumbnail](https://github.com/GeoNode/geonode/blob/28357a5cdf0d93657033cdd78537567c98b39854/geonode/maps/utils/thumbnail.py#L42)) and, finally, call `ResourceManager.set_thumbnail` with a file-like parameter.
- [ ] Remove [these two lines](https://github.com/GeoNode/geonode/blob/28357a5cdf0d93657033cdd78537567c98b39854/geonode/maps/api/views.py#L166-L168) from the Maps API view

When this will be ready we will:
 - make any resource type API validate the content passed inside `thumbnail_url` (POST, PUT, PATCH) and **only accept URLs**. 
 - The MapStore client call the set_thumbnail API explicitly and stop patching the thumbnail_url field

  1. Go to [master](https://master.demo.geonode.org/)
  2. Try uploading thumbnail for document, dataset or geoapp
  
hnh

  3. Check the url of the thumbnail saved.

## Specifications

  - GeoNode version: master
  - Platform:
  - Additional details:

marthamareal:

## Checklist

For all pull requests:


giohappy:
@marthamareal from our call:

 - the `set_thumbnail` endpoint should also accept (PUT) file uploads. DRF [FileUploadParser](https://www.django-rest-framework.org/api-guide/parsers/#fileuploadparser) can be used for this. In the future I expect this endpont will replace the `thumbnail_upload` method that is still used by the metadata wizard thumbnail widget (for documents and geoapps)
 - make the validation more robust, instead of relying on exceptions. 
 - in case no thumbnail is passed in the request the method will return an error

giohappy (changes requested):
As reviewed in our live call the current PR has the following problems:

- you're using raw binary data for file upload, which is not what a web client usually does. A form data upload is the usual way, so the FileUpload parser must be replaced wtih the MultiPartParser
- relying on the try / catch exception is not robust. In several cases the based64 decode might try to decode a URL string, so a pattern matching must be done before passing the string to the base64 decoder
- even if a thumbnail is set an existing CuratedThumbnail takes precedence inside any API response. Since we're not using curated thumbnails anymore, we will remove any check inside the API, and we will directly return the `resource.thumbnail_url` . The next step will be (not in this PR) drop CuratedThumbnail and fix the metadata wizard widget accordingly

All these fixes has been committed to a [temporary branch](https://github.com/GeoNode/geonode/tree/marthamareal-issue-8561). Pull and merge them into your own private branch, then we will remove this temp branch.

What is left to you is fixing the tests (use form data upload instead of raw binary) and execute pep validation.

afabiani:
@giohappy @marthamareal are we still using the REST endpoint `set_thumbnail_from_bbox` ? Could we maybe get rid of it?

giohappy:
> @giohappy @marthamareal are we still using the REST endpoint set_thumbnail_from_bbox ? Could we maybe get rid of it?

@afabiani for the moment this new endpoint only covers the image thumbnails, and it will replace the legacy `thumbnail_upload` view that is currently used by the metadata wizard widget.

However I would be in favour of merging the two API v2 endpoints in a new PR.

giohappy (changes requested):
@marthamareal

 - there's one [test that is failing](https://app.circleci.com/pipelines/github/GeoNode/geonode/6385/workflows/1a9e1d09-7088-4a42-b585-5dafadd36677/jobs/24082/parallel-runs/0/steps/0-106)
 - let's make the ResourceBase.thumbnail_url [API field read-only](https://www.django-rest-framework.org/api-guide/fields/#read_only). I mean from the point of view of the API not the model itself, of course

marthamareal:
@giohappy I have updated this.

PR #8731, opened 2022-02-02
[Fixes #8667] Implement the upload of files from remote locations

### LINKED ISSUE #8667: Implement the upload of files from remote locations
The current implementation of the `upload` view, and in particular the [save_step_view](https://github.com/GeoNode/geonode/blob/1c4278ac12cdc7aae9226ab2d2b92586e0173c1c/geonode/upload/views.py#L133), only supports files uploaded as multipart data through `LayerUploadForm`.

We want to extend it to accept paths to a remote location, from which the file can be read and uploaded to Geoserver.

The high level requirements are:

 - for the moment we only implement support to filesystem paths (that of course are visible to GeoNode), but we want to make it generic to support additional implementations for other types of remote storages (S3, etc.). [smart_open](https://github.com/RaRe-Technologies/smart_open) could be a nice solution to be considered.
 - we don't want to make local copies of the remote file, including temporary copies.
 - files will be [uploaded directly through the gs_uploader instance](https://github.com/GeoNode/geonode/blob/1c4278ac12cdc7aae9226ab2d2b92586e0173c1c/geonode/upload/upload.py#L358) . I guess that the file interface should make it mostly transparent, without any modification.

GeoNode doesn't rely much on local copies of uploaded files. In 3.3.x, if available, a direct link for their download is provided, but it's not a prerequisite for the functionality of GeoNode. However we have to check all the steps in the upload procedure, in case in some place they expect to have the files copied to local locations. 

Notice that in this moment the support for mosaic granules is not well maintained (a new issue will be created for this), so this must be taken into account in implementing and testing this requirements.

This task targets both master and 3.3.x. The upload that is affected should still be the same (or very similar) on both branches. The big differences happen at the end of the import, where the GeoNode resource is finalized and created (master uses the Resource Manager, which doesn't exist in 3.3.x).

italogsfernandes:
references: #8667 

## What was done
* The `smart_open` library was added.
* DataRetriever and DataRetrieverGroup component were created.
* Added fields `*_path` to the UploadForm, now it's possible to add a file sending it (`base_file`) or by its path (`base_file_path`).
* The `*_path` field is handled by [smart_open](https://github.com/RaRe-Technologies/smart_open), so it can accept http, s3, server local files etc...
* Files are now transferred to a tmp folder during upload.
     * Even if the file is local, this is needed. Some processes only works if the uploaded files are the only files in the refereed folder.
* Upload process updated to make use of these components.

## How to test
* Request:
```
curl --location --request POST 'http://localhost:8000/api/v2/uploads/upload/' \
--header 'Authorization: Basic YWRtaW46YWRtaW4=' \
--header 'Cookie: csrftoken=V7g4BJKNmgqRFltoM5GizihCa9uV2eWD4g213CJQPDK1lROupIE44MArEWhtyLvv; sessionid=x20mqic58gc0oct5mkko30v1iwi8l294' \
--form 'permissions="{ \"users\": {\"AnonymousUser\": [\"view_resourcebase\"]} , \"groups\":{}}"' \
--form 'time="false"' \
--form 'charset="UTF-8"' \
--form 'base_file_path="https://github.com/GeoNode/gisdata/raw/master/gisdata/data/good/vector/san_andres_y_providencia_administrative.shp"' \
--form 'dbf_file_path="https://github.com/GeoNode/gisdata/raw/master/gisdata/data/good/vector/san_andres_y_providencia_administrative.dbf"' \
--form 'prj_file_path="https://github.com/GeoNode/gisdata/raw/master/gisdata/data/good/vector/san_andres_y_providencia_administrative.prj"' \
--form 'shp_file_path="https://github.com/GeoNode/gisdata/raw/master/gisdata/data/good/vector/san_andres_y_providencia_administrative.shp"' \
--form 'shx_file_path="https://github.com/GeoNode/gisdata/raw/master/gisdata/data/good/vector/san_andres_y_providencia_administrative.shx"'
```
* Expected response:
```json
{
   "bbox" : "-81.9087270,12.1189965,-79.9333322,15.5027746",
   "crs" : {
      "properties" : "EPSG:4326",
      "type" : "name"
   },
   "id" : 86,
   "status" : "finished",
   "success" : true,
   "url" : "/catalogue/#/dataset/84"
}
```

## Checklist

For all pull requests:

- [x] Commits adding **new texts** do use gettext and have updated .po / .mo files (without location infos)

### DIFF
--- a/geonode/resource/data_retriever.py
+++ b/geonode/resource/data_retriever.py
@@ -1,0 +1,144 @@
+import io
+import os
+import shutil
+import smart_open
+import tempfile
+
+from django.conf import settings
+from django.core.files.uploadedfile import UploadedFile
+
+
+BUFFER_CHUNK_SIZE = 64 * 1024
+
+
+def file_chunks_iterable(file, chunk_size=None):
+    """
+    Read the file and yield chunks of ``chunk_size`` bytes (defaults to
+    ``BUFFER_CHUNK_SIZE``).
+    """
+    chunk_size = chunk_size or BUFFER_CHUNK_SIZE
+    try:
+        file.seek(0)
+    except (AttributeError, io.UnsupportedOperation):
+        pass
+
+    while True:
+        data = file.read(chunk_size)
+        if not data:
+            break
+        yield data
+
+
+class DataRetriever(object):
+    def __init__(self, file):
+        self.temporary_folder = None
+        self.file_path = None
+
+        self._is_django_form_file = False
+        self._django_form_file = None
+        self._original_file_uri = None
+        self._smart_open_uri = None
+
+        if isinstance(file, UploadedFile):
+            # File provided by django form
+            self._django_form_file = file
+            self._is_django_form_file = True
+        elif isinstance(file, str):
+            self._is_django_form_file = False
+            self._original_file_uri = file
+            self._smart_open_uri = smart_open.parse_uri(file)
+        else:
+            raise ValueError()
+
+    def delete_temporary_file(self):
+        if (self.temporary_folder and
+            self.file_path and
+            os.path.exists(self.temporary_folder) and
+            os.path.isfile(self.file_path)
+        ):
+            # Remove File
+            os.remove(self.file_path)
+            # Verify and remove temp folder
+            folder_is_empty = len(os.listdir(self.temporary_folder)) == 0
+            folder_is_not_static_root = settings.STATIC_ROOT != os.path.dirname(os.path.abspath(self.temporary_folder))
+            if folder_is_empty and folder_is_not_static_root:
+                os.rmdir(self.temporary_folder)
+
+        self.temporary_folder = None
+        self.file_path = None
+
+    def get_path(self, allow_transfer=True):
+        if not self.file_path:
+            if allow_transfer:
+                self.transfer_remote_file()
+            else:
+                raise ValueError()
+        return self.file_path
+
+    def transfer_remote_file(self, temporary_folder=None):
+        self.temporary_folder = temporary_folder or tempfile.mkdtemp(dir=settings.STATIC_ROOT)
+        self.file_path = os.path.join(self.temporary_folder, self.name)
+        if self._is_django_form_file:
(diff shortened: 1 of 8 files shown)


giohappy:
@italogsfernandes could we rename the classes as the following?

```
DataRetriever -> DataItemRetriever
DataRetrieverGroups -> DataRetriever
```

italogsfernandes:
@giohappy It's updated

giohappy:
@afabiani as agreed:

 - the `DataRetriever` shouldn't care if we want to permanently store or not the files. It's something that only the Upload Session should care
 - we don't want to decide wether to store or not the files on the base of the files being uploaded or not. We might want to store the files even when they're passed by reference. So I would change the `spatial_files_uploaded` boolean to something like `store_files`
 - If this property is false we will completely slip the step per we save the files

afabiani:
@giohappy done

afabiani (approved):

giohappy (commented):
@afabiani I think we're not on the same page yet:

1 - You're using the `uploaded` value returned from the form to set the `UploaderSession.spatial_files_uploaded` field which, AFAICS, is only meant to store the informations if the files were uploaded or not
2 - You decide wether to store or not the files based on` Upload.store_spatial_files`, but I don't see where this field is set
3 - IMHO the Layer Form should accept an additional field `store_spatial_files`, to let the caller decide if the fiels (even the remote ones) should be stored or not. As we said yesterday this decision **shouldn't** depend inplicitely on if we pass files or paths to the for. This should be explict. The store_spatial_files might be True by default, and in that case also remote files would be stored permanently.
`UploaderSession.spatial_files_uploaded` and  `Upload.store_spatial_files` should depend on this explicit parameter set by the caller.
4 -  Do we really need the `UploaderSession.spatial_files_uploaded` field?

afabiani:
@giohappy I've added the filed "store_spatial_files" to the "LayerUploadForm"

giohappy:
@italogsfernandes this must be backported to 3.3.x

PR #8770, opened 2022-02-08
[Fixes #8737] Errors in devcontainer and docker

### LINKED ISSUE #8737: Errors in devcontainer

  1.
  2.
  3.

## Specifications

  - GeoNode version:
  - Platform:
  - Additional details:

italogsfernandes:
references #8737 

## What was done
* Some errors in the devcontainer and development docker-compose were fixed.
* Connection with geoserver is working.
* Nginx service is up again (it was not running).
* Updates in the .env file.
* Updates in the entrypoint: The `development` option was not working. 

## How to use
* Documentation:  https://docs.geonode.org/en/master/devel/docker/index.html

## Checklist

For all pull requests:


italogsfernandes:
As discussed with @mattiagiupponi, changed to draft while we find a better solution to the geoserver access.
* [ ] Set geoserver proxy url as `localhost:8080/geoserver` instead of `geoserver:8080/geoserver`
* [ ] Change the related .env variables.
* [ ] Update the docker setup to update the geoserver automatically, using a custom Docker image inherited from geonode/geoserver:2.19.
* [ ] Use paver to setup the data when initializing, instead of manually migrating and loading the fixture data.
* [ ] Setup docker to make possible to the owslib to access `localhost:8080` in the get map method and others (https://github.com/geopython/OWSLib/blob/master/owslib/map/wms111.py#L70).

giohappy:
@italogsfernandes why do we need to run the migrate and loaddata commands manually? Aren't they [run automatically by the entrypoint.sh](https://github.com/GeoNode/geonode/blob/master/entrypoint.sh#L53-L55)

italogsfernandes:
> @italogsfernandes why do we need to run the migrate and loaddata commands manually? Aren't they [run automatically by the entrypoint.sh](https://github.com/GeoNode/geonode/blob/master/entrypoint.sh#L53-L55)

Yes, this should happen automatically, but for some reason it's not.
There are also some issues with the communication between django and geoserver.
After looking for the reason why it's not working, I didn't find any. I'll ask @afabiani for help.

italogsfernandes:
Ready for review

afabiani (approved):

italogsfernandes:
> @italogsfernandes I guess this one should be backported to `3.3.x` also.

ok

mattiagiupponi:
Is the error on the thumbnail fixed?

afabiani:
@mattiagiupponi @italogsfernandes about the thumbnail issue the problem is that probably those URLs are not reachable from the container.

```
GEOSERVER_WEB_UI_LOCATION=http://localhost/geoserver/
GEOSERVER_PUBLIC_LOCATION=http://localhost/geoserver/
```

Let's maybe try to set those to

```
GEOSERVER_WEB_UI_LOCATION=http://geoserver:8080/geoserver/
GEOSERVER_PUBLIC_LOCATION=http://geoserver:8080/geoserver/
```

If still not working there's probably no much we can do except maybe try to pass through the GeoNode internal proxy

```
GEOSERVER_WEB_UI_LOCATION=http://localhost:8000/gs/
GEOSERVER_PUBLIC_LOCATION=http://localhost:8000/gs/
```

italogsfernandes:
@afabiani changing the .env variable is not working. It calls the geoserver service with no problems for the first call, but geoserver returns the proxy_url (localhost:80/geoserver) unrecheable from inside the django container. The following calls of the owslib makes uses of this url returned by geoserver.

https://github.com/geopython/OWSLib/blob/master/owslib/map/wms111.py#L70

giohappy:
@afabiani @italogsfernandes regarding Geoserver URLs I don't understand why they're so problematic in this context, given that we deploy GeoNode in production with no issue in this regard. What's so different inside a dev container context?

afabiani:
@giohappy `localhost` is different for every container, so, unless you do not use IP or names it is not possible to resolve the machines.

giohappy:
@afabiani but this is true even for the usual prduction deployments we do. My question is why this part is so different for a dev setup?

afabiani:
@giohappy in a production environment cannot be like this of course, however, after some more testing, if we change the GeoServer `proxy base url` to `http://geoserver:8080/geoserver` the `django` container will be able to reach the GeoServer one and render the thumbnail.

I also noticed that the documentation page is incomplete, the `oauth2` configuration must be also changed in order to make GeoServer and GeoNode communicate correctly.

I will push some updates shortly.

giohappy:
@afabiani I trust you but I still cannot understand why the public base urls are taking the docker hosts in this context instead of "http://localhost" (as we have in production, of course with a real host in that case).

Anyway, the important thing is to have all this documented if manual changes are strictly required...

afabiani:
@giohappy the `localhost` case is a special case. Unfortunately `localhost` is something different for any container, not sure if you see my point.

afabiani:
@giohappy nope, let me provide you an example:

 - let's say GeoServer declares it is exposed on `http://localhost/geoserver` via its `proxyBaseUrl`
 - the `django` container then tries to create a thumbnail by making queries to the `WMS`
 - any request will fail because `http://localhost/geoserver/wms` it wont be correctly resolved from inside the `django` container

PR #8777, opened 2022-02-09
Correctly assign a user when creating a new upload session.

alistairjmcintyre:
When creating a new `UploadSession` via `get_or_create`, the database has a constraint that a user must be specified. Thus, geonode generates an error when trying to create one with only the layer details specified.

The `default` dictionary on `get_or_create` allows the addition of variables for creation with modifying the search query for existing `UploadSession` objects.

## Checklist

For all pull requests:

### DIFF
--- a/geonode/layers/utils.py
+++ b/geonode/layers/utils.py
@@ -454,11 +454,11 @@ def file_upload(filename,
     # Create a new upload session
     if layer:
         latest_uploads = UploadSession.objects.filter(resource=layer).order_by('-date')
-        if latest_uploads.count() > 1:
+        if len(latest_uploads) >= 1:
             upload_session = latest_uploads.first()
+            upload_session.user = theuser
         else:
-            upload_session, _created = UploadSession.objects.get_or_create(resource=layer)
-        upload_session.user = theuser
+            upload_session = UploadSession.objects.create(resource=layer, user=theuser)
         upload_session.layerfile_set.all().delete()
     else:
         upload_session = UploadSession.objects.create(user=theuser)


geonode/layers/utils.py
```diff
@@ -457,7 +457,7 @@ def file_upload(filename,
         if latest_uploads.count() > 1:
             upload_session = latest_uploads.first()
         else:
-            upload_session, _created = UploadSession.objects.get_or_create(resource=layer)
+            upload_session, _created = UploadSession.objects.get_or_create(resource=layer, defaults={'user': theuser})
         upload_session.user = theuser
```
afabiani, line 461: we should probably remove this line too then.

alistairjmcintyre: I thought about this, but it depends on if the code should alter the user of an existing Upload Session?

afabiani: If I'm correctly interpreting the code, it looks like the intention was to update the Upload Session user in any case, but probably we are missing an `upload_session.save()` after the assignment `upload_session.user = theuser`.

In practice the associated user to an Upload Session will be used only to show the incomplete/pending uploads on the Upload page. A `superuser` will be able to see all of them anyway. So IMHO we should always update the user and make sure it will be correctly saved later on.

alistairjmcintyre: Should the upload session be immediately saved if it's just been created?

My assumption is that the `latest_uploads` object is retrieved because there may have been multiple upload sessions for the same resource (layer in this case).

I think there's 3 cases here:
 1. Multiple uploads exist already, and we want to reuse the first one (most recent)
 2. A single upload session exists and we want to use that one
 3. No upload session exists, so we want to create a new one
 
 The code might be better off looking something like this:
 
 ```
 latest_uploads = UploadSession.objects.filter(resource=layer).order_by('-date')
 if latest_uploads.count() >= 1:
     upload_session = latest_uploads.first()
     upload_session.user = theuser
 else:
     upload_session = UploadSession.objects.create(resource=layer, user=theuser)
```

It also makes it easier to do an `upload_session.save()` after the user update.

Let me know what you think @afabiani :)

afabiani: @alistairjmcintyre yes, I agree with your solution. Please proceed! Thanks :)

alistairjmcintyre: Awesome. I've pushed the change. I chose to omit the `save()` call as one is made further down the same method. 

I also swapped the `.count()` to a `len()` call as this won't do a `SELECT COUNT(*) ...` database query, instead just checking how many objects in the python iterable instead.

afabiani (approved):

PR #8807, opened 2022-02-14
[Fixes #8687] Simplify Upload Size Limiting setup

### LINKED ISSUE #8687: Simplify Upload Size Limiting setup
## Overview

* `file_upload_handler` size limit
    * We have 3 size limits settings available, but the `file_upload_handler` is too complex. (https://github.com/GeoNode/geonode/blob/master/geonode/upload/migrations/0034_uploadsizelimit.py#L13-L17)
    * This limit prevents the upload of large files to the server, even before the django form validations.
    * It should be always greater than the other sizes, so instead of exposing this config through the size limiting api. It can be better set as a Django settings variable and corresponding env variable.
    * This variable should indicate the multiplier to be used:
        * For example: 3 times, means that the `file_upload_handler` limit is going to be 3 times the value of the `total_upload_size_sum` limit.
    * Simplifying it, we can also prevent misconfiguration that can lead to the creation empty datasets/documents.
* Update comments on the env variables related to it, letting it clear that it's only used to create the objects and when modified it won't modify the current already existing value.
* Improve code comments in the UploadManager to better differentiate document limits and datasets limits.  https://github.com/GeoNode/geonode/blob/master/geonode/upload/models.py#L95-L101

mattiagiupponi:
Remove `DEFAULT_MAX_BEFORE_UPLOAD_SIZE` as configuration and converted into an enumerator
by default will be `(DEFAULT_MAX_UPLOAD_SIZE * 2) + 2097152 (2MB)`

## Checklist

For all pull requests:


giohappy:
@mattiagiupponi tests for execution requests api was already fixed.

geonode/upload/migrations/0034_uploadsizelimit.py
```diff
@@ -13,7 +15,7 @@
     {
         "slug": "file_upload_handler",
         "description": 'Request total size, validated before the upload process. This should be greater than "total_upload_size_sum".',
-        "max_size": settings.DEFAULT_MAX_BEFORE_UPLOAD_SIZE
+        "max_size": DEFAULT_MAX_BEFORE_UPLOAD_SIZE
     },
```
italogsfernandes, line 19: I think we can delete this object.

italogsfernandes (changes requested):

geonode/base/enumerations.py
```diff
@@ -503,3 +504,5 @@
 )
 
 LAYER_TYPES = ['vector', 'raster', 'remote', 'vector_time']
+
+DEFAULT_MAX_BEFORE_UPLOAD_SIZE = (settings.DEFAULT_MAX_UPLOAD_SIZE * 2) + 2097152
```
giohappy, line 508: @mattiagiupponi we don't want to use an enumeration. The size for the upload handler must be calculated according to this formula but dynamically from the dataset (or document) upload size set inside the DB.

geonode/upload/admin.py
```diff
@@ -60,21 +60,10 @@ def clean(self):
...
 
         if slug in after_upload_slugs_list:
-            handler_max_size = UploadSizeLimit.objects.filter(
-                slug='file_upload_handler'
-            ).values_list('max_size', flat=True)
-            if handler_max_size and max_size * 2 >= max(handler_max_size):
+            if max_size * 2 >= max_size_before_upload:
                 raise forms.ValidationError(_(
                     "To avoid errors, max size should be at least 2 times "
                     "smaller than the value of 'file_upload_handler'."
```
italogsfernandes, line 69: I think we also don't need this validation anymore, since max_size_before_upload will be always calculated using the document_upload_size and dataset_upload_size

italogsfernandes (changes requested):
Everything looks good, only some old validations needs to be removed because they don't make sense for a computed value.

geonode/upload/admin.py
italogsfernandes: By consequence the string in these files can be removed:
* geonode/locale/en/LC_MESSAGES/django.po:6592
* geonode/locale/fr/LC_MESSAGES/django.po:6730
* geonode/locale/pt/LC_MESSAGES/django.po:6818
* geonode/locale/pt_BR/LC_MESSAGES/django.po:6591
* geonode/upload/api/serializers.py:255

.env_dev
```diff
@@ -38,7 +38,7 @@ GEONODE_DB_CONN_TOUT=5
 DEFAULT_BACKEND_DATASTORE=datastore
 BROKER_URL=amqp://admin:admin@localhost:5672//
 CELERY_BEAT_SCHEDULER=celery.beat:PersistentScheduler
-ASYNC_SIGNALS=False
+ASYNC_SIGNALS=True
```
afabiani, line 41: Can we revert this one to `False`?

afabiani (changes requested):

.env_dev
mattiagiupponi: Done

PR #8817, opened 2022-02-17
Make celery wait for geonode to launch

EHJ-52n:
## Description

This PR ensures that the celery container is waiting for the geonode container to launch before itself.
By doing this, the migrations and other django init steps are only applied once.

This is achieved by two changes:

* Add netcat to the geonode container image used by django and celery service
* Add logic to wait via netcat for the geonode providing container.
  This is controlled by the following three environment variables:

  * `CELERY_DONT_WAIT_FOR_GEONODE`
  * `CELERY_GEONODE_CONTAINER_HOSTNAME`
  * `CELERY_GEONODE_CONTAINER_PORT`

  The first one can be used to disable the whole waiting procedure if not required.
  The other two are used by netcat to check for GeoNode

## Checklist

For all pull requests:


afabiani:
That's nice improvement @EHJ-52n thanks, I will test this and include in the `geonode-project` too.

giohappy:
@EHJ-52n if the reason for this PR is to avoid double initialization we should first check why it happens to you. Celery already shouldn't re-initialize the DB, therefore either something is not correctly set up or something must be fixed for the solution already in place.
I would avoid adding a new solution to the same problem if the only reason is that one.

If we find this feature useful beyond the initial reasons, ok, but in that case, we should avoid duplicated functionality.

gannebamm:
Additionally, this could be backported to 3.3.x? @EHJ-52n Do you think the amount of work to backport it is doable?

EHJ-52n:
> Additionally, this could be backported to 3.3.x? @EHJ-52n Do you think the amount of work to backport it is doable?

AFAIK, this doesn't include any geonode specific changes, it is version agnostic.

@giohappy I agree with you. I remember that we cleaned up the tasks invocation, too. Hence, this might be somehow obsolete. In the master branch, this was fixed in 45a954bf6f01552743830c23167a00fd239421ff.

But it is still not fixed in the GeoNode project repository: 
https://github.com/GeoNode/geonode-project/blob/master/src/entrypoint.sh#L50-L95

afabiani (changes requested):
Not needed anymore. This PR can be closed IMHO.

giohappy:
>Not needed anymore. This PR can be closed IMHO.

I agree

PR #8857, opened 2022-02-24
[Fixes #8347] Implement limiting for the number of upload sessions per user

### LINKED ISSUE #8347: Implement limiting for the number of upload sessions per user
In view of introducing additional controls over upload funtionality (e.g. [size limit](https://github.com/GeoNode/geonode/issues/8346), quota limits, etc.) we want to introduce the ability to control the maximum number of active upload sessions per user.

It will be controlled by a new global configuration (to be placed inside GeoNode settings module). A numerical value will set the numbers of active upload sessions that a user can execute. Any additional attempt to upload will be blocked until a previous session has completed. 

The implementation will:

- provide the means to let frontend clients cooperate (expose the limits to the client through configurations). For example, client might block the upload operation from the beginning by inspecting files size
- integrate with the Upload API to support the required messaging between the frontend and the upload backend (e.g. for reporting limits, errors, etc.)
- provide an internal API that can be used by GeoNode modules to query for the configured limits
- offer a high level interface to be integrated inside the Upload module (APIs / views), with an underlying concrete implementation that could be adapted for master (the target of this proposal) and 3.3.x (optional backporting).

italogsfernandes:
references: #8347

## Checklist

For all pull requests:


giohappy:
@italogsfernandes you did more than what was the intention of the issue. We meant to have a single limit for all users, not per user. The "per user" was referred to the concurrent uploads done by the user.

Is there a global parallelism limit If we don't want to configure a specific for single users?

italogsfernandes:
@giohappy yes, it has a global parallelism limit.
It tries to:
* Get limits for the user, if he has one.
* Then limit for the group, if he has one. (for example, the admins group can upload many datasets at one, more than the global limit) 
* Finally, it will look for the global limit.

By default the global limit is set to 5, and it's the only needed limit.
The users and groups limits can be set if it's the case but it's not mandatory.

The following method: https://github.com/GeoNode/geonode/pull/8857/files#diff-a2bccf9e0cea1b8db1dcc77dd0e0c6c5a19b8f94f23678c00104043365bb598eR97
takes care of the implementation of that logic.

giohappy:
Thanks for the clarification @italogsfernandes.
@mattiagiupponi please proceed with the review.

afabiani (changes requested):
Even if not yet widely used, we have currently other ways to upload/ingest some datasets, i.e.: through the `ingest` endpoints of the `ResourceManager` (ref. `[code](https://github.com/GeoNode/geonode/blob/master/geonode/resource/manager.py#L394)`)

There are also some `async api endpoints` exposing access to thos methods (ref. [code](https://github.com/GeoNode/geonode/blob/master/geonode/base/api/views.py#L679))

Finally, be careful on how you will eventually set the limits on such endpoints, since there are other componets leveraging on them, like the `harvesters` (ref. [code](https://github.com/GeoNode/geonode/blob/master/geonode/harvesting/harvesters/base.py#L308))

@giohappy what do you suggest here?

giohappy:
@afabiani IMHO all the controls that have been implemented lately (upload limits, concurrency) will need to be ported to the new resource manager once the upload in master will be replaced with it.
For the moment I'm fine to have it only for the current logic (which is the one also used by the upload API).

afabiani (dismissed):

afabiani:
So far it works fine, the only thing is that when exceeding the parallelism size the error message is not quite clear.
I would implement something similar to the upload size if possible

image

image

mattiagiupponi (changes requested):
I don't see any test for this feature, is possible to add some of them?
For the admin page, serializer, upload, etc...

giohappy:
>So far it works fine, the only thing is that when exceeding the parallelism size the error message is not quite clear.
I would implement something similar to the upload size if possible

@afabiani I agree. I will ask David to implement it. What is the message returned by the API?

giohappy:
@afabiani for this and other error responses (example the upload size) could we implement internal error codes, to make the client able to manage them in a robust way?

italogsfernandes:
> I don't see any test for this feature, is possible to add some of them? For the admin page, serializer, upload, etc...

@mattiagiupponi Tests were added, ready for review.

giohappy:
@italogsfernandes the same approach adopted for the upload size limit should be employed for consistency.

Regarding the returned error, it will be adopted once [this PR](https://github.com/GeoNode/geonode/pull/8884) will be merged

giohappy:
@italogsfernandes after reviewing the business logic and the management of user / group specific limits we came to the conclusion that only the initial proposal of a single global limit should be implemented.

There are several reasons for this decision:

- there's not an explicit logic to decide whether to apply a limit for a user if it's a member of both of a group with a limit and a group without limits. The priority should depend on an explicit priority, including "no limit" rules. In general, if a user is a member of a group without limits this should win. However, there are fewer cases where the reverse is required.
- the admin of rules is counterintuitive:
   - you have an editable slug input that is however filled by the admin logic
   - if you change the user for a limit a new limit is created, instead of replacing the current one
   -  applying the same limit to multiple users or groups requires creating one limit for each of them. A 1:N relationship is expected here, and rules could have been searched by back-ref instead of through the slug value

By the way the main concern here is the business logic. So, **long story short, let's remove the user / group limits (from the model and everywhere) and just keep the global limit as we for the upload size limits**.

italogsfernandes:
@giohappy Updated, now it has only a limit to be set.

geonode/upload/admin.py
```diff
@@ -59,5 +62,15 @@ def has_delete_permission(self, request, obj=None):
         return super(UploadSizeLimitAdmin, self).has_delete_permission(request, obj)
 
 
+class UploadParallelismLimitAdmin(admin.ModelAdmin):
+    list_display = ('slug', 'description', 'max_number',)
+
+    def has_delete_permission(self, request, obj=None):
+        if obj and obj.slug == "default_max_parallel_uploads":
```
mattiagiupponi, line 69: Via the admin interface, the `default_max_parallel_uploads` is not available for the user to change the value because it will be created at the first upload. Adding it to the fixture could fix this problem.
Plus I'll suggest removing the possibility to add new parallelism since they will not have any effect, should be enough to do:
```python
def has_add_permission(self, request):
    return False
```

geonode/upload/forms.py
```diff
@@ -154,6 +160,14 @@ def _get_files_paths_or_objects(self, cleaned_data):
 
         return uploaded, files
 
+    def validate_parallelism_limit_per_user(self):
+        max_parallel_uploads = self._get_max_parallel_uploads()
+        parallel_uploads_count = self._get_parallel_uploads_count()
+        if parallel_uploads_count >= max_parallel_uploads:
```
mattiagiupponi, line 166: It could be a problem to have the current parallel uploads at the same as max uploads?
So having this: `if parallel_uploads_count > max_parallel_uploads` could raise many issues?

mattiagiupponi (changes requested):

geonode/upload/forms.py
italogsfernandes: in this case, this means that: (in the case of a limit of 5)
* We already have 5 active uploads.
* This form validation is called before creating one more upload.
* If we already have the limit and we create another one, we will end with 6 active uploads.

So I think it will be better to keep the `>=` instead of `>`.

mattiagiupponi: great, thanks for the clarification

geonode/upload/admin.py
italogsfernandes: Updated

mattiagiupponi (approved):

PR #8904, opened 2022-03-07
[Fixes #8887] Fixes and adjustments required for dataset direct download

### LINKED ISSUE #8887: Fixes and adjustments required for dataset direct download
The following points must be addressed within `dataset_download` 

- zip output only works for vector datasets. If you try the download of a raster dataset an error is returned, so a test is also required for this case
- the usage of gs_restconfig Catalogue is not useful here, since we're not interacting with Geoserver REST APIs. Moreover, it causes strange errors, probably because of not well-formed requests. This was happening on master demo, but it's been fixed with a manual fix on the server where it's been replaced with the global `http_client` instance which is available from `geonode.utils`. Therefore using GeoNode's `http_client` is adviced
- an `access_token` must be set only if not None. It's None for anonymous users
- a `csrf_exempt` decorator should be added to avoid blocking downloads from third party clients

mattiagiupponi:
This PR introduces:
1. substitution of `GeoserverCatalog` client with `HttpClient`
1. Adding `@csrf_exempt` to the view
1. assumption that the download format is  `'application/zip' if dataset.is_vector() else 'image/tiff'` 
1. adding the `access_token` to the URL only if the user is not anonymous
1. Add test to ensure that the download format is the correct one based on the dataset_type

@giohappy Can you confirm the assumption on point 3?

## Checklist

For all pull requests:

### DIFF
--- a/geonode/layers/views.py
+++ b/geonode/layers/views.py
@@ -52,7 +52,7 @@ from django.views.decorators.http import require_http_methods
 
 from geonode import geoserver
 from geonode.layers.metadata import parse_metadata
-from geonode.proxy.views import fetch_response_headers
+from geonode.proxy.views import download, fetch_response_headers
 from geonode.resource.manager import resource_manager
 from geonode.geoserver.helpers import set_dataset_style, wps_format_is_supported
 from geonode.resource.utils import update_resource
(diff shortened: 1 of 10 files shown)


geonode/layers/views.py
```diff
@@ -812,40 +811,38 @@ def dataset_download(request, layername):
...
+    if download_format and not wps_format_is_supported(download_format, dataset.subtype):
         logger.error("The format provided is not valid for the selected resource")
         return JsonResponse({"error": "The format provided is not valid for the selected resource"}, status=500)
 
     # getting default payload
     tpl = get_template("geoserver/dataset_download.xml")
     ctx = {
         "alternate": dataset.alternate,
-        "download_format": download_format
+        "download_format": 'application/zip' if dataset.is_vector() else 'image/tiff'
```
giohappy, line 824: @mattiagiupponi you're not using the (optional) `download_format` obtained from the `export_format` param

giohappy (approved):

PR #9214, opened 2022-04-22
[Fixes #9212] Login with basic auth or token headers or apikey GET parameter

### LINKED ISSUE #9212: Login with basic auth or token headers or apikey GET parameter
When the `LOCKDOWN_GEONODE` setting is active it isn't possible to embed views inside third-party sites.

A generic solution is proposed to allow the creation of user sessions from Basic Auth or Bearer token auth headers, or from the token set inside an `apiky` parameter.

A new `LoginWithHeaderOrKeyMiddleware` can be activated with a new `LOGIN_WITH_HEADER_OR_KEY` setting, which is False by default.
When the middleware is activated any request containing auth parameters or headers will log in the user and a session is created. 

The creation of a session is required to support the requests performed by the client inside the embedded maps because the client doesn't include the headers or parameters from the initial request, and so a login request is shown if sessions cookies are not created.

giohappy:
ref #9212 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/base/auth.py
+++ b/geonode/base/auth.py
@@ -103,7 +103,7 @@ def get_auth_token(user, client=settings.OAUTH2_DEFAULT_BACKEND_CLIENT_NAME):
 
 def get_auth_user(token, client=settings.OAUTH2_DEFAULT_BACKEND_CLIENT_NAME):
     try:
-        access_token = AccessToken.objects.all(token=token).first()
+        access_token = AccessToken.objects.filter(token=token).first()
         if access_token and access_token.is_valid():
             return access_token.user
     except Exception:
@@ -170,7 +170,7 @@ def get_token_from_auth_header(auth_header, create_if_not_exists=False):
             return get_auth_token(user) if not create_if_not_exists else get_or_create_token(user)
     elif re.search('Bearer', auth_header, re.IGNORECASE):
         token = re.compile(re.escape('Bearer '), re.IGNORECASE).sub('', auth_header)
-        access_token = AccessToken.objects.all(token=token).first()
+        access_token = AccessToken.objects.filter(token=token).first()
         if access_token and access_token.is_valid():
             return access_token
     return None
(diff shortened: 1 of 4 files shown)


geonode/base/auth.py
```diff
@@ -105,8 +105,9 @@ def get_auth_user(access_token, client=settings.OAUTH2_DEFAULT_BACKEND_CLIENT_NA
     try:
         Application = get_application_model()
         app = Application.objects.get(name=client)
-        user = AccessToken.objects.filter(token=access_token, application=app).order_by('-expires').first().user
-        return user
+        access_token = AccessToken.objects.filter(token=access_token, application=app).order_by('-expires').first()
+        if access_token.is_valid():
```
mattiagiupponi, line 109: in case the `access_token` does not exist, the `first()` will return a None.
To make it safer is better to validate that the token exists, like:

```python
if access_token and access_token.is_valid():
    # do stuff
```

geonode/base/auth.py
```diff
@@ -170,7 +171,10 @@ def get_token_from_auth_header(auth_header, create_if_not_exists=False):
         if user and user.is_active:
             return get_auth_token(user) if not create_if_not_exists else get_or_create_token(user)
     elif re.search('Bearer', auth_header, re.IGNORECASE):
-        return re.compile(re.escape('Bearer '), re.IGNORECASE).sub('', auth_header)
+        token = re.compile(re.escape('Bearer '), re.IGNORECASE).sub('', auth_header)
+        access_token = AccessToken.objects.get(token=token)
```
mattiagiupponi, line 175: as above, better to use the `first()` and check it the token is not `None` or use first the `exists()`

mattiagiupponi (changes requested):

geonode/base/auth.py
giohappy: done

mattiagiupponi (approved):

PR #9800, opened 2022-08-02
[Fixes #9799] Thesaurus selectbox performance in metadata editor is very slow

### LINKED ISSUE #9799: Thesaurus selectbox performance in metadata editor is very slow
## Expected Behavior
I added the Agrovoc(https://www.fao.org/agrovoc/) with about 40.000 keywords to Geonode and wanted to use it as a Thesaurus to classify my metadata. I expected the metadata editor to process the data rather fast.

## Actual Behavior
But it takes a long time for the selectbox in the metadata editor to show results. And if i search again it takes the same amount of time again to find another word.

  1. Download Agrovoc from https://agrovoc.fao.org/latestAgrovoc/agrovoc_core.nt.zip
  2. Unzip the file
  3. Run load-command from https://github.com/zalf-rdm/geonode-agrovoc-importer

## Specifications

  - GeoNode version: 4.x
  - Installation method (manual, GeoNode Docker, SPCGeoNode Docker): Docker, Local, Kubernetes
  - Platform: Linux
  - Additional details:

mwallschlaeger:
References: #9799

As described in the issue. The requests to the database are very slow. This is basically related to a for loop used in the current implementation. To evaluate my changes i wrote a small benchmark to show the performance improvements done by this commit.

```
import argparse
import time

import django
django.setup()

from geonode.base.models import ThesaurusKeyword, ThesaurusKeywordLabel

parser = argparse.ArgumentParser(description='Test thesaurus performance in GeoNode')
parser.add_argument('-tid', dest='thesaurus_id', type=int, required=True, help="pass the thesaurus id from the Thesaurus Table to select catalog to run on")
parser.add_argument('-lang', dest='language', default="en", type=str, help="pass language the results are searched for")
args = parser.parse_args()

start_time = time.time()
print("starting run with current GeoNode implementation ...")
qs_local = []
qs_non_local = [("", "------")]
for key in ThesaurusKeyword.objects.filter(thesaurus_id=args.thesaurus_id):
    label = ThesaurusKeywordLabel.objects.filter(keyword=key).filter(lang=args.language)
    if label.exists():
        qs_local.append((label.get().keyword.id, label.get().label))
    else:
        qs_non_local.append((key.id, key.alt_label))

qs = qs_local + qs_non_local
print("run finished in: {} and found {} objects ...".format(round(time.time() - start_time, 5), len(qs)))

print()

start_time = time.time()
print("starting run with suggested improvement ...")

keyword_id_for_given_thesaurus = ThesaurusKeyword.objects.filter(thesaurus_id=args.thesaurus_id)  
qs_keyword_ids = ThesaurusKeywordLabel.objects.filter(lang=args.language, keyword_id__in=keyword_id_for_given_thesaurus).values("keyword_id")
not_qs_ids = ThesaurusKeywordLabel.objects.exclude(keyword_id__in=qs_keyword_ids).order_by("keyword_id").distinct("keyword_id").values("keyword_id")

qs_local = list(ThesaurusKeywordLabel.objects.filter(lang=args.language, keyword_id__in=keyword_id_for_given_thesaurus).values_list("keyword_id", "label"))
qs_non_local = list(keyword_id_for_given_thesaurus.filter(id__in=not_qs_ids).values_list("id", "alt_label"))

qs=qs_local + [("", "-------")] + qs_non_local
print("run finished in: {} and found {} objects ...".format(round(time.time() - start_time, 5), len(qs)))
```

When i run the script with the agrovoc loaded into the database (https://github.com/zalf-rdm/geonode-agrovoc-importer) on my Desktop PC i get the following results:

```
❯ python debug_test.py -tid 13
starting run with current GeoNode implementation ...
run finished in: 112.2935 and found 40251 objects ...

starting run with suggested improvement ...
run finished in: 0.26096 and found 40251 objects ...
```
So the new implementation is something like 400 times faster than the current implementation.

In my opinion the way of handling alt_labels so labels which are not available in the current language could also be improved.

## Checklist

For all pull requests:


geonode/base/forms.py
```diff
@@ -337,16 +337,14 @@ def _define_choicefield(self, item, required, tname, lang):
...
-                qs_non_local.append((key.id, key.alt_label))
+        keyword_id_for_given_thesaurus = ThesaurusKeyword.objects.filter(thesaurus_id=item)
+        qs_keyword_ids = ThesaurusKeywordLabel.objects.filter(lang=lang, keyword_id__in=keyword_id_for_given_thesaurus).values("keyword_id")
+        not_qs_ids = ThesaurusKeywordLabel.objects.exclude(keyword_id__in=qs_keyword_ids).order_by("keyword_id").distinct("keyword_id").values("keyword_id")
+
+        qs_local = list(ThesaurusKeywordLabel.objects.filter(lang=lang, keyword_id__in=keyword_id_for_given_thesaurus).values_list("keyword_id", "label"))
+        qs_non_local = list(keyword_id_for_given_thesaurus.filter(id__in=not_qs_ids).values_list("id", "alt_label"))
 
-        return qs_non_local + qs_local
+        return qs_local + [("", "-------")] + qs_non_local
```
mattiagiupponi, line 347: Hi @mwallschlaeger I guess here is better to maint the `[("", "-------")]` as first, so during the preview in the select, is at the beginning. Otherwise, I guess it will be in the middle of the list.

mattiagiupponi (changes requested):
Hi, @mwallschlaeger I see that no tests are added, the old ones are still valid? Or do they need some improvement?
(I'm asking since the build was failing :) )

geonode/base/forms.py
mwallschlaeger: yes and no. I think its good to have the [("", "-------")] at first. But I would suggest to put it in between the qs_local and the qs_non_local as well, so the user knows that there is a reason why there are appearing entries in another language. 

Again for my usecase with the AGROVOC there are a lot of entries not available in all languages. That means that even if a users searches for something there are always many many entries which are not related to the search.

Imagining looking for something like "v" in a list of car manufactures. Therefore I would suggest an output like:
```
---------
Vandenbrink
Vauxhall
Vector
Vencer
Venturi
Venucia
VinFast
VLF
Volkswagen
Volvo
---------
北京汽车制造厂有限公司
绅宝汽车
银翔
金杯
```

So in this case the second "---------" separates the translated entries from the not translatable ones.

mattiagiupponi: Yep having it in both places is a good option

mwallschlaeger:
i got your point on this. I think its good to have the "------" in the beginning of the list. But in some cases e.g. in the case of the AGROVOC you have a huge list of items not available in every language. Therefore I suggest adding a "------" in between the qs_local entries and the qs_non_local entries so users can determine where the local entries end. 

Imagine a list of car manufactures we search in with "v". But some manufactures have no translation for your language. The list would look something like:
```
-------
Vandenbrink
Vauxhall
Vector
Vencer
Venturi
Venucia
VinFast
VLF
Volkswagen
Volvo
-------
比亚迪汽车
上汽通用五菱汽车股份有限公司
奇瑞汽車股份有限公司
廣州汽車集團股份有限公司
東風汽車集團有限公司
```

In my opinion its good to show the qs_non_locals last and separate them by another "-------" entry.

geonode/base/forms.py
```diff
@@ -337,16 +337,14 @@ def _define_choicefield(self, item, required, tname, lang):
...
-        qs_local = []
-        qs_non_local = [("", "------")]
-        for key in ThesaurusKeyword.objects.filter(thesaurus_id=item.id):
-            label = ThesaurusKeywordLabel.objects.filter(keyword=key).filter(lang=lang)
-            if label.exists():
-                qs_local.append((label.get().keyword.id, label.get().label))
-            else:
-                qs_non_local.append((key.id, key.alt_label))
+        keyword_id_for_given_thesaurus = ThesaurusKeyword.objects.filter(thesaurus_id=item)
+        qs_keyword_ids = ThesaurusKeywordLabel.objects.filter(lang=lang, keyword_id__in=keyword_id_for_given_thesaurus).values("keyword_id")
```
mattiagiupponi, line 341: Hi @mwallschlaeger 
It looks like there is some issue with changing the language. I'm trying it on master.
I'm trying it in Italian to see how it behaves. Setting in Italian, I have: `lang = it-it`,
image

 so when it looks for the qs_keyword_ids, the queryset is always empty
image
 since it is saved with `lang=it`
image

So at the end qs_local is empty, 
image
while qs_non_local is full:
image

geonode/base/views.py
```diff
@@ -301,23 +302,28 @@ def get_results(self, context):
...
-            else:
-                if self.q in key.alt_label:
-                    qs_non_local.append(key)
-                elif not self.q:
-                    qs_non_local.append(key)
-
-        return qs_non_local + qs_local
+
+        # help to get real language string
+        requested_lang = get_language_from_request(self.request)
```
mattiagiupponi, line 309: If you have time, can you please add a couple of tests on the APi?

mattiagiupponi (changes requested):

geonode/base/views.py
mwallschlaeger: i updated this mechanism again to be more clean in my opinion.

mwallschlaeger: i am going to add a testcase for different language codes here.

mwallschlaeger:
OOOK. The main problem with the languages is that in django shell in django.utils.translation.get_language() returns "en" but if django.utils.translation.get_language() got executed via the browser you get a language including the country like "en-us". Therefore, i added a function which removes the country code. Sadly there is no function in the django translation module which reliable returns the language without country code. So the search functions in geonode.base.view and geonode.base.form now try to first find entries with language code including the country, because for some languages this makes sense. If no entries are found for this, we change the language to only include the language ("en"). And if this results still contain nothing geonode shows the alt_labels.

I also added some tests which set the language with and without country code and try to get results from the _get_thesauro_keyword_label function in ThesaurusAvailableForm.

In current implementation geonode always returns alt_labels because its not able to handle the language_code including country code ...

mattiagiupponi (changes requested):
overall looks ok, there is just an issue with the flake8 formatting. Can you please fix it so the CircleCI test will run?

mattiagiupponi (approved):

PR #10071, opened 2022-09-29
[Fixes #10070] Let the resource manager handle also raw sld

### LINKED ISSUE #10070: Let the resource manager handle also raw sld
When calling the set_style function, the resource/concrete manager expects that an absolute path is provided. Example here:
https://github.com/GeoNode/geonode/blob/601be61843e057f9f2c99bec84e4459768bc086e/geonode/geoserver/tasks.py#L117-L122
It opens the value without checking if it is a string or not.

Also the style update is redundant:
https://github.com/GeoNode/geonode/blob/601be61843e057f9f2c99bec84e4459768bc086e/geonode/geoserver/helpers.py#L410-L420

We can easily use the creative style in override since it goes here only if the style with the layer name is not found. It can be reduced to this:
```python
        try:
            style = gs_catalog.create_style(
                saved_dataset.name, sld,
                overwrite=True, raw=True,
                workspace=saved_dataset.workspace)
        except Exception as e:
            logger.exception(e)
```

Let's be able to handle also the strings, by using something like:
```python
sld = None
if isinstance(base_file, str) or isinstance(base_file, bytes):
    sld = base_file
    base_file = None
else:
    sld = open(base_file, "rb").read()
```

**Additional context**
This is specially needed in the new importer

mattiagiupponi:

## Checklist

For all pull requests:

### DIFF
--- a/geonode/geoserver/helpers.py
+++ b/geonode/geoserver/helpers.py
@@ -407,17 +407,11 @@ def set_dataset_style(saved_dataset, title, sld, base_file=None):
         except Exception as e:
             logger.exception(e)
     else:
-        style = gs_catalog.get_style(saved_dataset.name, workspace=saved_dataset.workspace) or \
-            gs_catalog.get_style(saved_dataset.name)
         try:
-            if not style:
-                style = gs_catalog.create_style(
-                    saved_dataset.name, sld,
-                    overwrite=True, raw=True,
-                    workspace=saved_dataset.workspace)
-            elif sld:
-                style.style_format = _extract_style_version_from_sld(sld)
-                style.update_body(sld)
+            style = gs_catalog.create_style(
+                saved_dataset.name, sld,
+                overwrite=True, raw=True,
+                workspace=saved_dataset.workspace)
         except Exception as e:
             logger.exception(e)
 
--- a/geonode/geoserver/tasks.py
+++ b/geonode/geoserver/tasks.py
@@ -114,12 +114,19 @@ def geoserver_set_style(
         if lock.acquire() is True:
             log_lock.debug(f"geoserver_set_style: Acquired lock {lock_id} for {instance.name}")
             try:
-                sld = open(base_file, "rb").read()
+                sld = None
+                if isinstance(base_file, str) or isinstance(base_file, bytes):
+                    sld = base_file
+                    base_file = None
+                else:
+                    sld = open(base_file, "rb").read()
+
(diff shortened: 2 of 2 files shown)


geonode/geoserver/helpers.py
```diff
@@ -407,17 +407,11 @@ def set_dataset_style(saved_dataset, title, sld, base_file=None):
...
-            gs_catalog.get_style(saved_dataset.name)
         try:
-            if not style:
-                style = gs_catalog.create_style(
-                    saved_dataset.name, sld,
-                    overwrite=True, raw=True,
-                    workspace=saved_dataset.workspace)
-            elif sld:
-                style.style_format = _extract_style_version_from_sld(sld)
-                style.update_body(sld)
```
afabiani, line 420: Not sure we want to remove this check. @giohappy ?

mattiagiupponi: based on what I saw, that update_body call geoserver again with the same call of the create_style. Imo is a duplicated call

afabiani: @mattiagiupponi as far as I remember that was invoked in order to correctly set the `style_format` on GeoServer which was causing some issues later on, when trying to update it.

giohappy: @afabiani @mattiagiupponi why don't we set the style_format parameter for `create_style`, in case `sld` is defined, instead of issuing the `update_body` afterward?

mattiagiupponi: imo the call is still redundant. With some tests, it looks like the update_body does not update the resource body in GS. By calling just with the create_style in override, everything works fine. 
In the end, the two calls are the same, so all that code is a bit redundant

mattiagiupponi: `style_format` check readded as parameter in the call

afabiani: @mattiagiupponi what if the style already exists in the GeoServer catalog?

mattiagiupponi: it has `overwrite=True` which means that the sld provided will override the existing one.

afabiani (approved):

geonode/geoserver/helpers.py
mattiagiupponi: well, it can throw an exception if the sld provided is invalid from geoserver point of view. I that case we log it

PR #10271, opened 2022-11-08
[Fixes #10270] Document creation via API v2

### LINKED ISSUE #10270: Document creation via API v2
The existing document API does not permit the user to upload a new document

Let the API v2 be able to upload document and thanks to the storage manager, to clone them to be available in the geonode environment

mattiagiupponi:
References: #10270

## Checklist

For all pull requests:


geonode/base/models.py
```diff
@@ -777,7 +777,7 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin, ItemBase):
     extra_metadata_help_text = _(
         'Additional metadata, must be in format [ {"metadata_key": "metadata_value"}, {"metadata_key": "metadata_value"} ]')
     # internal fields
-    uuid = models.CharField(max_length=36, unique=True, default=str(uuid.uuid4))
+    uuid = models.CharField(max_length=36, unique=True, default=uuid.uuid4)
```
mattiagiupponi, line 780: @afabiani this fix is needed since the last migration available for the UUID, is marked as a string of the function itself and not a result.
https://github.com/GeoNode/geonode/blob/a41fbfdf9597b4a33f20fa242b5893b678fbbccb/geonode/base/migrations/0080_alter_resourcebase_uuid.py#L17

This end-up up having an error because the system tries to save the UUID as a "function of...." 

This fix, will let geonode decide the UUID (as object) and the fix in the save_method below, helps it to convert it to string if needed

mattiagiupponi:
@afabiani @giohappy The API expects that a file path is sent instead of a file. In this way, we can rely on the storage_manager to clone the file in the local upload folder instead of making the user upload the file along with the request.
Do you agree or should the API be able to upload just the file instead of a path?

mattiagiupponi:
> @mattiagiupponi we should support the upload too, as we do for datasets. That's good to have this option (which we're still waiting for datasets too!) but now we would end up with the following counterintuitive situations:
> 
> Upload from path	Upload from file
> Dataset	NO	YES
> Document	YES	NO

Actually, the dataset pass from the upload which is another flow. Ideally should be to have the upload able to handle also the document.
Btw I got your point, we can add also the file and try to handle both of them

mattiagiupponi:
> @mattiagiupponi we should support the upload, as we do for datasets. That's good to have this option (which we're still waiting for datasets too!) but now we would end up with the following counterintuitive situations:
> 
> Upload from path	Upload from file
> Dataset	NO	YES
> Document	YES	NO

@giohappy with the last changes, the API now accepts two ways:

- `file_path`: str with the file path, so this is the expected payload
```bash
curl --location --request POST 'http://localhost:8000/api/v2/documents' \
--data-raw '{
    "document": {
        "title": "New document",
        "metadata_only": true,
        "file_path": "/home/mattia/example.json"
    }
}'
```

- `doc_file`: a file via a multi-part request:

```bash
curl --location --request POST 'http://localhost:8000/api/v2/documents' \
--form 'title="Super Title2"' \
--form 'doc_file=@"/C:/Users/user/Pictures/BcMc-a6T9IM.jpg"' \
--form 'metadata_only="False"'
```

giohappy (dismissed):
@mattiagiupponi the upload works fine, however, the default permissions are not the same as when I upload the document from the UI. 
It looks like no permissions are set for documents uploaded from the API.

mattiagiupponi:
> @mattiagiupponi the upload works fine, however, the default permissions are not the same as when I upload the document from the UI. It looks like no permissions are set for documents uploaded from the API.

the last commit should resolve this issue. @marthamareal can you please give a try to this new functionality?

marthamareal (approved):

afabiani (approved):

PR #10592, opened 2023-01-26
[Fixes #10521] Make metadata wizard more flexible for custom metadata

### LINKED ISSUE #10521: Make metadata wizard more flexible for custom metadata
Hi @mattiagiupponi 

I saw you worked on the [Extrametadata implementaion ](https://github.com/GeoNode/geonode/blame/31ad2ac459bf184d870d62d833dcfcd5394a40a6/geonode/base/models.py#L2118). Hence I hope it is okay to ping you directly.

I'm just starting implementing cutom metadata for a client, where it would be nice using the existing new ExtraMetadata Model. _Further to keep it generic so that the community can benefit from it._  

### My Idea would be as following:

1. Make the current [dataset_metadata](https://github.com/GeoNode/geonode/blob/8ccf5fb1ecb9e23b9cea63414d51dc7f391498b9/geonode/layers/views.py#L419) more flexible 
2. contrib-apps or geonode project decorate the view for passing additional data and customize panels.html

### 1. Make dataset_metadata more flexible

geonode/layer/views.py
```python
def dataset_metadata(
        request,
        layername,
        template='datasets/dataset_metadata.html',
        ajax=True,
        panel_template='layouts/panels.html', <-- allow panels.html to be extended
        custom_metadata=None): <- pass additional context
      ...

    return render(request, template, context={
        ...
        "panel_template": panel_template,
        "custom_metadata": custom_metadata,
        ...
```
datasets/dataset_metadata.html
```html
        <div id="mdeditor_form" class="col-md-12 form-controls">
          {% form dataset_form using panel_template %} <-- use the panel from settings.py || default
          {# dataset_form|as_bootstrap #}
        </div>
```

### 2. Example of custom metadata implementation

Doing so one would not need to touch the core anymore but could decorate the method as needed from geonode_project ot a contrib app.

As a poc:

my_app/my_view.py
```python
from django.conf import settings
from django.http import HttpResponse
from geonode.base.models import ExtraMetadata
from geonode.layers.views import dataset_metadata, _resolve_dataset, _PERMISSION_MSG_METADATA
import json

def custom_metadata_decorator(view_func):
    def _wrapped_view(request, *args, **kwargs):
        layer = _resolve_dataset(
            request,
            kwargs.get('layername'),
            'base.change_resourcebase_metadata',
            _PERMISSION_MSG_METADATA
        )
        if request.method == 'POST':
            # Call the original view function
            response = view_func(request, *args, **kwargs)
            if response.status_code == 200:
                # Handle additional form input this should be made dynamic from settings ?
                color = request.POST.get('resource-color')
                data = {'color': color}
                json_data = json.dumps(data)
            try:
                extra_meta = ExtraMetadata.objects.get(resource=layer)
                extra_meta.metadata = json_data
                extra_meta.save()
            except ExtraMetadata.DoesNotExist:
                extra_meta = ExtraMetadata.objects.create(resource=layer, metadata=json_data)
                layer.metadata.add(extra_meta)
            return response
        else:
            # On GET, add the extra metadata to the context 
            custom_metadata = {}
            try:
                resource = layer.metadata.first()
                custom_metadata = json.loads(resource.metadata)
            except Exception as e:
                ...
            
            default_panel_template = 'layouts/panels.html'
            panel_template = getattr(settings, 'PANEL_TEMPLATE', default_panel_template)
            response = view_func(request, 
                                *args, 
                                panel_template=panel_template, 
                                custom_metadata=custom_metadata, 
                                **kwargs)
            return response

    return _wrapped_view

dataset_metadata = custom_metadata_decorator(dataset_metadata)
```

my_app/templates/custom_panels.html
```html
{% extends "layouts/panels.html" %}

{% block dataset_abstract %}
{{ block.super }}<br>

<div id="req_item">
    <span><label for="id_resource-title">Color</label>
    <input 
      type="text" 
      name="resource-color" 
      value="{{ custom_metadata.color }}" 
      maxlength="255" 
      placeholder="Color" 
      required="" 
      id="id_resource-color">
</div>

{% endblock %}

{% block layer_extra_metadata %}
{% endblock %}
```

image

---

My question are, 
- do you have a better approach in mind instead of a decorating the existing view (also in regard of mapstore2 and accessing the data)? (I would not patch the ResouceBase Model anymore but rely on the json field)
- further I wonder why a ManyToMany relation is used. In which case would a dataset have several ExtraMetadatas attached?

best regards

Toni

t-book:
References: #10521

The aim of this PR is **not** to implement extra metadata but rather to allow contrib apps to implement their own requirements of extrametadata forms_. (Every contrib app is then responsible for generating a metadata form and handling validation._) The foundation for implementation of extra metadata is the m2m relation from BaseResource Model to ExtaMetdata

**You can see my ongoing work of a contrib app here**
_* Please note that the app is still under development and some things can still change._

- [Repo](https://github.com/csgis/custom_metadata)
- [Docs](https://geonode-custom-metadata-app.readthedocs.io/en/latest/)

[... Text has been shortend as parts were obsolete ... ]

---

- [x] Commits changing the **settings**, **UI**, **existing user workflows**, or adding **new functionality**, need to include documentation updates
- [x] Commits adding **new texts** do use gettext and have updated .po / .mo files (without location infos)


giohappy:
Thanks for this contribution @t-book!

I've not dug into all the details, and I haven't had the time to test it yet, but it looks like a good solution in the short term to have a generic way to extend the metadata management and publishing.

In the future, I think we should separate metadata from the resource base model, and provide a solid mechanism to extend the model without relying on "blob" fields like the `extra_metadata` (which, btw, is serving us quite well for the moment :) )

WRT the addition to the `catalogue_post_save` I suppose it's there to let a custom metadata template use the extra metadata information, right?

t-book:
Thanks a lot @giohappy for your brief feedback. I know at first sight it looks a lot but in general these are 3-4 for changes multiplied by the apps. I agree in future we should provide this functionaly in core. By now one could just hook into the metadata process and create own forms. 

The idea is that the metadata handling is not tight coupled to my contrib app, but one could define a custom app or version that can be registered as `CUSTOM_METADATA_APP` in settings and then gets respected by the main project:

https://github.com/GeoNode/geonode/blob/a1613355c1b586b168ffa7334a086b5a25a0f3ee/geonode/urls.py#L70

> (which, btw, is serving us quite well for the moment :) )

Nice! Even better my contrib app also renders forms even if the metadata has been added via REST ;)


> WRT the addition to the catalogue_post_save I suppose it's there to let a custom metadata template use the extra metadata information, right?

yes exactly, 

https://github.com/GeoNode/geonode/blob/a1613355c1b586b168ffa7334a086b5a25a0f3ee/geonode/catalogue/models.py#L84-L95

if `EXTRA_METADATA_ENABLED` is enabled, the instance is enriched with the metadata m2m. Which then is available for XML rendering.  There was no need to update the `full_metadata.xml` template handling as it already has been dynamic. So contrib apps can just define ther own version:


```  
</gmd:pointOfContact>
       {% endwith %}
       <gmd:AdditionalData>
       {% for extra in layer.extra_metadata %}
         {% for key, value in extra.items %}
           {% with key=key|valid_xml_element %}
               <gmd:{{ key }}>{{ value }}</gmd:{{ key }}>
           {% endwith %}
         {% endfor %}
       {% endfor %}
     </gmd:AdditionalData>
```
https://github.com/csgis/custom_metadata/blob/c130ca604292c422a4e32bb056e4de0420429593/templates/custom_metadata/full_metadata.xml#L304-L310

and render the extra fields where ever they want.

giohappy:
Thanks @t-book for the clarifications.

giohappy:
@t-book the PR misses tests. I iwlll mark it as WIP for the moment, ok?

t-book:
Dear @giohappy 

I have taken a look at the tests and must confess that I am at a loss at the moment.
For example, trying to do a simple testing of the if/else in the base serializer class
https://github.com/GeoNode/geonode/blob/fc605993a703262da459100686026d65501043bf/geonode/base/api/serializers.py#L487-L492

as follows does not work because the settings are always read (from the settings) when the class is loaded and the `EXTRA_METADATA_ENABLED` can never get `True`(set by the decorator).

```python
    @override_settings(EXTRA_METADATA_ENABLED=True) # has no effect on ResourceBaseSerializer
    def test_serializer_metadata_deferred_behavior(self):
        serializer = ResourceBaseSerializer(self.layer)
        metadata_field = vars(serializer.__class__)['_declared_fields'].get('metadata')
        self.assertEqual(metadata_field.deferred, True)
```

After many attempts to fix this problem in various ways, I consider testing class attribute code that relies on settings to be almost hopeless. _I have not found an example in the geonode core of how others have dealt with this. Do you know of one?_
        
**Currently, I can only imagine two solutions.** 
- Either to run these tests with another .env [settings file](https://github.com/GeoNode/geonode/blob/master/.env_test) that sets EXTRA_METADATA_ENABLED to `True` and allows tests to check the behaviour.
- Or we say goodbye to the idea of making the core more "pluggable" for contrib apps. In this case, a plugin/geonode_project would have to overwrite all views independently. A bit sad after all the work.

I am grateful for your opinion.

mattiagiupponi:
Hi @t-book 
Unluckily putting the variable in that way makes the serializer initiated and the test ignores it (I don't know why).
I tried to debug and it looks like the ResourcebaseSerializer is called even before the test starts.
A possible test could be to create the metadata as a field:

In the serializer:
```python
metadata = MetadataRelationField(ExtraMetadataSerializer)
```
The field:
```python
class MetadataRelationField(DynamicRelationField):
    def __init__(self, serializer_class, many=True, queryset=None, embed=False, sideloading=None, debug=False, **kwargs):
        deferred = getattr(settings, "EXTRA_METADATA_ENABLED", False)
        super().__init__(serializer_class, many, queryset, embed, sideloading, debug, deferred=deferred, **kwargs)
```

I didn't deep try it, but from a quick debug looks like the variable is correctly assigned.
Can you try it this way or by `patching` the settings instead of using the `override_settings`?

geonode/base/api/serializers.py
```diff
@@ -485,7 +484,12 @@ def __init__(self, *args, **kwargs):
 
         self.fields["favorite"] = FavoriteField(read_only=True)
 
-    metadata = DynamicRelationField(ExtraMetadataSerializer, embed=False, many=True, deferred=True)
+    # Load metadata_records for contrib apps
+    if getattr(settings, "EXTRA_METADATA_ENABLED", False):
```
mattiagiupponi, line 488: Hi @t-book i would evaluate setting the variable as follow, by providing the default argument it will be always false if not declared

```python
deferred = getattr(settings, "EXTRA_METADATA_ENABLED", False)
```

geonode/urls.py
```diff
@@ -65,6 +65,23 @@
     url(r"^messages/", include(msg_urls)),
 ]
 
+# custom_metadata
+if settings.EXTRA_METADATA_ENABLED:
```
mattiagiupponi, line 69: I would evaluate letting the external contrib app injects the URLs patterns needed for the following reasons:
- We are introducing in the core a concept that could be not part of it. This means that for each contrib app, the core should be aware of the URLs available (that is what we want to avoid)
- It may be more dynamic and safe to let the contrib app inject its own URL like (for example) we did for the [geonode importer](https://github.com/geosolutions-it/geonode-importer/blob/4911c526041d4887b7ba82955f73d1ef258bb5db/importer/apps.py#L22-L34). In this way any changes to the URLs (adding, delete, update) does not require a change in the core

geonode/maps/views.py
```diff
@@ -572,6 +581,15 @@ def map_metadata_detail(request, mapid, template="maps/map_metadata_detail.html"
             group = None
     site_url = settings.SITEURL.rstrip("/") if settings.SITEURL.startswith("http") else settings.SITEURL
     register_event(request, EventType.EVENT_VIEW_METADATA, map_obj)
+
+    # Load metadata_records for contrib apps
+    if getattr(settings, "EXTRA_METADATA_ENABLED", False):
```
mattiagiupponi, line 586: I have two thoughts about this:
1) I would evaluate having a function that puts in common this part of the code since it is repeated many times
2) is it actually needed to force the instance to have an additional field? the metadata itself as a json can be customized by changing the json structure in the settings:
https://github.com/GeoNode/geonode/blob/cb8d2aac6dae90ab61a266d902b5b1a38f957431/geonode/settings.py#L2161-L2167
It here looks to me like we are forcing the core to have a particular structure for an external library which creates a dependency. Instead by changing the structure of the settings you can "force" your own app to follow the schema that is declared. Saying this, the `extra_metadata` field and the `normal metadata field` can have the same structure shared between the core and the app and, in the end, is only the new template to define HTML as he wants (ps: the schema can change for each resource)

mattiagiupponi (changes requested):

t-book:
Thank you so much @mattiagiupponi !

> Unluckily putting the variable in that way makes the serializer initiated and the test ignores it (I don't know why).

This was driving me crazy! Also patching the settings did not change anything. Really strange.
I will go through your comments and update things!

t-book:
Hello @giohappy  and @mattiagiupponi ,

I have revised the PR with Mattia's comments. A lot of things have now been solved in the plugin, which has greatly simplified the PR.

Here is a short overview of changed files:

## Serializers
As all (documents, maps, layers, apps,) inherit from `ResourceBaseSerializer` which can provide the metadata,
I would just allow the children to use this field:

https://github.com/GeoNode/geonode/blob/62353d1c067f8956f4f9566989755ec869be701f/geonode/documents/api/serializers.py#L61

My plugin then sets `?include[]=metadata` via a middleware to retrieve the metadata. Doing so the response json includes the extra_metadata that can be used in mapstores detail panel without any furter request. 

## Views
These changes allow to use the view with custom context/and templates. As an example:

https://github.com/GeoNode/geonode/blob/62353d1c067f8956f4f9566989755ec869be701f/geonode/documents/views.py#L296-L303

The logic for retreiving the `custom_metadata` (aka the form data) has been moved to the plugin.

_As an alternative, it was considered to override the view completely or possibly to solve the whole thing via context processors. However, more code in the plugin means that it is harder to keep up with the changes in the core. The small change could therefore be a good middle ground._

## Templates
The templates were changed to set an extra step in the wizard and to read panels.html via a variable.

## urls, settings
The previous changes to the URLs and settings have been undone. The plugin controls the necessary setup via apps.ready().

All commits have been squashed.

Thank a lot for your time!

mattiagiupponi (changes requested):
Hi @t-book 
Overall now looks good. Since we are going to change the API serializer, can you please add a test to each resource to be sure that the metadata field is returned only if it is added as a parameter in the URL?

t-book:
Thanks @mattiagiupponi will add tests for the api.

t-book:
buona sera @mattiagiupponi 

> can you please add a test to each resource

I've added the following unit tests:

1. geonode.maps.api.tests.MapsApiTests.test_extra_metadata_included_with_param
2. geonode.documents.api.tests.DocumentsApiTests.test_extra_metadata_included_with_param
3. geonode.geoapps.api.tests.GeoAppsApiTests.test_extra_metadata_included_with_param
4. geonode.layers.api.tests.LayersApiTests.test_extra_metadata_included_with_param

These are simple identical tests (per resource type) to check if the GET param updates the JSON response as expected.

#### Example

https://github.com/GeoNode/geonode/blob/9df5e35756d6d059747d131bff908c271cda31f4/geonode/documents/api/tests.py#L80-L89


As working on these test files I also ordered the imports accordian to PEP8

- Standard library imports come first, in alphabetical order
- Third-party imports come next, also in alphabetical order
- Local application/library specific imports come last, again in alphabetical order



Hopes the update is fine.
Thanks again for your review and time!

T.

P.S.: Where I'm a bit unsure is if it was okay to merge https://github.com/GeoNode/geonode/pull/10592/commits/62353d1c067f8956f4f9566989755ec869be701f
from upstream. I guess this should have been better been an _update with rebase_?

mattiagiupponi (approved):
LGTM, @giohappy can you give a check too?

t-book:
@afabiani @giohappy Any chance to give this a look? (sorry this should not sound pushy)

afabiani:
@t-book since you started from an external repository the backport bot won't be able to automatically create the backport PR. Could you maybe take care of this against the 4.1.x branch?

afabiani:
@t-book nevermind, the bot was able to create the PR automatically

Issue #10746, opened 2023-03-07
Setup rate limits for all_auth


t-book:
**Is your feature request related to a problem? Please describe.**
Setup rate limits for several account actions

**Describe the solution you'd like**
Use the build in functionality of all_auth
https://django-allauth.readthedocs.io/en/latest/configuration.html?highlight=ACCOUNT_RATE_LIMITS#configuration

**Describe alternatives you've considered**
Do nothing

giohappy:
@t-book default settings are already configured, as the documentation says and as can be seen [here](https://github.com/pennersr/django-allauth/blob/0.51.0/allauth/account/app_settings.py#L175).
Are you suggesting to tune them differently?

giohappy:
Closing as part of a repository cleanup: this issue was last updated before 2025-01-01. Not a judgment on the report itself — feel free to reopen or file a fresh issue against `main` if it's still relevant. Thanks for the report.

PR #10748, opened 2023-03-07
[Fixes: #10746] Setup rate limits for all_auth

t-book:
Rate limits for all_auth. see #10746 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -334,7 +334,9 @@ STATICFILES_FINDERS = (
 CACHES = {
     # DUMMY CACHE FOR DEVELOPMENT
     "default": {
-        "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+        "django.core.cache.backends.dummy.DummyCache"
+        if DEBUG
+        else "django.core.cache.backends.locmem.LocMemCache"
     },
     # MEMCACHED EXAMPLE
     # 'default': {
@@ -2257,3 +2259,22 @@ SUPPORTED_DATASET_FILE_TYPES = [
         "needsFiles": ["shp", "prj", "dbf", "shx", "csv", "tiff", "zip", "xml"],
     },
 ]
+
+"""
+These all_auth rate limits apply if DEBUG is False
+5/m f.e. means 5 times per minute
+"""
+ACCOUNT_RATE_LIMITS = {
+    # Change password view (for users already logged in)
+    "change_password": "5/m",
+    # Email management (e.g. add, remove, change primary)
+    "manage_email": "10/m",
+    # Request a password reset, global rate limit per IP
+    "reset_password": "2/m",
+    # Rate limit measured per individual email address
+    "reset_password_email": "5/m",
+    # Password reset (the view the password reset email links to).
+    "reset_password_from_key": "20/m",
+    # Signups.
+    "signup": "20/m",
+}
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -334,7 +334,7 @@ STATICFILES_FINDERS = (
 CACHES = {
     "default": {
         # DUMMY CACHE FOR DEVELOPMENT
-        'BACKEND': "django.core.cache.backends.dummy.DummyCache"
+        "BACKEND": "django.core.cache.backends.dummy.DummyCache"
         if DEBUG
         else "django.core.cache.backends.locmem.LocMemCache"
     },
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -332,9 +332,9 @@ STATICFILES_FINDERS = (
 )
 
 CACHES = {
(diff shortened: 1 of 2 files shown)


geonode/settings.py
```diff
@@ -334,7 +334,9 @@
 CACHES = {
     # DUMMY CACHE FOR DEVELOPMENT
     "default": {
-        "BACKEND": "django.core.cache.backends.dummy.DummyCache",
```
afabiani, line 337: I guess you removed the `BACKEND` key here erroneously

giohappy:
@t-book default rate limits should be already provided by django_allauth (see my comment https://github.com/GeoNode/geonode/issues/10746#issuecomment-1460154081).

@afabiani @t-book the changing the cache backend is quite a critical component. Are we sure we should do it so light-hearted? I think it will require extensive testing.

geonode/settings.py
t-book: Oh yes! Thanks @afabiani

t-book:
@giohappy Yes definitly. This change should be tested carefully. Maybe let's keep it as draft:

> default rate limits should be already provided by django_allauth 

https://github.com/pennersr/django-allauth/blob/7c50e111078e481aa65804f88f2366a3c013b909/allauth/account/app_settings.py#L175-L191

From what I know this will not work out of the box:

> In order to be secure out of the box various rate limits are in place. The rate limit mechanism is backed by a Django cache. **Hence, rate limiting will not work properly if you are using the DummyCache.**

https://django-allauth.readthedocs.io/en/latest/configuration.html?highlight=cache#configuration

afabiani:
> @t-book default rate limits should be already provided by django_allauth (see my comment [#10746 (comment)](https://github.com/GeoNode/geonode/issues/10746#issuecomment-1460154081)).
> 
> @afabiani @t-book that changing the cache backend is quite an important change. Are we sure we should do it so light-hearted? I think it will require extensive testing.

Agree, it is also not recommended to use `LocMemCache` in a production system (as far as I know)

t-book:
> Agree, it is also not recommended to use LocMemCache in a production system (as far as I know)

The downsides I know of:

- It is not very memory efficient
- as it stores data in memory, data will be lost on a server restart 
- and it is not scalable across multiple servers

On the other hand I'm unsure if adding Redis or Memcached to the stack is really needed for this task. 
For sure it is good to think about / discuss it.

t-book:
I've removed the RATE_LIMITS from settings.py as default values are fine.
Further I will also test this with LDAP enabled ...

t-book:
@giohappy I've seen we already install memcached
https://github.com/GeoNode/geonode/blob/a4a21b72b0ceb6ebf60d397078d0b42892869396/Dockerfile#L45
So this means we could do the following to setup the rate limits:
- Enable memcached for production which allows all_auth to use the rate limits (caching backend needs to be tested also contrib app like risks which looks like using fe caching)
- add the error 429.html template provided by this PR

giohappy:
@t-book AFAIR caching was disabled because it was giving problems with DRF. Before re-enabling it we would need to test the side effects thoroughly.

You know, caching with such a dynamic API (based on current user, current permissions, etc.) is not an easy task. It will have to be managed view by view, taking care to not make the cache kick in in places where it shouldn't.

t-book:
@giohappy you are absolutely right! This step should be treated with caution. Sadly all_auth relies on caching for its limits. Another option might be creating a custom middleware that blocks DOS attacks for mailing endpoints. I will think about it.

t-book:
In case the caching is no option we could add a small app to contrib_apps that prevents email flooting and relies on the DB. Untested but to show the idea ...

model
```python
from django.db import models

class RequestTracker(models.Model):
    ip_address = models.CharField(max_length=255)
    url = models.CharField(max_length=255)
    timestamp = models.DateTimeField(auto_now_add=True)
```

middleware
```python
from django.core.exceptions import PermissionDenied
from django.utils import timezone
from .models import RequestTracker

class RequestLimiterMiddleware:
    def __init__(self, get_response):
        self.get_response = get_response

    def __call__(self, request):
        if request.method == 'POST' and request.path in ['/account/signup/', '/account/password/reset/', '/account/login/']:
            ip_address = request.META.get('HTTP_X_FORWARDED_FOR') or request.META.get('REMOTE_ADDR')
            current_time = timezone.now()
            one_minute_ago = current_time - timezone.timedelta(minutes=1)
            num_requests = RequestTracker.objects.filter(ip_address=ip_address, url=request.path, timestamp__gte=one_minute_ago).count()
            if num_requests >= 5:
                raise PermissionDenied('Too many requests in one minute')
            RequestTracker.objects.create(ip_address=ip_address, url=request.path)
        response = self.get_response(request)
        return response
```

PR #10842, opened 2023-03-23
Enumerate unzipped files

ridoo:
The last base_file candidate overrides previously ones. Also, different files with the same extension (file1.csv, file2.csv) are not be accessible from file_paths as they get overridden, too.

An enumerated list of unzipped files should resolve this. 

## Checklist

For all pull requests:


ridoo:
Seems that one test fails (`test_geoserver_proxy_strip_paths (geonode.geoserver.tests.test_helpers.HelperTest) ... FAIL`) according to restrictions on `/geoserver/rest/layers`. Not sure right now, if this relates to the adjustments made in geoservers `rest.properties` ...

t-book:
A little improvement might be to use a context manager (with ...)

https://github.com/GeoNode/geonode/blob/d10741f8015d50afe4fe3bef8eaadf228fe78dc4/geonode/storage/data_retriever.py#L208

gannebamm:
@afabiani I see the same tests failing for recent PRs e.g. the dependabot bumping. Can you confirm @ridoo idea? Is this due to the new rest security rules?

gannebamm:
> @afabiani I see the same tests failing for recent PRs e.g. the dependabot bumping. Can you confirm @ridoo idea? Is this due to the new rest security rules?

@ridoo please see #10887. If the issue gets resolved restart the tests to see if this will run successfully

ridoo:
@mattiagiupponi thanks for having a look at this. 

It seems to be unclear to me, what is actually acceptable content of an uploaded zip file. However, what about a simple sort before doing the whole magic (see my commit https://github.com/GeoNode/geonode/pull/10842/commits/573c3dc6b51c96df2b1dbd8a25fe61d0e7d74d27)

mattiagiupponi:
> It seems to be unclear to me, what is actually acceptable content of an uploaded zip file. However, what about a simple sort before doing the whole magic (see my commit [573c3dc](https://github.com/GeoNode/geonode/commit/573c3dc6b51c96df2b1dbd8a25fe61d0e7d74d27))

In GeoNode one zip contains just 1 layer (relation 1-1), but here since we are introducing the idea to have possible multiple `base_file` is important to keep the consistency with the previous behavior, anyway I don't think that a sorted could be enough I guess it would be safer to have a code which can group directly by file name instead of relying on the sorting function.

(IMO) A good way would be to:
- scan the zip file
- group the file available in the zip by name
- based on the file extension, verify that the needed file for that resource is available (this information is available via the [get_supported_datasets_file_types](https://github.com/GeoNode/geonode/blob/9eb871d2e23a606b5b3e7ea783d57937a4d3b8f0/geonode/utils.py#L1904-L1923)
- Return the first completed set found

@giohappy what do you think?

ridoo:
Hey @mattiagiupponi do you see any chance to see this PR merged into an upcoming release (4.2.x or even 4.1.x)?

If you need further input from our side, please let me know! :sun_with_face:

giohappy:
@ridoo sorry for the late reply. We can plan to have it in 4.2.0, but I think we need more thinking to implement a clean solution for the grouping of files.

ridoo:
@giohappy no worries .. are you going to share the discussion on this ticket or elsewhere?

ridoo:
@giohappy do you have any updates on this one? If you need further clarifications or work on my side please let me know

giohappy:
@ridoo from what I see your proposal doesn't work in several ways:

#### 1. We don't support multiple base_files
Every single (multi)file to be handled by the importer must have a `base_file` key. Outside the issue of sorting on grouping, you end up with a data structure where you have 1 `base_file` and multiple `shp_file`, `shx_file`, etc. 
The last "group" of files that are processed [overwrite](https://github.com/GeoNode/geonode/blob/573c3dc6b51c96df2b1dbd8a25fe61d0e7d74d27/geonode/storage/data_retriever.py#L218) the single available base_file value. I suspect this is what happens in the screenshot from @mattiagiupponi:

image

I suspect that [your test](s) is a false positive because you only take one group of files (the ones that generate the shp_file, etc. keys without postfixes. If you run the test on all the other files in [_files](https://github.com/GeoNode/geonode/blob/573c3dc6b51c96df2b1dbd8a25fe61d0e7d74d27/geonode/storage/tests.py#L619) you would get an assertion error since it wouldn't find a matched `base_file`

#### 2. The importer can only handle one multi(file) at a time
The importer creates one (multi)format at a time. The [create](https://github.com/GeoNode/geonode-importer/blob/master/importer/api/views.py#L84) method triggers only one handler. If the zip file you send contains multiple files, it will [get the handler](https://github.com/GeoNode/geonode-importer/blob/master/importer/api/views.py#L118) only for the single `base_file` inside the retrieved paths.
So, in the end, the additional files would be discarded.
For the problem at point 1, you might end up with a base_file file that doesn't match with the other unprefixed keys.

Multifile support must be rethought IMHO, both at the `StorageManager` and the `ImporterViewSet` level

ridoo:
> @ridoo from what I see your proposal doesn't work in several ways:
> 
> #### 1. We don't support multiple base_files
> Every single (multi)file to be handled by the importer must have a `base_file` key. Outside the issue of sorting on grouping, you end up with a data structure where you have 1 `base_file` and multiple `shp_file`, `shx_file`, etc. The last "group" of files that are processed [overwrite](https://github.com/GeoNode/geonode/blob/573c3dc6b51c96df2b1dbd8a25fe61d0e7d74d27/geonode/storage/data_retriever.py#L218) the single available base_file value. I suspect this is what happens in the screenshot from @mattiagiupponi:
> 
> image


Well, the multi-mixed content example was one of @mattiagiupponi 's tests. That is why I was asking on the criteria, what actually is acceptable content of a zip file. If I understand you correctly, only one file assemble (e.g. shp, shx, prj, dbf) is allowed. This would be ok to me, and does not interfere with the issue I tried to describe.

> I suspect that [your test](s) is a false positive because you only take one group of files (the ones that generate the shp_file, etc. keys without postfixes. If you run the test on all the other files in [_files](https://github.com/GeoNode/geonode/blob/573c3dc6b51c96df2b1dbd8a25fe61d0e7d74d27/geonode/storage/tests.py#L619) you would get an assertion error since it wouldn't find a matched `base_file`


Well, I do not mind (for this issue) if one file assemble overrides the other. What I tried to describe is the fact, that different handlers may decide on different base_files. In my case, my upload handler has a `json` file as `base_file` and multiple `csv` files as related files. See my comment: https://github.com/GeoNode/geonode/pull/10842#issuecomment-1482658050


> Multifile support must be rethought IMHO, both at the `StorageManager` and the `ImporterViewSet` level

I agree here, but as said, this was not the intent of this issue.

giohappy:
@ridoo I admit I lost track of this issue by reading only the comments. I've read the initial description again and now things are much clearer :grinning:.

Okay, so the goal was to not break the upload in case multiple files are put inside the uploaded zip file and let the handlers manage the enumerated paths.

BTW, even taking into account only specific cases, your patch would support only a very specific case: you have a custom handler registered at the top of the handlers registry.

Example: 
You have a zip containing 1 json file + many csv files. Both `json` and `csv` are good candidates for the `base_file`. The base_file selection would depend on the (lexicographical) order, am I wrong?

So, when the `ImporterViewSet` selects the right handler, it do it by [asking each handler](https://github.com/GeoNode/geonode-importer/blob/master/importer/orchestrator.py#L47-L50) if it can handle the fileset, and so the first one in the list will be selected, and with the factory handlers, it could either the `geojson` or the `csv` handler.

If your custom handler is at the top of the handlers registry, it might return immediately because you know how to handle that specific fileset.

If this is your expected scenario that's fine, although I don't think we can control the ordering of the registry at the moment.

ridoo:
> Okay, so the goal was to not break the upload in case multiple files are put inside the uploaded zip file and let the handlers manage the enumerated paths.

That's right. 
 
> Example: You have a zip containing 1 json file + many csv files. Both `json` and `csv` are good candidates for the `base_file`. The base_file selection would depend on the (lexicographical) order, am I wrong?

To my understanding right now, the unpack process is just making a best guess on the base_file, depending on what `geonode.utils.get_supported_datasets_file_types()` returned first item is. This is ok, as I am free to re-set the base_file in my handler.
 
> So, when the `ImporterViewSet` selects the right handler, it do it by [asking each handler](https://github.com/GeoNode/geonode-importer/blob/master/importer/orchestrator.py#L47-L50) if it can handle the fileset, and so the first one in the list will be selected, and with the factory handlers, it could either the `geojson` or the `csv` handler.
> 
> If your custom handler is at the top of the handlers registry, it might return immediately because you know how to handle that specific fileset.

This is the way I have it configured and it works.

> If this is your expected scenario that's fine, although I don't think we can control the ordering of the registry at the moment.

Ok, perhaps this issue targets multiple things here. The main issue, actually, is that unpacking files overrides files having the same extensions. This is why the PR is calle "Enumerate unzipped files", ie. two csv files in a zip would be unpacked to `files[csv_1]` and `files[csv_2]`

giohappy:
> This is the way I have it configured and it works.

@ridoo ok, the ordering of the handlers in the settings is respected [when they're registered](https://github.com/GeoNode/geonode-importer/blob/master/importer/handlers/apps.py#L22-L30)

@mattiagiupponi to me this PR is harmless. It doesn't break the current behavior (which is undefined behavior in case the zip contains multiple filesets) and it offers support to handlers that can handle multiple file sets.

mattiagiupponi:
> @mattiagiupponi to me this PR is harmless. It doesn't break the current behavior (which is undefined behavior in case the zip contains multiple filesets) and it offers support to handlers that can handle multiple file sets.

i agree

@ridoo can you please fix the flake8 and black issues? So we can have also the greenline from CircleCI.
Flake8 is complaining about this:
```
^@^@All done! ✨ 🍰 ✨
503 files would be left unchanged.
geonode/storage/data_retriever.py:224:39: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
geonode/storage/data_retriever.py:231:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```
I guess replacing the `type(existing) == list` with `isinstance(existing, list)` should be enough

ridoo:
Oh yes, will do this next

mattiagiupponi:
Hi @ridoo, a small test to be sure that works as expected it would be nice

ridoo:
@mattiagiupponi please have a look at my (really simple) test on enumeration/indexing multiple files with the same extension.

I just added a copy of zipped example.csv file. As ordering is undefined, I changed an existing test as well and commented where appropriate.

mattiagiupponi (approved):
@ridoo it looks good to me

Issue #8814, opened 2022-02-16
Configure SOCIALACCOUNT_LOGIN_ON_GET to make social logins work again


giohappy:
As a temporary workaround for the changes [introduced in django-allauth 0.47.0](https://github.com/pennersr/django-allauth/commit/4f9efa9e1d63e61a4f6cd067ef6c34b5343ef4fa) this setting must be set to `True`, otherwise the login flows won't work.

See https://github.com/GeoNode/geonode/issues/8806

Issue #11185, opened 2023-06-21
GNIP 94: Generic and pluggable OIDC SocialAccount Provider for GeoNode


afabiani:
# GNIP 94: Generic and pluggable OIDC SocialAccount Provider for GeoNode

## Overview

Currently GeoNode provides by default 2 quite old and outdated `SOCIALACCOUNT` providers, `LinkedIn` and `Facebook`, which are based on `OAuth2` and make use of very old plugins.

Today the reference protocol is `OIDC`, and GeoNode users increasingly need to be able to link their accounts to providers that support and implement this technology. Two of the most commonly used providers are Google and Microsoft Azure, for example.

The purpose of this proposal is to revise the `SocialAccount` and `SocialProvider` classes in GeoNode in order to not only make them capable of handling the more modern OIDC protocol but also to take advantage of some of its features, such as extracting user information from the `id_token` after validating its origin, among others.

Furthermore, a significant benefit of this technology is that the protocol is now a widely respected standard among all providers. This allows us to create dynamic, modular, and scalable structures that enable GeoNode to handle almost all use cases through a few configuration parameters.

### Proposed By

@afabiani 

### Assigned to Release

This proposal is for GeoNode 4.1.2+.

### State

* [x] Under Discussion
* [x] In Progress
* [x] Completed
* [ ] Rejected
* [ ] Deferred

## The `geonode_openid_connect` social account provider
This will be the generic class which will allow us handling the OIDC social login

```python
class GenericOpenIDConnectProvider(OAuth2Provider):
    id = "geonode_openid_connect"
    name = getattr(settings, "SOCIALACCOUNT_PROVIDERS", {}).get(PROVIDER_ID, {}).get("NAME", "GeoNode OpenIDConnect")
    account_class = import_class_module(
        getattr(settings, "SOCIALACCOUNT_PROVIDERS", {})
        .get(PROVIDER_ID, {})
        .get(
            "ACCOUNT_CLASS",
            "geonode.people.socialaccount.providers.geonode_openid_connect.provider.GenericOpenIDConnectProviderAccount",
        )
    )

    def get_default_scope(self):
        scope = getattr(settings, "SOCIALACCOUNT_PROVIDERS", {}).get(PROVIDER_ID, {}).get("SCOPE", "")
        return scope

    def get_auth_params(self, request, action):
        ret = super(GenericOpenIDConnectProvider, self).get_auth_params(request, action)
        if action == AuthAction.REAUTHENTICATE:
[...]

PR #11218, opened 2023-07-04
GNIP 94: Generic and pluggable OIDC SocialAccount Provider for GeoNode

afabiani:
References: https://github.com/GeoNode/geonode/issues/11185

## TODO
- [x] Extend the GNIP in order to recognize an optional attribute to set the `manager` role for a user at login time 
- [x] Documentation
- [x] Unit Tests

## Checklist

For all pull requests:


giohappy (changes requested):
@afabiani some tests failing

geonode/settings.py
```diff
@@ -1956,56 +1955,64 @@ def get_geonode_catalogue_service():
...
+    "COMMON_FIELDS": {"email": "email", "last_name": "family_name", "first_name": "given_name"},
+    "ACCOUNT_CLASS": "allauth.socialaccount.providers.google.provider.GoogleAccount",
+    "ACCESS_TOKEN_URL": "https://oauth2.googleapis.com/token",
+    "AUTHORIZE_URL": "https://accounts.google.com/o/oauth2/v2/auth",
+    "ID_TOKEN_ISSUER": "https://accounts.google.com",
+    "OAUTH_PKCE_ENABLED": True,
+}
+
+SOCIALACCOUNT_PROVIDERS = {
+    "geonode_openid_connect": _AZURE_SOCIALACCOUNT_PROVIDER,
```
giohappy, line 2011: @afabiani why aren't we including also the `_GOOGLE_SOCIALACCOUNT_PROVIDER` here, if Azure is configured by default?

geonode/utils.py
```diff
@@ -17,10 +17,10 @@
 #
 #########################################################################
 
-import itertools
 import os
 import gc
 import re
+import imp
```
giohappy, line 23: @afabiani the `imp` module is deprecated in favor of `importlib`, which is already used in several places in GeoNode.

geonode/settings.py
afabiani: @giohappy you need to choose one or another in this implementation. I rfactored this a bit so it is a bit more clear

giohappy (commented):
@afabiani could we add some tests?
We might take inspiration from the allauth tests for the single providers, like [this](https://github.com/pennersr/django-allauth/blob/main/allauth/socialaccount/providers/google/tests.py)

giohappy:
This would also fix https://github.com/GeoNode/geonode/issues/8814

giohappy:
@afaI see that the extraction and promotion of the group manager have been implemented. However I don't see where a user is demoted in case at the next login we find that it's not a manager anymore.

giohappy:
ok, I see that if a token has the claim for the groups, that user will be removed from the groups at each login, and reassigned to the groups reported inside the claims.

This could be problematic if the user is also assigned to a group manually since the membership will be deleted at the next login (if the same user has the groups declared in its token).
**We will have to add a note on this inside the documentation** @afabiani

giohappy (approved):

Issue #11263, opened 2023-07-14
Link for downloading GeoLite2-City.tar.gz is not working


NyakudyaA:
# Problem

The link in https://github.com/GeoNode/geonode/blob/master/geonode/monitoring/management/commands/updategeoip.py#L30C64-L30C84

Does not resolve.

![image](https://github.com/GeoNode/geonode/assets/2510900/3e06089b-5d15-4e07-aac1-99c29d1ac7a3)

# Solution
Fix the link or use alternative links with the DB

giohappy:
Closing as part of a repository cleanup: this issue was last updated before 2025-01-01. Not a judgment on the report itself — feel free to reopen or file a fresh issue against `main` if it's still relevant. Thanks for the report.

PR #11300, opened 2023-07-30
Add new env variable GEOIP_DATABASE_URL

NyakudyaA:
Fixes https://github.com/GeoNode/geonode/issues/11263

Even though you can a database and use the env variable `GEOIP_PATH=/mnt/volumes/statics/geoip.db` when deploying using CDI i.e Argo CD you have no option to mount volumes with a local DB


giohappy:
@NyakudyaA the `updategeoip` functionality was just dropped since it was a very old method that is not allowed anymore by MaxMind EULA.
I think the `GEOIP_PATH` should be enough to reference the DB location, isn't it?

giohappy:
>Not in all cases, in our case we are deploying through ArgoCD and we cannot mount the DB so we would prefer to have a working link to download the DB and if the link is not available/working we can still provide our own custom link to download the DB

Out of curiosity:

 - are you using the monitoring module?
 - do you need to execute regular updates of the GeoIP DB?

NyakudyaA:
> > Not in all cases, in our case we are deploying through ArgoCD and we cannot mount the DB so we would prefer to have a working link to download the DB and if the link is not available/working we can still provide our own custom link to download the DB
> 
> Out of curiosity:
> 
>     * are you using the monitoring module?
> 
>     * do you need to execute regular updates of the GeoIP DB?

We do use monitoring module. We rarely execute updates on the GeoIP DB. If the `updategeoip` method has been dropped maybe I can take a look and see if we still need this. Maybe if you can fix the URL to be correct then we can close this PR

giohappy:
>Maybe if you can fix the URL to be correct then we can close this PR

@NyakudyaA the GeoIP 2 DB can not be shared, it violates MaxMind EULA. You need your own copy of the DB and set the GEOIP_PATH to your convenience.
If you don't need to do frequent updates you could ship it inside your Docker image.

NyakudyaA:
No longer useful since monitoring has been dropped

PR #11308, opened 2023-07-31
Align to GeoNode master (4.2.0) and improve/fix docker initialization

afabiani:
This PR introduces improvements and fixes to docker-compose and initialization scripts in general:

- Removing some unused variables.
- Removing some unused and obsolete containers, such as Jenkins.
- Removing the incorrectly written and non-functioning part for setting the password of GeoServer.
- Inverting the startup dependency from GeoServer -> Django to Django -> GeoServer.
- Ensuring that OAUTH2 is properly initialized on restart in all its parts.
- Enhancing NGINX to make sure that GeoServer functions well behind any proxy, including HTTPS.

Some implementation details:

1. Environment file

`DOCKERHOST` and `DOCKER_HOST_IP` variables removed as useless and misleading. Instead, it exposed `NGINX_BASE_URL` which was calculated automatically and badly.

The variables
```
GEONODE_LB_HOST_IP=django
GEONODE_LB_PORT=8000
GEOSERVER_LB_HOST_IP=geoserver
GEOSERVER_LB_PORT=8080
```
are important, they are used by `NGINX` to create the upstream `proxyPass` and by GeoServer to create the internal `OAuth2` endpoints. They represent the name and port of the GeoNode and GeoServer containers. In general it is not necessary to change them, but in an environment like `Microsoft Azure` these must be set to the address of the `subnet` as the names are not internally resolved.

Increased `MAX_DOCUMENT_SIZE=200` because the previous default of `2Mb` did not allow rendering even one document.

Updated `create-envfile.py` to also enable `staging` mode to test `HTTPS` on `localhost`, which was previously totally ignored.

2. Dockerfiles

Removed useless stuff and started it from `geonode/geonode-base:latest-ubuntu-22.10`

3. Readme

Added a `Quick Start` section with `docker`

4. Docker Compose

Fixed healthchecks which are now faster and more efficient and switched Django -> GeoServer startup dependency, because it's the Django entrypoint that sets `OAutg2` parameters in the `GEOSERVER_DATA_DIR` before it starts.

Updated to `POSTgresql 15`

5. UWSGI + NGINX

Enabled `proxyPass` instead of `uwsgiPass` because we set `headers` correctly in `geonode.conf` only with `proxyPass`

6. GeoServer docker entry points

Fixed the `.sh` to set `OAuth2` which previously worked badly and only partially checking when there was the `localhost` value. Now they update everything and always.

7. nginx/geonode.conf.envsubst

Made `proxy passes` dynamic

`set $upstream $GEOSERVER_LB_HOST_IP:$GEOSERVER_LB_PORT;`

`set $upstream $GEONODE_LB_HOST_IP:$GEONODE_LB_PORT;`

Updated the GeoServer conf so that the GUI works fine behind any proxy.

8. A general reformat to the `black` code

9. Version bump

from `4.0.0.final` to `4.2.0.dev`


ridoo:
@afabiani as already pointed out these are welcomed changes.

What do you think about renaming the load balancer parameters, so that

```
set $upstream $GEOSERVER_LB_HOST_IP:$GEOSERVER_LB_HOST;
set $upstream $GEONODE_LB_HOST_IP:$GEONODE_LB_PORT;
```

becomes

```
set $upstream $LB_GEOSERVER_HOST:$LB_GEOSERVER_HOST;
set $upstream $LB_GEONODE_HOST:$LB_GEONODE_PORT;
```

This seems to be a minor change, but might get clearer the load balancer scope of the parameters. I also removed the `_IP` suffix as in other contexts you will not put an static IPs here (e.g. k8s).

giohappy (approved):

giohappy:
It's been already pushed.

rkhandha:
I have been testing new images that were pushed and I would like to give some feedback.  Please keep in mind I am new to geonode and kubernetes.

1. In image  geonode/geonode:latest-ubuntu-22.10,  I am no longer seeing blowfish deprecated error.  To run this in Openshift  as a non-root user I had to apply following permissions for my environment
    RUN touch $HOME/.override_env
    RUN ls -l $HOME/.override_env
    RUN chmod g+rwx -R /root 
    RUN chmod g+rwx -R /var/log
    RUN chmod g+rwx  /usr/src/geonode

2.  Creating configmap from the .env_dev, creating a service and Route, I was able to see the geonode screen, able to login and got admin screen, this is really a great progress. 
3. Geoserver is not working yet, I have deployed geoserver (image  geonode/geoserver:2.23.0)  as a container in the same pod, it is giving me 404 error Using the URLconf defined in geonode.urls, Django tried these URL patterns.  If I deploy geoserver as a separate pod, service, route I get an index page.  so not sure what is going on.  With image geonode/geoserver:2.23.1,  I do get geoserver page. 
4. With new image I can access geonode directly,  do I need to implement nginx

Thank you again appreciate all your help.

PR #11423, opened 2023-08-28
[Fixes #11421] Improvements for DownloadHandler

### LINKED ISSUE #11421: Improvements for DownloadHandler
As a follow-up of this issue https://github.com/GeoNode/geonode/issues/11296 we want to introduce a new set of improvement for the download handler. 
This improvements will help to make room for an API that supports multiple downloads (in the future)

### Datasets
- [x] Rename `DownloadHandler` to `DatasetDownloadHandler`, and move it under the `layers` module
- [x] The `DatasetDownloadHandler` implements an `is_ajax_safe` method which will return a boolean -> whether the `download_url` can be called with AJAX or not. For the default handler it will be:
  - `true` for `WPS`
  - `false` for all the other cases
- [x] Remove the logic for the creation of the "original" Link entries [here](https://github.com/GeoNode/geonode/blob/master/geonode/utils.py#L1485-L1508).
- [x] [Dataset.download_url](https://github.com/GeoNode/geonode/blob/master/geonode/layers/models.py#L329) will just return `build_absolute_uri`. The logic of whether to return any "original" Link (created manually or received from previous DBs) will be moved to the `DatasetDownloadHandler`.

### Documents

- [x] Also, documents must return whether AJAX is safe or not. I wouldn't bother too much here, since we have a convoluted download / link / href logic for documents. Maybe we can implement a `Document.download_is_ajax_safe` property returning `true` if the resource is LOCAL, `false` otherwise.

### API
- Implement a `download_urls` array property for the `ResourceBaseSerializer` with the following structure:
```
 "download_urls": [
   {
      "URL": .... , 
      "ajax_safe": true|false, 
      "default":true|false
   },
]	
```
- For the moment `download_urls` will contain only one entry;
   - For datasets, it will be populated with the properties obtained from the single `DatasetDownloadHandler`. **NOTICE** that for REMOTE resources (e.g. services) `download_urls` should return an empty array. Let's think if it will be the serializer to handle this case, or let `DatasetDownloadHandler` return the download_url object itself (which would be `None` for REMOTE datasets).
   - For documents, it will be populated from `Document.download_url` and  `Document.download_is_ajax_safe`
   - For the other resource types it will return an empty array
- Deprecate the `download_url` field.

mattiagiupponi:
ref #11421 
## Checklist

For all pull requests:

### DIFF
--- a/geonode/base/api/serializers.py
+++ b/geonode/base/api/serializers.py
@@ -320,7 +320,7 @@ class DownloadArrayLinkField(DynamicComputedField):
             if obj.download_url:
                 download_urls.append({"url": obj.download_url, "ajax_safe": obj.is_ajax_safe, "default": True})
             # then let's prepare the payload with everything
-            handler_list = get_dataset_download_handlers(additional_only=True)
+            handler_list = get_dataset_download_handlers()
             for handler in handler_list:
                 obj = handler(self.context.get("request"), _instance.alternate)
                 if obj.download_url:
--- a/geonode/base/api/serializers.py
+++ b/geonode/base/api/serializers.py
@@ -28,7 +28,9 @@ from django.forms.models import model_to_dict
 from django.contrib.auth import get_user_model
 from django.db.models.query import QuerySet
 from django.http import QueryDict
+from django.utils.module_loading import import_string
 
+from deprecated import deprecated
 from rest_framework import serializers
 from rest_framework_gis import fields
 from rest_framework.reverse import reverse, NoReverseMatch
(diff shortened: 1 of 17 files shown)


geonode/layers/download_handler.py
```diff
@@ -51,9 +51,39 @@ def get_download_response(self):
...
         """
         resource = self.get_resource()
+        if not resource:
+            raise Http404("Resource requested is not available")
         response = self.process_dowload(resource)
         return response
 
+    @property
+    def is_default(self):
+        return "DatasetDownloadHandler" in str(self)
```
giohappy, line 61: Probably the logic to select the "default" handler should be outside the handler itself.
The serializer should select the default handler. Something lilke:

- the serializer looks for a `settings.DEFAULT_DATASET_DOWNLOAD_HANDLER` with the path of the handler module
- if it's set, ok
- otherwise, it will select the first item from the `settings.DATASET_DOWNLOAD_HANDLERS` list

geonode/base/api/serializers.py
```diff
@@ -278,15 +280,51 @@ class DownloadLinkField(DynamicComputedField):
...
+        elif _instance.resource_type in ["document"]:
+            return [
+                {
+                    "url": _instance.download_url,
+                    "ajax_safe": _instance.download_is_ajax_safe,
+                }
+            ]
+        elif _instance.resource_type in ["dataset"]:
+            for item in settings.DATASET_DOWNLOAD_HANDLERS:
+                handler = import_string(item)
```
giohappy, line 318: IMHO the import should be done at import time, not inside the method call where it will be called each time the field is requested.

geonode/layers/download_handler.py
```diff
@@ -51,9 +51,35 @@ def get_download_response(self):
...
 
+    @property
+    def is_ajax_safe(self):
+        """
+        We use that USE_GEOSERVER boolean to decide if use the default
+        WPS download method or use the fallback on the proxy.
+        This means that by default if geoserver is used, the WPS
+        is used
+        """
+        return settings.USE_GEOSERVER
```
giohappy, line 67: @mattiagiupponi this assumption is wrong. AJAX is safe if we actually use WPS, but in case an "original" link is set we cannot ensure AJAX, since it could point to an external URL and the client should redirect the user to that URL.

giohappy (approved):

PR #11478, opened 2023-09-13
[Fixes #11477] Restore loopback port for memcached in local env files

### LINKED ISSUE #11477: Revert to using localhost for memcached inside development .env files
https://github.com/GeoNode/geonode/pull/8818 introduced the automatic creation of a `memcached` Docker service, with related configuations.
However, since `MEMCACHED_LOCATION` is also used by a non Docker setup, we should revert it to using the loopback address by default inside .env.dev and .env.local, otherwsie they won't work anymore outside Docker.
We can keep `memcached` location inside .env.sample since it'a meant to be used for Docker already (we use `db` and `geonode` hosts already, for example)

giohappy:


afabiani (approved):

giohappy:
do you agree @EHJ-52n?

EHJ-52n:
Having only one hour before merge is quite challenging. :sunglasses: 

Before answering, I would like to know the context, in which these to files are used?

giohappy:
>Having only one hour before merge is quite challenging. 

Sorry @EHJ-52n but we had to revert it quickly since it was going to break a couple of client.
These are used in a list of local environments for testing and development, where a real memcached service is not available.

PR #11565, opened 2023-10-05
[Fixes #11494] Implement relations between resources

### LINKED ISSUE #11494: GNIP 96: Implement relations between resources
# GNIP <num> - Implement relations between resources

## Overview

At the moment linked resources are only available:

 - between dataset and maps
 - between documents and datasets
 - between documents and maps
 
We want to create generic references between resources.

### Proposed By

- @etj - Emanuele Tajariol

### Assigned to Release

This proposal is for GeoNode 4.2.

### State

* [x] Under Discussion
* [x] In Progress
* [ ] Completed
* [ ] Rejected
* [ ] Deferred

### Motivation

- There is no way to find out which resources are linked to Dashboards or GeoStories
- Possibility to have links between multiple document translations 
- ...

### Requirements

We want to implement a generic model for relating any resource type taking into account the following:

 - The client should explicitly list the resources employed inside Dashboards and Geostories when they're saved. This is what we already do for MapLayers: the client includes the MapLayers payload when it saves a map, without the backend having to inspect the blob content
 - Users should be able to manually link additional resources to Maps, Dashboards, and Documents, beyond those set automatically by the client 

The linked_resources API endpoint will be extended to take into account the new link model.

### Proposal

We need a simple "source resource" --> "target resource" relationship between resources, in order to point out any basic coupling between such resources. 
At first sight, no other info is needed -- further fields are listed in the "future evolution" section.
We may have two different kinds of relationships:
- *internal*: 
  - resources that are linked by the client's inner logic (layers belonging to a map, documents used by a geoapp, ...)
  - Internal links may only be set using the API, leaving to the external app (or client app) the task to create or delete them.
- *external*: 
  - links created by the user, adding them to the metadata editing page.
  - The metadata editing page will only be able to view, add or delete "external" links.

The viewing pages may have panels listing the resources as:
- "This <resource|dataset|map|...> uses:" 
   - list of resources linked by internal links
      - items will have as info 
         - type (dataset, map, doc, ...)
         - title
         - link to the resource
- "This <resource|dataset|map|...> references:" 
   - list of resources linked by external links
- "This <resource|dataset|map|...> is referenced by:" 
   - list of resources that have this resource linked by either internal or external links

### Current status

At the moment we have two classes for setting relationships:
- [`MapLayer`](https://github.com/GeoNode/geonode/blob/4.1.2/geonode/maps/models.py#L227)
- [`DocumentResourceLink`](https://github.com/GeoNode/geonode/blob/4.1.2/geonode/documents/models.py#L153)

##### Payload

The `resources` entry should be replaced by:
- `linked_to`: very similar to `resources`, also adding the boolean `internal` attribute. It contains all the `LinkedResource` records where the current resource is the **source** object.
 - `linked_by`: It contains all the `LinkedResource` records where the current resource is the **target** object

The old `resource` entry will be first deprecated and then kept long enough to allow time for the client part to use the new data. 

##### Pagination
This entrypoint at the moment is paginated.

Since we'll have two different lists, the pagination would be confusing. The idea is to *remove the pagination* for this call (also, not so many linked resources are expected -- 640kresources ought to be enough for anybody ;) )
Also, the client at the moment is requesting `page=1&page_size=99999`, so pagination is not really used.

The pagination could be removed right away -- maybe by placing fake values in the items if they are needed to exist.

##### Filtering

Since we do expect few linked resources, filtering could be done client side when/if needed.

##### Example

See sample output in https://github.com/GeoNode/geonode/issues/11494#issuecomment-1744625374

#### Documents extents

Reading comments from older commits, it seems that a signal occurred when a map was saved, in order to update the extents of the documents linked to it. 
However, it seems to be a dead feature, since the logic supposedly implementing it does nothing.
We're not going to re-implement this feature since the dependency graph among the various resources may have loops that may cause infinite signals loop 

Class `DocumentResourceLink` 

Declare its Backwards Compatibility.

## Future evolution

- Relations info can be augmented with
  - Notes
  - [Relationship type](https://www.dublincore.org/specifications/dublin-core/rdf-relation-types/)
  - `client_owner`: a string representing the owner of the internal relation; for instance when a GeoStory references a dataset, the LinkedResource record would have `internal=true` and `client_owner='mapstore'`

## Feedback

## Voting

Project Steering Committee:

* Alessio Fabiani: 👍 
* Francesco Bartoli:
* Giovanni Allegri: +1
* Simone Dalmasso:
* Toni Schoenbuchner:
* Florian Hoedt: 👍 

## Links

Remove unused links below.

* [Email Discussion]()
* [Pull Request]()
* [Mail Discussion]()
* [Linked Issue]()

etj:
Implements changes described in #11494

## Checklist

For all pull requests:


geonode/base/api/serializers.py
```diff
@@ -787,7 +787,40 @@ class Meta:
...
 
 class SimpleResourceSerializer(DynamicModelSerializer):
+
     class Meta:
         name = "linked_resources"
         model = ResourceBase
         fields = ("pk", "title", "resource_type", "detail_url", "thumbnail_url")
+
+    def to_representation(self, instance: LinkedResource):
+        return {
```
mattiagiupponi, line 797: Evaluate to use the `super` instead of the plain json:

```python
def to_representation(self, instance: LinkedResource):
        data = super(PersonListSerializer, self).to_representation(instance)
        data.update({"title": f"{'>>> ' if instance.is_target else '<<< '} {instance.title}"})
        return data
```

In this way if we want to define other field we dont have to add them in the "to_representation"

mattiagiupponi (changes requested):

geonode/base/api/serializers.py
etj: Sure, anyway this class is going to be removed as soon as the client is aligned to use the new `linked_to` and `linked_by` elements. This is a quick and dirty implementation.

giohappy (dismissed):

mattiagiupponi (approved):

PR #12179, opened 2024-04-22
[Fixes #12124] GNIP 100: Assets

### LINKED ISSUE #12124: GNIP 100: Assets
# GNIP 100 - Assets

## Overview

We need a way to identify files (local, remote, in the cloud...) per se.
There's no way at the moment to identify data files by themselves, which are only referenced by the field `ResourceBase.files'.

Also, the StorageManager is pluggable, but only allows for a single storage backend at once.
By having different subclasses of `Asset` (e.g. `LocalAsset`, `S3Asset`, ...) we may have a GeoNode instance handling datafiles on different data store backends.

### Proposed By

- @etj - Emanuele Tajariol
 
### Assigned to Release

This proposal is for GeoNode 4.3 (?)

### State

* [x] Under Discussion
* [x] In Progress
* [ ] Completed
* [ ] Rejected
* [ ] Deferred

### Motivation

- Data may be stored on different backends
- Data is not coupled with a single ResourceBase
- May simplify the handling of basic data (images, etc) in GeoStories, which at the moment use full featured (and heavy) ResourceBases metadata also for simple files.
- Allows the possibility to link a single `ResourceBase` with multiple data files (think for instance about a Document having multiple PDF files for different languages).
- Allows the definition of a directory hierarchy as a single data asset, making it possible to publish complex data.
- [Simplifies/improves handling of authorization for data access](https://github.com/GeoNode/geonode/issues/12124#issuecomment-2032406597).

## Proposal

We introduce the concept of **Asset** as generic data, that may be linked to a ResourceBase.
A **LocalAsset** represents data stored in the filesystem (either a single file or a directory tree).

The `Asset` class will replace and augment the information stored at the moment in the `ResourceBase.files` field.

An Asset is associated with a Resource through a Link, which also tells the URL through which the Asset will be available to the GeoNode users. 
- For a LocalAsset, the URL may be a service that checks for authorization before returning the data. 
- For a RemoteAsset (we're not discussing its implementation in this proposal), it could be a replication of the RemoteAsset URL, or also a GeoNode local endpoint that proxies a remote content, maybe providing some authentication to the remote service -- this is just an example, we don't have any real use case at the moment.

### Other usages for assets

Since the Asset object is quite simple, we could use it for other purposes as well; for instance, at the moment we use "unadvertised" `ResourceBase` instances for providing simple data to GeoStories (images, PDFs, ...). Instead of using such a heavy object, we could just use LocalAssets for this purpose. 

Also, more Assets may be associated with an existing ResourceBase; this behavior replicates what GeoNetwork is already doing, that is having multiple data resources pointed by a single metadata record.

### Permissions

In the future there could be different permissions for a Resource and its linked Assets, anyway for the sake of simplicity, **as a first step we may grant on the asset the very same permissions of the linked ResourceBases**.

In the case we want to associate an Asset to more than one Resource, the Asset will be available if the user has download privileges on at least one of the associated Resources.

### Implementation

GeoNode asset diagram vpd

#### Model:
- Add `Asset` class, and its subclass `LocalAsset`
- Refact `ResourceBase` and `Link` classes

#### Logic:
- Replace the usage of `files` with `Asset` instances

#### DB migration:
- Documents:
  - migration is straightforward, since there's a single file
- Datasets:
  - Single file --> Asset
  - Shapefile (shp+dbf+shx+prj?) or other multifile formats--> ? 
     - Multiple options according to Asset type:
        - create a zipfile ? 
        - publish as the root entry of a directory? (*There are cases where we want an Asset to be a directory with subdir*)
        - Every Assethandler should be able to create its own Asset content and related link url
          - shp file should create a zip file, since such data is only used for download
    - What if there is a SLD (or any other satellite file)? is it a separate Asset?

#### API:
 - Make sure existing API still returns file info as before

#### Authorization

A user has access to an `Asset` data iff such Asset is associated with at least one ResourceBase for which the user has download permissions.

### Backwards Compatibility

- API: old `files` array can be preserved in output 

#### Decoupled uploads

A user may upload an Asset without having to associate it to a Resource.
Unassociated Assets may be used to automatically create ResourceBases and attach the asset to them.

#### Deprecate Documents

Once Assets gain their characterization, the Document object will not have much of a meaning, also considering that users upload as a Document any object that is not published as a Layer. 
This means that we will be able to remove the Document class, and convert its instances into ResourceBases with an Asset handling the former document's data.

#### Cleanup uploaded files

Some old installations have the uploaded data into `/data`.  
The recent importer stores the uploaded data into `.../STATIC_ROOT/uploaded`, and GeoServer publishes the geotiff from that directory
The final migration to Assets will store the files in `.../STATIC_ROOT/assets`, and GeoServer shall publish the files from there.
In order to clean up such obsolete setups, a migration script could be done that:
- moves the files from `uploaded/` to `assets/`
- updates the `LocalAsset.location` field 
- updates the GeoServer config to point to the new path
Such a migration script should only be run by hand since we do not know if such files are also used for other purposes outside GeoNode, so the admin can choose to run it or not.

## Feedback

## Voting

Project Steering Committee:

* Alessio Fabiani:
* Francesco Bartoli:
* Giovanni Allegri: +1
* Toni Schoenbuchner: +1
* Florian Hoedt:

## Links

Remove unused links below.

* [Email Discussion]()
* [Pull Request]()
* [Mail Discussion]()
* [Linked Issue]()

etj:
First implementation of assets.

Also these changes are needed: https://github.com/GeoNode/geonode-importer/pull/235

## Checklist

For all pull requests:


geonode/base/models.py
```diff
@@ -2127,3 +2128,50 @@ class GroupGeoLimit(models.Model):
 class ExtraMetadata(models.Model):
     resource = models.ForeignKey(ResourceBase, null=False, blank=False, on_delete=models.CASCADE)
     metadata = JSONField(null=True, default=dict, blank=True)
+
+
+class AssetManager(PolymorphicManager):
+    pass
+
+
+class Asset(PolymorphicModel):
```
mattiagiupponi, line 2137: If we have a submodule `assets` i guess this should be defined there

geonode/base/models.py
```diff
@@ -2127,3 +2128,50 @@ class GroupGeoLimit(models.Model):
 class ExtraMetadata(models.Model):
     resource = models.ForeignKey(ResourceBase, null=False, blank=False, on_delete=models.CASCADE)
     metadata = JSONField(null=True, default=dict, blank=True)
+
+
+class AssetManager(PolymorphicManager):
+    pass
```
mattiagiupponi, line 2134: What is the usage of having an empty manager?

geonode/base/models.py
```diff
@@ -2127,3 +2128,50 @@ class GroupGeoLimit(models.Model):
...
+        return super().__str__()
+
+
+def cleanup_asset_data(instance, *args, **kwargs):
+    from geonode.assets.handlers import asset_handler_registry
+
+    asset_handler_registry.get_handler(instance).remove_data(instance)
+
+
+signals.post_delete.connect(cleanup_asset_data, sender=LocalAsset)
```
mattiagiupponi, line 2177: If you connect the signal to "Asset" it should work without defining (in the future) one for each asset

```
signals.post_delete.connect(cleanup_asset_data, sender=Asset)
```

geonode/base/models.py
```diff
@@ -1278,10 +1276,12 @@ def instance_is_processed(self):
 
     @property
     def is_copyable(self):
-        from geonode.geoserver.helpers import select_relevant_files
-
         if self.resource_type == "dataset":
-            allowed_file = select_relevant_files(get_allowed_extensions(), self.files)
+            from geonode.assets.utils import get_default_asset
+            from geonode.geoserver.helpers import select_relevant_files
```
mattiagiupponi, line 1281: Can we just move on top this imports? This API si widely called. Having this here means re-import them everytime. They are cached, yes but part of them can be still re-imported:
https://stackoverflow.com/a/46662587/7597536

geonode/documents/api/views.py
```diff
@@ -119,23 +121,28 @@ def perform_create(self, serializer):
...
             if doc_url:
                 payload["doc_url"] = doc_url
                 payload["sourcetype"] = enumerations.SOURCE_TYPE_REMOTE
 
             resource = serializer.save(**payload)
 
+            if file:
+                manager = StorageManager(remote_files={"base_file": file})
+                manager.clone_remote_files()
+                create_asset_and_link(
```
mattiagiupponi, line 133: Instead of cloning the remote files and then copy them again in the `settings.ASSET_ROOT`, whey dont we improve the `clone_remote_files` by giving a destination path? In this way we can avoid to perform 3 operations on the file.
In geonode deployment with a huge quanity of document, this file system work can be an issue

geonode/documents/tasks.py
```diff
@@ -104,15 +106,24 @@ def create_document_thumbnail(self, object_id):
...
-    if document.files:
-        doc_path = storage_manager.path(document.files[0])
+
+    # get asset of the resource
+    asset = get_default_asset(document)
+    if not asset and not document.doc_url:
+        raise Exception("Document has neither an associated Asset nor a link, cannot generate thumbnail")
+
+    if asset:
+        ah = asset_handler_registry.get_handler(asset)
```
mattiagiupponi, line 116: better to be lil more explicit with the names, it makes the read easier

geonode/resource/manager.py
```diff
@@ -523,12 +544,12 @@ def copy(
...
-                    to_update = {}
-                    try:
-                        to_update = storage_manager.copy(_resource).copy()
-                    except Exception as e:
-                        logger.exception(e)
 
+                    links = self._copy_data(instance, target=_resource)
+                    files = []
+                    for link in links:
+                        files.extend(link.asset.location)
```
mattiagiupponi, line 551: cant' we just return the location from the `_copy_data` function instead of the whole object? If yes, this `for` is not needed

geonode/base/api/views.py
```diff
@@ -1525,3 +1528,71 @@ def base_linked_resources_payload(instance, user, params={}):
...
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(serializer.data)
+
+    def _get_file(self, request, pk, attachment: bool):
+        asset = get_object_or_404(Asset, pk=pk)
+        if bad_response := get_perms_response(request, asset):
```
mattiagiupponi, line 1576: can't this be replaced by the `IsOwnerOrAdmin` permission class? If yes, let's improve the perms class (if needed)

mattiagiupponi (changes requested):

geonode/resource/manager.py
etj: adding as comment
```python
# we're just merging all the files together: it won't work once we have multiple assets per resource
# TODO: get the files from the proper Asset
```

geonode/base/models.py
etj: that would create a circular reference error

geonode/base/models.py
etj: just a temporary placeholder

geonode/assets/utils.py
```diff
@@ -0,0 +1,120 @@
...
+    if visibile_res.exists():
+        logger.debug("Asset: access allowed by Resource")
+        return None
+    elif user and user.is_authenticated:
+        return HttpResponse(status=403)
+    else:
+        return HttpResponse(status=401)
+
+
+def get_default_asset(resource: ResourceBase) -> Asset or None:
```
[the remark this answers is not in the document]
etj: It returns an Asset, not an AssetHandler.

The first one is the default default bc at the moment we have only a single Asset.
When we'll have multiple assets per resource there will be a way to choose the default one (highest priority?, a pluggable selector?)
`get_resource_asset` would imply there is only one asset per resource

geonode/documents/api/views.py
etj: keeping the current behaviour so far

PR #12254, opened 2024-05-21
[Fixes #12253] Improvements to the proxy view

### LINKED ISSUE #12253: Improvements to the proxy service
The following changes will be implemented for the proxy view:

 - Avoid extending the proxy whitelist on every request
 - Accept requests that match a configurable list of string inside the URL path or the querystring. Two new (tuple) settings will be introduced:
   - `PROXY_ALLOWED_PARAMS_NEEDLES`
   - `PROXY_ALLOWED_PATH_NEEDLES`

giohappy:
## Checklist

For all pull requests:

### DIFF
--- a/geonode/proxy/views.py
+++ b/geonode/proxy/views.py
@@ -73,7 +73,6 @@ def proxy(
     url=None,
     response_callback=None,
     sec_chk_hosts=True,
-    sec_chk_rules=True,
     timeout=None,
     allowed_hosts=[],
     headers=None,
@@ -109,23 +108,12 @@ def proxy(
             ):
                 proxy_urls_registry.register_host(url.hostname)
 
-        if url.hostname not in proxy_urls_registry.get_proxy_allowed_hosts():
-            # Check Remote Services base_urls
-            for _s in Service.objects.all():
-                _remote_host = urlsplit(_s.base_url).hostname
-                proxy_urls_registry.register_host(_remote_host)
-
         if not validate_host(extract_ip_or_domain(raw_url), proxy_urls_registry.get_proxy_allowed_hosts()):
             return HttpResponse(
                 "The path provided to the proxy service" " is not in the PROXY_ALLOWED_HOSTS setting.",
                 status=403,
                 content_type="text/plain",
             )
-
-    # Security checks based on rules; allow only specific requests
-    if sec_chk_rules:
-        # TODO: Not yet implemented
-        pass
 
     # Collecting headers and cookies
     if not headers:
--- a/geonode/proxy/views.py
+++ b/geonode/proxy/views.py
@@ -110,7 +110,7 @@ def proxy(
 
         if not validate_host(extract_ip_or_domain(raw_url), proxy_urls_registry.get_proxy_allowed_hosts()):
             return HttpResponse(
-                "The path provided to the proxy service" " is not in the PROXY_ALLOWED_HOSTS setting.",
+                "The path provided to the proxy service is not allowed.",
                 status=403,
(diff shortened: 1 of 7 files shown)


geonode/proxy/views.py
```diff
@@ -104,52 +101,24 @@ def proxy(
...
-                len(ows_tokens) == 4
-                and "version" == ows_tokens[0]
-                and StrictVersion(ows_tokens[1]) >= StrictVersion("1.0.0")
-                and StrictVersion(ows_tokens[1]) <= StrictVersion("3.0.0")
-                and ows_tokens[2].lower() in ("getcapabilities")
-                and ows_tokens[3].upper() in ("OWS", "WCS", "WFS", "WMS", "WPS", "CSW")
+    if sec_chk_hosts:
+        if url.hostname not in proxy_urls_registry.get_proxy_allowed_hosts():
+            if not any(needle.lower() in url.query.lower() for needle in PROXY_ALLOWED_PARAMS_NEEDLES) and not any(
+                needle.lower() in url.path.lower() for needle in PROXY_ALLOWED_PATH_NEEDLES
```
ridoo: @giohappy Does it make sense to allow hosts to be proxied, just because some paths or query parameters are allowed? As an administrator, I would like to configure allowed hosts explicitly. A bad URL (maybe from a user input) which contains an arbitrary needle (path or query does not matter here) would be allowed by the proxy.

I see `PROXY_ALLOWED_PARAMS_NEEDLES` and `PROXY_ALLOWED_PATH_NEEDLES` are configurable, but these match for all hosts which are actually /not/ allowed.

IMO, servers which expose a Web API (e.g. OGC WMS) which is not CORS enabled (I guess, this is what this is all about) is not configured correctly. Not sure, if the operator wants all traffic to be routed through the GeoNode proxy. That would seem to me more as a quick and dirty workaround rather than a good productive configuration. 

What do you think? Did I miss something?

giohappy: hi @ridoo I understand your concerns, and I agree that this solution is far from ideal. 

This has been implemented to meet the logic that MS implements now. The client allows adding external layers to a map, and it implements the reverse logic compared to GeoNode: **if the URLs are not listed inside a `useCORS` array (that GeoNode sets empty) they are automatically proxied**.  Basically [it does](https://github.com/geosolutions-it/MapStore2/blob/master/java/web/src/main/resources/proxy.properties#L26-L30) what we're doing here.

If we do not support this, the catalog functionality that will be available since GeoNode 4.3.0 (thanks to a more extended integration with MS) won't work.
This is a functionality that many users wanted from MS, and this is why we are making this (temporary) move.
The final solution will be to make MS support the reverse logic, which is the one implemented by GeoNode until now: **it should go straight to the origin server, unless stated differently (`PROXY_ALLOWED_HOSTS`).**.

BTW, the two settings (which [will be set by the client](https://github.com/GeoNode/geonode-mapstore-client/commit/d6b63464b3550508459091aca94e3b2c34ec106f)) let administrators change this behavior.

giohappy: @ridoo The only alternative for now is to make the two settings an opt-in. However also the MS catalog plugin should be an opt-in in that case, otherwise, we will have a tool that will never work.

ridoo: Hey @giohappy, so this is more a (temporary) compatibility thing. Thanks for the background.

giohappy: >Hey @giohappy, so this is more a (temporary) compatibility thing. Thanks for the background.

correct @ridoo.

Issue #11511, opened 2023-09-19
GNIP 97: New metadata editor


etj:
# GNIP 97 - New metadata editor

## Overview

The metadata editor still relies on a legacy template engine which is quite hard to customize.
Also it uses unmaintained vulnerable js libraries that should be removed. 

### Proposed By

- @etj Emanuele Tajariol

### Assigned to Release

This proposal is for GeoNode 5

### State

* [x] Under Discussion
* [X] In Progress
* [x] Completed
* [ ] Rejected
* [ ] Deferred

### Motivation

- Allow easier customization
- Allow different editing frontends
- Make a clear distincton between descriptive metadata and operational metadata (owner, group)
- Remove old unmaintained vulnerable js libraries

## Proposal

The idea is to switch to standard procedures to document the metadata schema and to return the metadata info.

The metadata schema is presented using the [JSON schema](https://json-schema.org/) conventions.

The GeoNode API has been extended to provide
- `/api/v2/metadata/schema` metadata schema
- `/api/v2/metadata/instance/<PK>` resource metadata instances handling (GET/PUT)
- `/api/v2/metadata/xxx` other utility calls for the metadata handling (autocomplete endpoints referred into the schema, etc)

One of the basic driving features has been the **customization of the schema**. The implementation relies on a set of metadata handlers that can be easily extended. We embedded into GeoNode the optional `inspire` module, mostly a (working) showcase that documents how the GeoNode base model can be customized.

#### Intro

The package `geonode.metadata` contains all the metadata code.

The `manager` module contains the base engine that calls methods on the registered field handlers.
**Handlers** are the objects that manage the field definition (the json schema), and that take care of converting database objects into json schema instances and viceversa. 
A single handler may handle one or more fields.

#### `MetadataManager`

The `geonode.metadata.manager.MetadataManager` class provides these main methods:

```py
  def get_schema(self, lang=None)`
```
   Builds and returns the metadata json schema.
   The schema is created by looping on the sorted set of metadata handlers, asking them to update the schema incrementally via `def update_schema(self, jsonschema: dict, context, lang=None)`.
   Handlers may add their own field, remove fields added by previously called 
   The schema localization is only related to the `title` and `description` annotations

   This method is used by the `/api/v2/metadata/schema` endpoint.
   The default language used is the one set in the cookies, or it can be overriden by adding the query param `lang`.

```py
   def build_schema_instance(self, resource: ResourceBase, lang=None)
```
[...]

PR #12794, opened 2024-12-19
GNIP 97: New metadata editor

etj:
Complete rewriting of the metadata editor according to [GNIP 97](https://github.com/GeoNode/geonode/issues/11511)

## Checklist

For all pull requests:


gannebamm:
Can you please add some very short instructions on how to test it? Either here or (better?) at the GNIP?

giohappy:
@gannebamm today the GNIP will be updated. There's so much to write! We hope it will be ready by the end of the day, otherwise we will complete it in January, after holidays.

etj:
@gannebamm I added some info on the internals at https://github.com/GeoNode/geonode/issues/11511

Since we didn't merge the client part yet, you should install it by hand
```bash
pip install git+https://github.com/GeoNode/geonode-mapstore-client.git@metadata_manager_build#egg=django_geonode_mapstore_client
```
We didnt want to set this branch in the requirementsin the PR.

etj:
Hi @gannebamm, happy new year!  
Did you manage to test the new editor?

geonode/base/api/views.py
```diff
@@ -1517,11 +1517,25 @@ def base_linked_resources_payload(instance, user, params={}):
         linked_by_visib_ids = linked_by_visib.values_list("id", flat=True)
         linked_by = [lres for lres in linked_by_over_loopable if lres.source.id in linked_by_visib_ids]
 
-        ret["linked_by"] = LinkedResourceSerializer(
-            instance=linked_by, serialize_source=True, embed=True, many=True
-        ).data
```
mattiagiupponi, line 1522: is there any reason why the serializer is no longer used?

geonode/inspire/metadata.py
```diff
@@ -0,0 +1,306 @@
...
+THESAURUS_ACCESS_USE = "conditionsapplyingtoaccessanduse"
+
+FIELDVAL_ACCESS_USE_FREETEXT = "http://inspire.ec.europa.eu/metadata-codelist/ConditionsApplyingToAccessAndUse/freeText"
+# Should be "otherRestrictions", see https://github.com/GeoNode/geonode/issues/12745
+FIELDVAL_RESTRICTION_TYPE_DEFAULT = "limitation not listed"
+
+CONTEXT_ID = "inspire"
+
+
+class INSPIREHandler(MetadataHandler):
```
mattiagiupponi, line 31: From my undestanding, the inspire app is not totally required, is more a documentation app. If is not used, i would propose to switch it as documentation or if is needed for testing purpose only, i would document this and move the app in the tests.
Is open to discussion

geonode/metadata/api/views.py
```diff
@@ -0,0 +1,272 @@
...
+        detail=False,
+        methods=["get", "put", "patch"],
+        url_path=r"instance/(?P<pk>\d+)",
+        url_name="schema_instance",
+        permission_classes=[
+            UserHasPerms(
+                perms_dict={
+                    "default": {
+                        "GET": ["base.view_resourcebase"],
+                        "POST": ["change_resourcebase_metadata"],
```
mattiagiupponi, line 90: Should this be PUT si is the method used in the function?

mattiagiupponi (changes requested):

etj:
Hi @ridoo 

#####  `geonode:handler`

> The handler will know the schema parts it is responsible for, won't it?

The `MetadataManager` class is responsible for dispatching the field handling to the proper handler: the `geonode:handler` info tells which handler is to be called.
The handling override mechanism foresees that an handler may declare itself as the main handler for an existing field, so different handlers may work on the same field, but only the last declared one will be invoked, that is the one that lastly set himself as `geonode:handler`.

See this for example:
https://github.com/GeoNode/geonode/blob/7e81443a2a3db3c3e92bb30882c75ccbd69fc732/geonode/inspire/metadata.py#L52-L58

#####  `geonode:required`

> so for which cases geonode:required is necessary? 

JSON schema defines the required fields (using the annotation `required`) in a level above the sub schema definition.
The handlers only work on their own sub schema.
The `MetadataManager` uses the `geonode:required` custom annotations to build the upper level `required` array: https://github.com/GeoNode/geonode/blob/7e81443a2a3db3c3e92bb30882c75ccbd69fc732/geonode/metadata/manager.py#L78-L85

geonode/base/api/views.py
etj: It **is** used indeed.
Please note that the code you're pointing to has changed name from `base_linked_resources_payload` to  `base_linked_resources_instances`.  The new `base_linked_resources_payload` calls `base_linked_resources_instances` (now also used elsewere for DRY), and still uses the serializer, returning the same content as before.

geonode/inspire/metadata.py
etj: Pls note that I used the inspire app for an example in this very thread (https://github.com/GeoNode/geonode/pull/12794#issuecomment-2586817938). It is a really good point for documentation and testing handler overriding. I know it is something not used outside Europe, but I guess we need some real override implementation inside geonode.

geonode/metadata/api/views.py
```diff
@@ -0,0 +1,272 @@
...
+                return Response(response, status=400 if errors else 200)
+
+        except ResourceBase.DoesNotExist:
+            result = {"message": "The dataset was not found"}
+            return Response(result, status=404)
+
+
+def tkeywords_autocomplete(request: WSGIRequest, thesaurusid):
+
+    lang = remove_country_from_languagecode(get_language())
```
mattiagiupponi, line 133: Side note: i see that here we use `get_language()` while or the schema we use `get_language_from_request(request) ` is not better to use the same function enverywhere?

geonode/metadata/api/views.py
```diff
@@ -0,0 +1,272 @@
...
+
+    return JsonResponse({"results": ret})
+
+
+class ProfileAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request and self.request.user:
+            qs = get_available_users(self.request.user)
+        else:
+            qs = get_user_model().objects.all()
```
mattiagiupponi, line 211: I dont think makes much sense. Inside the `get_available_users` we filter the user based on what the user can see, but if we dont give a user, will return all. I guess here is better to return 0 users if the user is not provided. (even because even the Anonymous user will match the first `if` so this else will never be met)

geonode/metadata/handlers/abstract.py
```diff
@@ -0,0 +1,149 @@
...
+
+        return label
+
+    @staticmethod
+    def _localize_subschema_label(context, subschema: dict, lang: str, annotation_name: str):
+        if annotation_name in subschema:
+            subschema[annotation_name] = MetadataHandler._localize_label(context, lang, subschema[annotation_name])
+
+
+class MetadataFieldException(Exception):
```
mattiagiupponi, line 140: I guess to make it more linear, will be better to move this 3 exception is another file `exceptions.py`

geonode/upload/migrations/0050_alter_uploadsizelimit_max_size.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 4.2.16 on 2024-12-13 07:55
+
+from django.db import migrations, models
```
mattiagiupponi, line 3: I guess this migration is not needed. The maxsize is an environment variable that can change based on the user that creates the migrations. I think that it should be removed and we should keep a static value anyway is updatable via admin interface

mattiagiupponi (changes requested):

geonode/metadata/api/views.py
```diff
@@ -0,0 +1,272 @@
...
+            return Response(schema)
+
+        else:
+            response = {"Message": "Schema not found"}
+            return Response(response)
+
+    # Handle the JSON schema instance
+    @action(
+        detail=False,
+        methods=["get", "put", "patch"],
```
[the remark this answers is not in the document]
etj: it was used by the client in the first implementations, to be removed indeed

geonode/metadata/api/views.py
etj: `get_language_from_request` does not work with `WSGIRequest`, so in that case `get_language` is the only way. I would still use `get_language_from_request` whenever possibile, bc it's the right way the language should be handled

geonode/upload/migrations/0050_alter_uploadsizelimit_max_size.py
etj: @Gpetrak you added this along with the tests. Is it needed?
https://github.com/GeoNode/geonode/blob/metadata_manager/geonode/upload/migrations/0050_alter_uploadsizelimit_max_size.py

mattiagiupponi (approved):

PR #12829, opened 2025-01-16
[Fixes #12828] New remote datasets are not registered inside proxy al…

### LINKED ISSUE #12828: New remote datasets are not registered inside proxy allowed hosts when GeoNode runs asynchoronously
The link post save signal should register new remote datasets (3dtiles) inside the proxy allowed hosts registry.
This works fine when run sync but fails when run asynchronously. A restart of the GeoNode instance is required for the registry to pick up the new Link URLs.

Something is wrong [here](https://github.com/GeoNode/geonode/blob/unfold_admin/geonode/proxy/utils.py#L55-L60) probably due to the async process still running.

mattiagiupponi:
…lowed hosts when GeoNode runs asynchoronously

## Checklist

For all pull requests:


giohappy (changes requested):
We should avoid using `globals()` like this.
And I don't understand why changes do not modify the `proxy_urls_registry` content, since it's a singleton instance at the module level.
Let's investigate this better @mattiagiupponi

giohappy (changes requested):
@mattiagiupponi we cannot rely on reading from memached (or, worst, the filesystem) on each request to the proxy. In the case of 3dtiles, for example, it would be a bottleneck that worsen the proxy performances even more.

To avoid implementing a signaling system between celery and django, we can take an optimistic approach: we add the URL to the proxy **before** the celery task is created by the importer, when we're still inside the django process.
In case the remote layer creation fails, for whatever reason, we will have a registered host that doesn't map to a layer, but I don't think it's a big problem.

We can have a mechanism that reloads the registry from the DB once in a while, to realign the registry to the DB.
For example, the time of the last registry initialization can be stored in the registry. On each request we can check the time and if the registry is stale (1 day?) we reinitialize the registry.

mattiagiupponi:
backport label removed, in case is required the backport must be done manually

geonode/proxy/utils.py
```diff
@@ -45,6 +50,10 @@ def unregister_host(self, host):
         self.proxy_allowed_hosts.remove(host)
 
     def get_proxy_allowed_hosts(self):
+        if self._last_registry_load is None or (now() - self._last_registry_load).days >= getattr(
```
giohappy, line 53: @mattiagiupponi instead of doing a `getattr` to read the `PROXY_RELOAD_REGISTRY_THRESHOLD_DAYS` on each request it can be read once when the registry I instantiated.
Every small optimization helps with this proxy....

giohappy (changes requested):

giohappy (approved):

PR #12940, opened 2025-02-25
[Fixes #12939] Add rootless support for geonode, cited on GNIP 101 (#12769)

cmotadev:
# Geonode Docker, now rootless!!!

## Objectives
- Make geonode (django/celery) to run rootless
- Apply some Dockerfile best practices, to reduce image size and number of layers
- Evict to make deeper modifications, like change libraries (except the use of curl and wget - in this case we removed wget and changed to curl)

### General
- Add minimal Dockerfile labels
- Grouped apt-get commands in one RUN
- All writable files and dirctories were granted to root group (docker engine and kubernetes add container user to root group)
- No file or directories are created in runtime. if the container needs to write files, a directory are created on build time and receive chmod g=u, if the container need to create files in a system directory, the file are touched and chmoded in build time
- preserved legacy operations like sourcing .bashrc and .override_env from $HOME dir. In case of .bashrc, the entrypoint script checks if the running user has entry on /etc/passwd. if not, copy a .bashrc template from skel dir.
- created a function to parse bool entries

### Specific
- Add Environment variable to set container user ID: GEONODE_USER (default 0 - root)
- Point django, celery and invoke logs to /var/log/geonode
- Tag all images on docekr-compose-dev.yml as "dev"
- Sanitize ENV and ARGs
- Removed "service cron restart" -> We can schedule tasks on genode using external cron (calling docker or docker-machine) or use a CronJob on kubernetes based systems (rancher, openshift, minikube...)

## Next steps: 
- Make similar modifications proposed by this PR and apply to geonode-project repo
- Apply docker best practices to geonode base image (further work)

## Post Scriptum
(PS1) Geonode Volumes: If run geonode as uid 0 (root), then stop, change to unprivileged user, and start, geonode will crash due permission issues. There is two things to do: 1) Remove all volumes before change UID and start or 2) mount all volumes in a dummy container and grant write privileges to root group.

(PS2) Setting GEONODE_UID=0, geonode will run as root, like all other versions.

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

For all pull requests:

- [x] Confirm you have read the [contribution guidelines](https://github.com/GeoNode/geonode/blob/master/CONTRIBUTING.md) 
- [ ] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in the documentation)
[...]

PR #12966, opened 2025-03-12
[Fixes #12939] Add rootless support for geonode, cited on GNIP 101 (#12769)

### LINKED ISSUE #12769: DRAFT GNIP 101 : Reduce docker attack surface
# GNIP 101 - Improving Docker Security

**To PSC: There is nothing to vote upon, yet**

## Overview

According to docker security pages (https://docs.docker.com/engine/security/) using docker with default settings does provide an attack surface. The basic approach to lower the attack surface is to enable [rootless](https://docs.docker.com/engine/security/rootless/) processes inside the containers. The GeoNode docker composition currently uses the root user for starting the container processes. Disabling root and adding other security related functionality could significantly reduce the attack surface.

### Proposed By

Florian Hoedt, Thünen-Institute

### Assigned to Release

This proposal is not yet scheduled for a release.

### State

* [x] Under Discussion
* [ ] In Progress
* [ ] Completed
* [ ] Rejected
* [ ] Deferred

### Motivation

An upcoming talk about [GeoNode in production](https://talks.osgeo.org/foss4g-2024/talk/BCYFPL/ ) at FOSS4G 2024 which will cover topics like:

- Addressing security requirements like rootless execution and random UID support
- Improving security by separating credentials and using Podman instead of Docker

sparked my interesst. I see that improving the current security model is an important task but also one which requires an substential amount of ressources. I created this GNIP as an invitation to the downstream project to contribute to the main project. I see people like @cmotadev working on that. Maybe he can post some links to the downstream project´s github as a reference how to achieve the above mentioned goals.

## Proposal

The options to achieve the mentioned goal are manifold. The fist level of reducing the attack surface is to enable rootless docker for the containers as done in the geoSGB use-case described above.

### Backwards Compatibility

Currently, the amount of changes is not yet known. If those changes do not effect the django code but solely how docker containers are started they could be backwards compatible. If django code needs changes the work of backporting those is likely to big.

## Future evolution

Apart from using podman and a rootless docker we could implement [AppArmor](https://docs.docker.com/engine/security/apparmor/) profiles per container to further reduce the attack surface. An example of this for an nginx container is listed [here](https://docs.docker.com/engine/security/apparmor/#nginx-example-profile). 

## Feedback

## Voting

Project Steering Committee:

* Alessio Fabiani:
* Francesco Bartoli:
* Giovanni Allegri:
* Toni Schoenbuchner:
* Florian Hoedt:

### LINKED ISSUE #12939: Add rootless support for geonode
Add rootless support for geonode django/celery, in addition to this [pull request: rootless geonode docker](https://github.com/GeoNode/geonode-docker/pull/56).

## Objectives
- Make geonode (django/celery) to run rootless
- Apply some Dockerfile best practices, to reduce image size and number of layers
- Evict to make deeper modifications, like change libraries (except the use of curl and wget - in this case _we removed wget and changed to curl_)

giohappy:
Original contribution by @cmotadev in https://github.com/GeoNode/geonode/pull/12940

# Geonode Docker, now rootless!!!

## Objectives
- Make geonode (django/celery) to run rootless
- Apply some Dockerfile best practices, to reduce image size and number of layers
- Evict to make deeper modifications, like change libraries (except the use of curl and wget - in this case we removed wget and changed to curl)

### General
- Add minimal Dockerfile labels
- Grouped apt-get commands in one RUN
- All writable files and dirctories were granted to root group (docker engine and kubernetes add container user to root group)
- No file or directories are created in runtime. if the container needs to write files, a directory are created on build time and receive chmod g=u, if the container need to create files in a system directory, the file are touched and chmoded in build time
- preserved legacy operations like sourcing .bashrc and .override_env from $HOME dir. In case of .bashrc, the entrypoint script checks if the running user has entry on /etc/passwd. if not, copy a .bashrc template from skel dir.
- created a function to parse bool entries

### Specific
- Add Environment variable to set container user ID: GEONODE_USER (default 0 - root)
- Point django, celery and invoke logs to /var/log/geonode
- Tag all images on docekr-compose-dev.yml as "dev"
- Sanitize ENV and ARGs
- Removed "service cron restart" -> We can schedule tasks on genode using external cron (calling docker or docker-machine) or use a CronJob on kubernetes based systems (rancher, openshift, minikube...)

## Next steps: 
- Make similar modifications proposed by this PR and apply to geonode-project repo
- Apply docker best practices to geonode base image (further work)

## Post Scriptum
(PS1) Geonode Volumes: If run geonode as uid 0 (root), then stop, change to unprivileged user, and start, geonode will crash due permission issues. There is two things to do: 1) Remove all volumes before change UID and start or 2) mount all volumes in a dummy container and grant write privileges to root group.

(PS2) Setting GEONODE_UID=0, geonode will run as root, like all other versions.

## Documentation on rootless best practices
 - https://developers.redhat.com/articles/2021/11/11/best-practices-building-images-pass-red-hat-container-certification#best_practice__3__set_group_ownership_and_file_permissions
 - https://sysdig.com/learn-cloud-native/dockerfile-best-practices/

## Checklist

For all pull requests:

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -3,11 +3,14 @@ version: '3.9'
 # Common Django template for GeoNode and Celery services below
 x-common-django:
   &default-common-django
-  image: geonode/geonode:local
+  image: geonode/geonode:dev
   build:
       context: ./
-      dockerfile: Dockerfile
+      args:
+        GEONODE_BASE_IMAGE: geonode/geonode-base:latest-ubuntu-22.04
+        GEONODE_VERSION: master
   restart: unless-stopped
+  user: '${GEONODE_UID}'
   env_file:
     - .env
   volumes:
@@ -52,15 +55,17 @@ services:
 
   # Nginx is serving django static and media files and proxies to django and geonode
   geonode:
-    image: geonode/nginx:1.25.3-latest
+    image: geonode/nginx:dev
     container_name: nginx4${COMPOSE_PROJECT_NAME}
     env_file:
       - .env
     environment:
       - RESOLVER=127.0.0.11
+    user: '${GEONODE_UID}'
     ports:
-      - "${HTTP_PORT}:80"
-      - "${HTTPS_PORT}:443"
+      # change container ports to run rootless nginx (80->8080 and 443->8443)
+      - "${HTTP_PORT}:8080"
+      - "${HTTPS_PORT}:8443"
     volumes:
       - nginx-confd:/etc/nginx
       - nginx-certificates:/geonode-certificates
@@ -69,17 +74,20 @@ services:
 
   # Gets and installs letsencrypt certificates
   letsencrypt:
-    image: geonode/letsencrypt:2.6.0-latest
+    image: geonode/letsencrypt:dev
     container_name: letsencrypt4${COMPOSE_PROJECT_NAME}
     env_file:
       - .env
+    user: '${GEONODE_UID}'
     volumes:
       - nginx-certificates:/geonode-certificates
     restart: unless-stopped
+    # deploy:
+    #   replicas: 0
 
   # Geoserver backend
   geoserver:
-    image: geonode/geoserver:2.24.3-latest
+    image: geonode/geoserver:dev
     container_name: geoserver4${COMPOSE_PROJECT_NAME}
     healthcheck:
       test: "curl -m 10 --fail --silent --write-out 'HTTP CODE : %{http_code}\n' --output /dev/null http://geoserver:8080/geoserver/ows"
@@ -87,6 +95,7 @@ services:
       interval: 60s
       timeout: 10s
       retries: 2
+    user: '${GEONODE_UID}'
     env_file:
       - .env
     ports:
@@ -105,9 +114,10 @@ services:
         condition: service_healthy
 
   data-dir-conf:
-    image: geonode/geoserver_data:2.24.3-latest
+    image: geonode/geoserver_data:dev
     container_name: gsconf4${COMPOSE_PROJECT_NAME}
     entrypoint: sleep infinity
+    user: '${GEONODE_UID}'
     volumes:
       - geoserver-data-dir:/geoserver_data/data
     restart: unless-stopped
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,8 +6,11 @@ x-common-django:
   image: geonode/geonode:latest-ubuntu-22.04
   build:
       context: ./
-      dockerfile: Dockerfile
+      args:
+        GEONODE_BASE_IMAGE: geonode/geonode-base:latest-ubuntu-22.04
+        GEONODE_VERSION: master
   restart: unless-stopped
+  user: '${GEONODE_UID}'
   env_file:
     - .env
   volumes:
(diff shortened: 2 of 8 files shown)


docker-compose.yml
```diff
@@ -34,7 +37,7 @@ services:
       retries: 2
     environment:
       - IS_CELERY=False
-    entrypoint: ["/usr/src/geonode/entrypoint.sh"]
+    # entrypoint: ["/usr/src/geonode/entrypoint.sh"]
```
mattiagiupponi, line 40: Is there any techincal reason why this is commented?

docker-compose-dev.yml
```diff
@@ -105,9 +114,10 @@ services:
         condition: service_healthy
 
   data-dir-conf:
-    image: geonode/geoserver_data:2.24.3-latest
+    image: geonode/geoserver_data:dev
```
mattiagiupponi, line 117: this will not be removed by the other PR for getting rid of the geoserver_data?

mattiagiupponi:
Made the following test:

1) build the images for `nginx`, `geoserver` and `geoserver_data` from [geonode-docker](https://github.com/GeoNode/geonode-docker/pull/59) repo
2) set in the `.env` file the `GEONODE_UID` as 1000 since is my local user ID
image
3) update the `docker-compose` file to re-enable the `entrypoint.sh` file, otherwise Django and celery were not running
4) update the `docker-compose` file to use the root_less images generated at point 1
image
5) run docker-compose file as my user (not root)
image
6) Geonode is still running as root

image

Did i miss some step? @cmotadev @giohappy ?

mattiagiupponi:
ok, with @giohappy we made the following two changes:
1) re-comment the `entrypoint.sh` in the docker-compose file
2) the entrypoint in the Dockerfile should be updated as the following:` ENTRYPOINT ["/usr/src/geonode/entrypoint.sh"]` by using the absolute path of the entrypoint. In this way the user is used instead of root:
image

docker-compose.yml
cmotadev: This was commented because of the entrypoint, even in django or celery is the same. I think there is no reason to maintain this on docker compose. It was transported to dockerfile.

https://github.com/GeoNode/geonode/blob/33f56a33b6ed8782949c6fa2acfd7a7825e14e4e/Dockerfile#L52-L56

If you need to run the container with no entrypoint, you have to set `--entrypoint=""`on `docker run`command. In `docker compose`, itś only `entrypoint: ""`

I suggested this because all run commands need this entrypoint. Switching back does the same effect, but it has to declare the entrypoint script

docker-compose.yml
```diff
@@ -46,7 +49,7 @@ services:
         condition: service_healthy
     environment:
       - IS_CELERY=True
-    entrypoint: ["/usr/src/geonode/entrypoint.sh"]
+    # entrypoint: ["/usr/src/geonode/entrypoint.sh"]
```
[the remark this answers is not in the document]
cmotadev: Same: https://github.com/GeoNode/geonode/pull/12966#discussion_r1996337807

docker-compose-dev.yml
cmotadev: yes, when no longer needed

PR #13140, opened 2025-04-29
[Fixes #12901] Metadata: Enforcing group permission on save

### LINKED ISSUE #12901: Metadata: enforce group permission on save
When a user edits metadata, the autocomplete request correctly returns only the groups visible to the user.

Make sure that if a user forges a PUT request containing a group that should not be visibile, the operation fails.

Gpetrak:
This PR was created according to this issue #12901 
The tests were fixed while new tests were added.

## Checklist

For all pull requests:


geonode/base/api/serializers.py
```diff
@@ -131,6 +131,29 @@ class Meta:
...
 
+    def update(self, instance, validated_data):
+        user = self.context["request"].user
+
+        # Check if 'group' is being updated
+        new_group = validated_data["group"]
+
+        if not (user.is_superuser or user.is_staff):
+            # Check if the user is a member of the defined group
+            allowed_groups = GroupProfile.objects.filter(groupmember__user=user)
```
mattiagiupponi, line 142: i suggest in this case to use the util [get_user_groups](https://github.com/GeoNode/geonode/blob/90a90cb7597a7031b16949f8335e1d304ea13ba1/geonode/security/utils.py#L232) or [get_user_visible_groups](https://github.com/GeoNode/geonode/blob/90a90cb7597a7031b16949f8335e1d304ea13ba1/geonode/security/utils.py#L242) based on the issue requirement

geonode/metadata/handlers/group.py
```diff
@@ -68,8 +68,40 @@ def get_jsonschema_instance(self, resource, field_name, context, errors, lang=No
...
+            return
+
         if id is not None:
-            gp = GroupProfile.objects.get(pk=id)
-            resource.group = gp.group
+
+            if user.is_superuser or user.is_staff:
+                allowed_groups = GroupProfile.objects.all()
+            else:
+                allowed_groups = GroupProfile.objects.filter(groupmember__user=user)
```
mattiagiupponi, line 91: if you want it more compressed, you can also write it as follow:

```python
allowed_groups = GroupProfile.objects.all()
if not user.is_superuser:
    allowed_groups = allowed_groups.filter(groupmember__user=user)
```

geonode/metadata/handlers/group.py
```diff
@@ -68,8 +68,40 @@ def get_jsonschema_instance(self, resource, field_name, context, errors, lang=No
...
+
+            try:
+                gp = allowed_groups.get(pk=id)
+                resource.group = gp.group
+            except GroupProfile.DoesNotExist as e:
```
mattiagiupponi, line 96: In general is more efficient to use the [exists() rather then using the try/except block](https://stackoverflow.com/a/23567560/7597536). This can be converted in something like:

```python
gp = allowed_groups.filter(pk=id)
if gp.exists():
    resource.group = gp.group.first()
else:
    self._set_error(
        errors,
        [field_name],
        self.localize_message(
            context,
            "metadata_group_error_missing_group",
            {"fieldname": field_name, "exc": e},
        ),
    )
```

mattiagiupponi (changes requested):

geonode/metadata/handlers/group.py
Gpetrak: @mattiagiupponi thank you for your comment, it’s definitely more concise. That said, I’m a bit concerned that this approach might result in two database queries when the user is not a superuser. What do you think?

mattiagiupponi: Good point, we can do something like this:

```python
group_filter = {}
if not user.is_superuser:
    group_filter['groupmember__user'] = user
allowed_groups = GroupProfile.objects.filter(**group_filter)
```

geonode/base/api/serializers.py
mattiagiupponi:
```
gp = user.groups.get(pk=new_group.pk)
```

Gpetrak: Done

mattiagiupponi (approved):

mattiagiupponi (approved):

PR #13315, opened 2025-06-30
[Fixes #13311]: Implement caching for User Permission retrieval for Resource Base Endpoints Reducing number of queries in Database.

### LINKED ISSUE #13311: Implement caching for User Permission retrieval for Resource Base Endpoints Reducing number of queries in Database.
## Overview

The proposal below implements caching for user permissions retrieval, with the following key considerations:

- Permissions are requested in many endpoints (and internal methods). We want to abstract the caching for permissions so that other methods and APIs can leverage it
- Permissions are critical and sensitive information. The implementation must take into account the security implications of information leaks, such as cross-session interference (mixing users' related information between user sessions) or violations of access constraints to the permissions API endpoint
- Execute performance tests to verify that the cache improves response times and DB I/O. Verify any detrimental effect on Memcached access when cache requests increase
- Sufficient test coverage is required

### Implementation Plan

1. Make cache set based on user and resource in the permission get permission class, making default to False.
2. Call the class from resource serializer passing cache=True.
3. Delete the cache in ResourceManager if  something changes in permission for the particular resource.
4. Execute Performance Test.
5. Write test code for the Resource Endpoint if cache works correctly and does not affect the response, handling different users.
### State

* [ ] Under Discussion
* [x] In Progress
* [ ] Completed
* [ ] Rejected
* [ ] Deferred

### Backwards Compatibility

This is Backward compatible.

## Future evolution

This implemented cache can be used in other part of the system endpoints.

sijandh35:
Fixes #13311
This PR optimizes the performance of the Resource API as described in the issue [GeoNode/geonode#13311](https://github.com/GeoNode/geonode/issues/13311). As mentioned in the issue, a test case has been written to validate the improvements. 

Screenshot from 2025-06-30 11-21-03

Fig: This figures shows initial curl request results without implementing cache on resource endpoint.

Screenshot from 2025-06-30 11-23-31

Fig: This figures shows curl request results  implementing cache on resource endpoint.

A local Python test script was also executed to evaluate cache performance and scalability. The results are shown below:

Screenshot from 2025-06-30 12-51-54

Fig: Performance test results demonstrating cache scalability.

## Checklist

For all pull requests:

### DIFF
--- a/geonode/base/api/serializers.py
+++ b/geonode/base/api/serializers.py
@@ -780,13 +780,17 @@ class ResourceBaseSerializer(DynamicModelSerializer):
             if not user.can_change_resource_field(instance, field) and field in validated_data:
                 validated_data.pop(field)
         return super().update(instance, validated_data)
-    
+
     def get_perms(self, instance):
         """
         Returns the permissions for the resource instance using Django cache.
         """
         request = self.context.get("request")
-        permissions = permissions_registry.get_perms(instance=instance, user=request.user,is_cache=True) if request and request.user and instance else []
+        permissions = (
+            permissions_registry.get_perms(instance=instance, user=request.user, is_cache=True)
+            if request and request.user and instance
+            else []
+        )
         return permissions
 
     def save(self, **kwargs):
--- a/geonode/base/api/serializers.py
+++ b/geonode/base/api/serializers.py
@@ -787,7 +787,7 @@ class ResourceBaseSerializer(DynamicModelSerializer):
         """
         request = self.context.get("request")
         permissions = (
-            permissions_registry.get_perms(instance=instance, user=request.user, is_cache=True)
+            permissions_registry.get_perms(instance=instance, user=request.user, use_cache=True)
             if request and request.user and instance
             else []
         )
--- a/geonode/base/api/serializers.py
+++ b/geonode/base/api/serializers.py
@@ -654,7 +654,7 @@ class ResourceBaseSerializer(DynamicModelSerializer):
     download_url = DownloadLinkField(read_only=True)
     favorite = FavoriteField(read_only=True)
     download_urls = DownloadArrayLinkField(read_only=True)
-    perms = DynamicRelationField(PermsSerializer, source="id", read_only=True)
+    perms = serializers.SerializerMethodField(read_only=True)
     links = DynamicRelationField(LinksSerializer, source="id", read_only=True)
 
     # Deferred fields
@@ -780,6 +780,14 @@ class ResourceBaseSerializer(DynamicModelSerializer):
             if not user.can_change_resource_field(instance, field) and field in validated_data:
                 validated_data.pop(field)
         return super().update(instance, validated_data)
+    
+    def get_perms(self, instance):
+        """
+        Returns the permissions for the resource instance using Django cache.
+        """
+        request = self.context.get("request")
+        permissions = permissions_registry.get_perms(instance=instance, user=request.user,is_cache=True) if request and request.user and instance else []
+        return permissions
 
     def save(self, **kwargs):
         extent = self.validated_data.pop("extent", None)
--- a/geonode/base/api/tests.py
+++ b/geonode/base/api/tests.py
@@ -3662,7 +3662,6 @@ class TestPermissionsCaching(GeoNodeBaseTestSupport):
         test_resource = self.resources[0]
 
         admin_perms_1 = permissions_registry.get_perms(instance=test_resource, user=self.admin_user, use_cache=True)
-
         admin_perms_2 = permissions_registry.get_perms(instance=test_resource, user=self.admin_user, use_cache=True)
 
         self.assertEqual(admin_perms_1, admin_perms_2)
@@ -3674,8 +3673,7 @@ class TestPermissionsCaching(GeoNodeBaseTestSupport):
         self.assertIn("change_resourcebase_permissions", admin_perms_1)
         self.assertIn("delete_resourcebase", admin_perms_1)
 
-    def test_permission_annonomous_users(self):
-
+    def test_permission_anonymous_users(self):
         permissions_registry = PermissionsHandlerRegistry()
         test_resource = self.resources[0]
         anonymous_user = AnonymousUser()
@@ -3688,12 +3686,10 @@ class TestPermissionsCaching(GeoNodeBaseTestSupport):
         self.assertEqual(anon_perms_1, anon_perms_2)
 
     def test_owner_user_permissions_caching(self):
-
         permissions_registry = PermissionsHandlerRegistry()
         test_resource = self.resources[0]
 
         user_perms_1 = permissions_registry.get_perms(instance=test_resource, user=self.test_user, use_cache=True)
-
         user_perms_2 = permissions_registry.get_perms(instance=test_resource, user=self.test_user, use_cache=True)
 
         self.assertEqual(user_perms_1, user_perms_2)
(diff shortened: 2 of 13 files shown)


sijandh35:
@giohappy Sure,Updated.

geonode/base/api/serializers.py
```diff
@@ -781,6 +781,18 @@ def update(self, instance, validated_data):
                 validated_data.pop(field)
         return super().update(instance, validated_data)
 
+    def get_perms(self, instance):
+        """
+        Returns the permissions for the resource instance using Django cache.
+        """
+        request = self.context.get("request")
+        permissions = (
+            permissions_registry.get_perms(instance=instance, user=request.user, is_cache=True)
```
mattiagiupponi: i would use something like `use_cache` instead of `is_cache`

geonode/security/registry.py
```diff
@@ -62,14 +63,32 @@ def fixup_perms(self, instance, payload, include_virtual=True, *args, **kwargs):
...
         For example before return the payload, we can virtually remove perms
         to the resource
         include_user_add_resource -> If true add the add_resourcebase to the user perms if the user have it
         """
+        cache_key = None
+        if is_cache:
+            if user:
+                cache_key = f"resource_perms:{instance.pk}:{user.pk if not user.is_anonymous else 'anonymous'}"
+            else:
+                cache_key = f"resource_perms:{instance.pk}:__ALL__"
```
mattiagiupponi: better to have a small comment in the code or at function level describing how the cache works and what it `__ALL__` in the cache, also it would be nice to have this as a separated method and have a small test to be sure that the keys are consistent

geonode/security/registry.py
```diff
@@ -81,7 +100,12 @@ def get_perms(self, instance, user=None, include_virtual=True, include_user_add_
         if user:
             if include_user_add_resource and user.has_perm("base.add_resourcebase"):
                 payload["users"][user].extend(["add_resourcebase"])
+            if is_cache and cache_key:
+                # set cache for 7 days
+                cache.set(cache_key, payload["users"][user], 604800)
```
mattiagiupponi: i would set the expiration time of the cache as settings, in that way at least is easily customizable

geonode/resource/manager.py
```diff
@@ -489,11 +494,27 @@ def exec(self, method: str, uuid: str, /, instance: ResourceBase = None, **kwarg
...
+    def invalidate_permissions_cache(self, uuid: str, /, instance: ResourceBase = None) -> bool:
+        """Invalidate the cache for the given resource."""
+        _resource = instance or ResourceManager._get_instance(uuid)
+        if _resource:
+            permissions = permissions_registry.get_perms(instance=_resource.get_real_instance())
+            users = Profile.objects.filter(
+                Q(groups__in=permissions["groups"].keys()) | Q(id__in=[x.id for x in permissions["users"].keys()])
+            )
+            for user in users:
+                cache.delete(f"resource_perms:{_resource.pk}:{user.pk}")
```
mattiagiupponi: if you want is possible to use also the `cache.delete_many` which will batch truncate the cache.

It takes a list of cache keys, so something like this should work:

```python
cache.delete_many([f"resource_perms:{_resource.pk}:{u.pk}" for u in users])
```

https://docs.djangoproject.com/en/4.2/topics/cache/#django.core.cache.cache.delete_many (to be tested with memcached and localcache)

mattiagiupponi (changes requested):
Hi @sijandh35 
There are a couple of scenario that, in my opinion, are not covered:

1) Why dont we  set the cache in the `set_permissions` method of the resource manager? The system is already working on the perms, so it might make sense to save the cached perms there, so the permissions registry can just read it from the cache without setting them. Plus, in this way, the api will be already fast in the first call. 
2) I cannot see any system where the cache is deleted when a user / group  is deleted. If we dont clean the cache when one of them is removed from GeoNode, we risk to have cached entries for 1 week that does not exists in GeoNode anymore. This will be even more dangerous with an higher cache value (for the resource itself, this behaviour is covered by the `remove_permissions` method of the resource manager)

What you think about those points?

giohappy:
>Why dont we set the cache in the set_permissions method of the resource manager?

@mattiagiupponi IMHO the management of the permissions cache is not in the roles of the resource manager. It should rather be something. It's an internal thing of the permission registry, that should be transparent to external users of the service (resource mana, api, etc.)

sijandh35:
> Hi @sijandh35 There are a couple of scenario that, in my opinion, are not covered:
> 
> 1. Why dont we  set the cache in the `set_permissions` method of the resource manager? The system is already working on the perms, so it might make sense to save the cached perms there, so the permissions registry can just read it from the cache without setting them. Plus, in this way, the api will be already fast in the first call.

@mattiagiupponi I have set the permission in the permission_registry because I thought it would be safe, as it is responsible for calculating the final permission level even if something added explicitly like user related or group related from other part of system. Another reason I considered adding it there is that there will already be resources in the system, so we would either have to run a management command for earlier resources to set it otherwise it wont be set for older one. I agree with your point that it will be faster for newer resources or resources that have changed after implementing the cache, but I thought it would be safe if we set it in the final calculated permission. What do you think about this?

> 2. I cannot see any system where the cache is deleted when a user / group  is deleted. If we dont clean the cache when one of them is removed from GeoNode, we risk to have cached entries for 1 week that does not exists in GeoNode anymore. This will be even more dangerous with an higher cache value (for the resource itself, this behaviour is covered by the `remove_permissions` method of the resource manager)

Yes I totally agree and thank you for pointing it. I will update the PR considering this scenario.

giohappy:
For the first point, I agree with @sijandh35 (as I said in my previous comment).

mattiagiupponi:
ok @giohappy @sijandh35 for the point 1, please avaluate the point 2 so we can have a clean cache in case of deletion

geonode/base/api/tests.py
```diff
@@ -3588,3 +3592,482 @@ def __csw_request_factory():
...
+        permissions_registry.get_perms(instance=test_resource, user=self.admin_user, use_cache=True)
+        permissions_registry.get_perms(instance=test_resource, user=anonymous_user, use_cache=True)
+        permissions_registry.get_perms(instance=test_resource, group=self.test_group, use_cache=True)
+
+        # Verify the generated keys match what's in cache
+        self.assertIsNotNone(cache.get(actual_admin_key), "Admin cache key should exist in cache")
+        self.assertIsNotNone(cache.get(actual_anon_key), "Anonymous cache key should exist in cache")
+        self.assertIsNotNone(cache.get(actual_group_key), "Group cache key should exist in cache")
+
+    def test_cache_invalidation_on_permission_change(self):
```
giohappy: This tests that permissions are invalidated if the cache is removed, but a test for real permission changes is required.

geonode/base/api/tests.py
```diff
@@ -3588,3 +3592,482 @@ def __csw_request_factory():
...
+        all_perms_1 = permissions_registry.get_perms(instance=test_resource, use_cache=True)
+        all_perms_2 = permissions_registry.get_perms(instance=test_resource, use_cache=True)
+
+        self.assertEqual(all_perms_1, all_perms_2)
+
+        # Verify structure contains both users and groups
+        self.assertIn("users", all_perms_1)
+        self.assertIn("groups", all_perms_1)
+
+    def test_user_deletion_cache_invalidation(self):
```
giohappy: The user is deleted at the end of the test. I don't understand what's the purpose of this test.

geonode/base/api/tests.py
```diff
@@ -3588,3 +3592,482 @@ def __csw_request_factory():
...
+            self.assertIsNotNone(cache.get(cache_key))
+
+        permissions_registry.delete_resource_permissions_cache(temp_group, user_clear_cache=False)
+
+        for cache_key in group_cache_keys:
+            self.assertIsNone(
+                cache.get(cache_key), f"Group cache key {cache_key} should be cleared after group deletion"
+            )
+
+    def test_resource_deletion_cache_invalidation(self):
```
giohappy: The same here. The resource is deleted at the end, so I don't understand if this test should verify permissions invalidation after the resource is deleted or something else.

geonode/base/api/tests.py
mattiagiupponi: +1

geonode/base/api/tests.py
mattiagiupponi: is a test to validate that the delete cache method works correctly, I would suggest to delete the user and evaluate that the cache is automatically deleted as expected

geonode/base/api/tests.py
mattiagiupponi: its evaluate that by calling the `delete_resource_permissions_cache` the cache is correctly deleted. 
I suggest to evaluate that by deleting the resource, the cache is correctly deleted

geonode/security/registry.py
```diff
@@ -62,27 +68,173 @@ def fixup_perms(self, instance, payload, include_virtual=True, *args, **kwargs):
...
-        The permissions payload can be edited by each permissions handler.
-        For example before return the payload, we can virtually remove perms
-        to the resource
-        include_user_add_resource -> If true add the add_resourcebase to the user perms if the user have it
+        Clear the cache for resource permissions when activity related to resource is performed.
+        This ensures that permissions are recalculated on the next request.
+        """
+
+        from geonode.base.models import ResourceBase
+        from geonode.people.models import Profile
```
mattiagiupponi, line 118: can we move those on top to avoit the import everytime the function is called? or is raising some recursing error?

mattiagiupponi (changes requested):

geonode/security/registry.py
sijandh35: @mattiagiupponi  yes it is added there due to circular import .

mattiagiupponi:
@giohappy I made some test (stress test too) and everything looks fine. I think we can merge it

mattiagiupponi (approved):

PR #13396, opened 2025-07-30
[Fixes #13378] Implement POST API for asset creation.

### LINKED ISSUE #13378: Implement POST API for assets
### Description

`Assets` represent uploaded data. They're currently used to represent file uploads connected to resources. The connection is indirect, through `Link` model. A `Link` represents a connection between a `Resource` and an `Asset` when the `Link.asset` FK is populated.

The `assets` module provides the model, a few read-only API endpoints, and some utilities methods to simplify the creation and management 
assets.

We want to extend the assets API with a create method to support the creation of a new asset and, optionally, its connection to an existing resource.
The API will implement the authentication to allow the POST method only for authenticated users and having the `add_resourcebase` permission (the `UserHasPerms(perms_dict={"default": {"POST": ["base.add_resourcebase"]}})` permission class could be used).

The AP will accept the usual `FILE` object (the actual uploaded file) and an optional resource id parameter for the resource to be connected with the asset.

The `assets.utils` methods will be used to create the asset and, if a resource id is provided, link it.  In this moment, we have the `assets.utils.create_asset_and_link()`, which accepts a resource, creates the asset and creates the link with `assets.utils.create_link()`- 
We could create `assets.utils.create_asset()`, which will be called in case we only need to create the asset without linking. `assets.utils.create_asset_and_link()` will call `create_asset` and `create_link` instead.

sijandh35:

## Checklist

For all pull requests:

### DIFF
--- a/geonode/assets/serializers.py
+++ b/geonode/assets/serializers.py
@@ -119,6 +119,7 @@ class LocalAssetSerializer(AssetSerializer):
             if not permissions_registry.user_has_perm(
                 user, resource.get_self_resource(), "change_resourcebase", include_virtual=True
             ):
+                logger.debug("The user does not have permissions to change the resource selected")
                 raise PermissionDenied("The user does not have permissions to change the resource selected")
 
         if not title:
@@ -134,6 +135,7 @@ class LocalAssetSerializer(AssetSerializer):
         file_path = retrieved_paths.get("file")
 
         if not file_path:
+            logger.debug("Could not save the file.")
             raise serializers.ValidationError("Could not save the file.")
 
         if resource_id:
--- a/geonode/assets/serializers.py
+++ b/geonode/assets/serializers.py
@@ -22,6 +22,8 @@ from django.contrib.auth import get_user_model
 
 from dynamic_rest.serializers import DynamicModelSerializer
 from dynamic_rest.fields.fields import DynamicComputedField
+from rest_framework import serializers
+
 
 from geonode.assets.models import (
     Asset,
@@ -60,15 +62,35 @@ class AssetSubclassField(DynamicComputedField):
 
 class AssetSerializer(DynamicModelSerializer):
 
-    owner = SimpleUserSerializer(embed=False)
-    asset_type = ClassTypeField()
-    subinfo = AssetSubclassField()
+    owner = SimpleUserSerializer(embed=False, read_only=True)
+    asset_type = ClassTypeField(read_only=True)
+    subinfo = AssetSubclassField(read_only=True)
+
+    file = serializers.FileField(allow_empty_file=False, use_url=False, write_only=True, required=True)
+    resource_id = serializers.IntegerField(write_only=True, required=False)
 
     class Meta:
         model = Asset
         name = "asset"
         # fields = ("pk", "title", "description", "type", "owner", "created")
-        fields = ("pk", "title", "description", "type", "owner", "created", "asset_type", "subinfo")
+        fields = (
+            "pk",
+            "title",
+            "description",
+            "type",
+            "owner",
+            "created",
+            "asset_type",
+            "subinfo",
+            "file",
+            "resource_id",
+        )
+        extra_kwargs = {
+            "title": {"read_only": True},
+            "description": {"read_only": True},
+            "type": {"read_only": True},
+            "created": {"read_only": True},
+        }
 
 
 class LocalAssetSerializer(AssetSerializer):
--- a/geonode/assets/serializers.py
+++ b/geonode/assets/serializers.py
@@ -17,7 +17,7 @@
 #
 #########################################################################
 import logging
-
+import os
 from django.contrib.auth import get_user_model
 
 from dynamic_rest.serializers import DynamicModelSerializer
@@ -35,6 +35,8 @@ from geonode.assets.models import (
     LocalAsset,
 )
 from rest_framework.exceptions import PermissionDenied
+from django.conf import settings
+
 
 logger = logging.getLogger(__name__)
 
@@ -105,8 +107,12 @@ class LocalAssetSerializer(AssetSerializer):
         resource_id = validated_data.pop("resource_id", None)
         title = validated_data.pop("title", None)
         description = validated_data.pop("description", None)
-        type = validated_data.pop("type", None)
+        asset_type = validated_data.pop("type", None)
         user = self.context["request"].user
+
+        if file and not os.path.splitext(file.name)[1].lower()[1:] in settings.ALLOWED_DOCUMENT_TYPES:
+            logger.debug("This file type is not allowed")
+            raise serializers.ValidationError("This file type is not allowed")
 
         if resource_id:
             resource = get_object_or_404(ResourceBase, pk=resource_id)
@@ -132,9 +138,17 @@ class LocalAssetSerializer(AssetSerializer):
 
         if resource_id:
             localasset, _ = create_asset_and_link(
-                resource, user, [file_path], title=title, description=description, asset_type=type
+                resource,
+                user,
+                [file_path],
+                title=title,
+                description=description,
+                asset_type=asset_type,
+                clone_files=False,
             )
         else:
-            localasset = create_asset(user, [file_path], title=title, description=description, asset_type=type)
+            localasset = create_asset(
+                user, [file_path], title=title, description=description, asset_type=asset_type, clone_files=False
+            )
 
         return localasset
--- a/geonode/assets/serializers.py
+++ b/geonode/assets/serializers.py
@@ -110,6 +110,8 @@ class LocalAssetSerializer(AssetSerializer):
         asset_type = validated_data.pop("type", None)
         user = self.context["request"].user
 
+        resource = None
+
         if file and not os.path.splitext(file.name)[1].lower()[1:] in settings.ALLOWED_DOCUMENT_TYPES:
             logger.debug("This file type is not allowed")
             raise serializers.ValidationError("This file type is not allowed")
@@ -138,7 +140,7 @@ class LocalAssetSerializer(AssetSerializer):
             logger.debug("Could not save the file.")
             raise serializers.ValidationError("Could not save the file.")
 
-        if resource_id:
+        if resource:
             localasset, _ = create_asset_and_link(
                 resource,
                 user,
--- a/geonode/assets/serializers.py
+++ b/geonode/assets/serializers.py
@@ -66,7 +66,7 @@ class AssetSubclassField(DynamicComputedField):
 
 class AssetSerializer(DynamicModelSerializer):
 
-    owner = SimpleUserSerializer(embed=False, required=False)
+    owner = SimpleUserSerializer(embed=False, required=False, read_only=True)
     asset_type = ClassTypeField(required=False)
     subinfo = AssetSubclassField(required=False)
 
@@ -75,6 +75,12 @@ class AssetSerializer(DynamicModelSerializer):
         name = "asset"
         # fields = ("pk", "title", "description", "type", "owner", "created")
         fields = ("pk", "title", "description", "type", "owner", "created", "asset_type", "subinfo")
+        extra_kwargs = {
+            "title": {"required": False},
+            "description": {"required": False},
+            "type": {"required": False},
+            "created": {"required": False, "read_only": True},
+        }
 
 
 class LocalAssetSerializer(AssetSerializer):
@@ -89,8 +95,7 @@ class LocalAssetSerializer(AssetSerializer):
             "title": {"required": False},
             "description": {"required": False},
             "type": {"required": False},
-            "location": {"required": False},
-            "created": {"required": False},
+            "location": {"required": False, "read_only": True},
         }
 
     def create(self, validated_data):
@@ -100,6 +105,9 @@ class LocalAssetSerializer(AssetSerializer):
         description = validated_data.pop("description", None)
         type = validated_data.pop("type", None)
         user = self.context["request"].user
+
+        if not title:
+            title = file.name
 
         handler = LocalAssetHandler()
         asset_dir = handler._create_asset_dir()
--- a/geonode/assets/serializers.py
+++ b/geonode/assets/serializers.py
@@ -17,23 +17,16 @@
 #
 #########################################################################
 import logging
-import os
+
 from django.contrib.auth import get_user_model
 
 from dynamic_rest.serializers import DynamicModelSerializer
 from dynamic_rest.fields.fields import DynamicComputedField
-from geonode.base.models import ResourceBase
-from geonode.assets.utils import create_asset, create_asset_and_link
-from django.shortcuts import get_object_or_404
-from rest_framework import serializers
-from geonode.security.registry import permissions_registry
 
 from geonode.assets.models import (
     Asset,
     LocalAsset,
 )
-from rest_framework.exceptions import PermissionDenied
-from django.conf import settings
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


geonode/assets/serializers.py
```diff
@@ -122,7 +131,6 @@ def create(self, validated_data):
             raise serializers.ValidationError("Could not save the file.")
 
         if resource_id:
```
Gpetrak: @sijandh35  the `resource_id` is checked twice. I know that is for a different reason but it would be better  to define the resource = None before the resource_id checks and replace `resource_id` with `resource` in the second check. In my opinion, it helps in code readability and maintenance.

geonode/upload/uploadhandler.py
```diff
@@ -147,24 +155,9 @@ def _handle_raw_input_without_file_stream(self, input_data, META, raw_content_le
 
                 _post.appendlist(field_name, force_str(data, encoding, errors="replace"))
             elif item_type == FILE:
-                # NOTE: Parse files WITHOUT a stream.
```
Gpetrak, line 150: I 'm wondering if the following change should properly handled by a separate more related issue. @giohappy what you think ?

geonode/assets/serializers.py
sijandh35: @Gpetrak updated.

mattiagiupponi:
Hi @sijandh35 which is the status of this PR? did you evaluate the comments?

geonode/assets/views.py
```diff
@@ -80,6 +90,12 @@ def list(self, request, *args, **kwargs):
         serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data)
 
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save(owner=self.request.user)
```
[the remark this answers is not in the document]
sijandh35: Need to pass it for AssetSerializer where owner is set to required False which can accept empty owner from the form and  owner need to set from login user.

sijandh35:
> Hi @sijandh35 which is the status of this PR? did you evaluate the comments?

Hi @mattiagiupponi , I’ve made the changes and commented on reviews also. I’ve re-requested your review.

geonode/assets/serializers.py
```diff
@@ -60,19 +70,89 @@ def get_attribute(self, instance):
...
+                raise PermissionDenied("The user does not have permissions to change the resource selected")
+
+        if not title:
+            title = file.name
+
+        handler = LocalAssetHandler()
+        asset_dir = handler._create_asset_dir()
+
+        storage_manager = StorageManager(remote_files={"file": file})
+        storage_manager.clone_remote_files(cloning_directory=asset_dir, create_tempdir=False)
```
etj: Gemini is right: pls do not call `_create_asset_dir()` explicitly: [`LocalAssetHandler.create()`](https://github.com/GeoNode/geonode/blob/4.4.3/geonode/assets/local.py#L59) automatically creates the dir and copies the file in the asset dir.

geonode/assets/serializers.py
```diff
@@ -60,19 +70,89 @@ def get_attribute(self, instance):
...
+                logger.debug("The user does not have permissions to change the resource selected")
+                raise PermissionDenied("The user does not have permissions to change the resource selected")
+
+        if not title:
+            title = file.name
+
+        handler = LocalAssetHandler()
+        asset_dir = handler._create_asset_dir()
+
+        storage_manager = StorageManager(remote_files={"file": file})
```
etj: Please **do not** use the StorageManager, use only the AssetHandler methods. 
LocalAssetHander works on the filesystem as the StorageManager, but when using assets you may not know if they share any back storage; using the storageManager prevents switching the AssetHandler and making the asset moved somewhere else.

geonode/assets/serializers.py
```diff
@@ -60,19 +70,89 @@ def get_attribute(self, instance):
...
+            if not permissions_registry.user_has_perm(
+                user, resource.get_self_resource(), "change_resourcebase", include_virtual=True
+            ):
+                logger.debug("The user does not have permissions to change the resource selected")
+                raise PermissionDenied("The user does not have permissions to change the resource selected")
+
+        if not title:
+            title = file.name
+
+        handler = LocalAssetHandler()
```
etj: We should use `AssetHandlerRegistry.get_default_handler()`

etj:
This PR has triggere some internal discussion, which has been reported in https://github.com/GeoNode/geonode/issues/13423.
I guess we may hold this PR until #13424 has been implemented.
@giohappy can you confirm?

giohappy:
ok @etj

giohappy:
@sijandh35 https://github.com/GeoNode/geonode/issues/13424 has been merged. 
Please fix the conflict and complete it.

sijandh35:
@giohappy , the conflict has been fixed and updated the PR for review.

geonode/assets/utils.py
```diff
@@ -98,27 +127,32 @@ def create_asset_and_link(
...
     asset = link = None
     try:
-        default_title, default_ext = os.path.splitext(next(f for f in files)) if len(files) else (None, None)
+        first_file = next(iter(files)) if len(files) else None
+        filename = (
+            first_file
+            if isinstance(first_file, (str, Path))
+            else getattr(first_file, "name", None) if first_file else None
+        )
+        default_title, default_ext = os.path.splitext(filename) if filename else (None, None)
```
giohappy: @sijandh35 why do you set the `default_title` to `None` in case we cannot get the title from the `filename`, then you make a triple or in `create_asset` and set to "Unkknow" if default_title is `None`?
Can't you just set `default_title` to "Unknonw" here?

geonode/assets/serializers.py
```diff
@@ -60,19 +68,81 @@ def get_attribute(self, instance):
...
+            "title": {"required": False},
+            "description": {"required": False},
+            "type": {"required": False},
+            "created": {"required": False, "read_only": True},
+        }
 
 
 class LocalAssetSerializer(AssetSerializer):
+    file = serializers.FileField(write_only=True, required=True)
+    resource_id = serializers.IntegerField(required=False)
```
etj: I would not bind the concept of resource to an Asset.
Assets may be related to one or more resources, but may also be standalone (to be bound later).
The upload endpoint should:
 - read the info about the asset (files, etc...)
 - read the info about the related resource (if any)
 - get the default asset handler
 - use the handler to create the asset out of the passed data
 - link the new asset to the resource

geonode/assets/views.py
```diff
@@ -56,6 +61,11 @@ class AssetViewSet(DynamicModelViewSet):
     serializer_class = AssetSerializer  # TODO: appropriate Serializer should be switched for each Asset instance
     pagination_class = GeoNodeApiPagination
 
+    def get_serializer_class(self):
+        if self.request.method == "POST" and "file" in self.request.FILES:
+            return LocalAssetSerializer
```
etj: Pls don't instantiate directly the serializer type, since the system may be configured to use other kinds of Assets.
There is a registry call to get the configured Asset type:
https://github.com/GeoNode/geonode/blob/0b50d16d83df5a409df34a2d49e44419d04b9ecb/geonode/assets/handlers.py#L77
Pls create the new Asset by using the generic AssetHandler methods.

etj (changes requested):

geonode/assets/serializers.py
giohappy: @etj I see your point and I would move have implemented two distinct endpoints:

 - a generic assets viewset that manages "pure" assets, without any linkage
 - an action under the resources API viewset, that accepts an asset and performs the linkage to that resource. By the way this change would only move the discussion

We need the upload and linkage to be atomic operations, in a single request. A vew has an associated serializer.
If we don't want the resource id to be inside the LocalAssetSerializer (that should be "pure", without linkage) we would still need a serializer that performs the creation of the asset and the linkage atomically. How would you approach this?

sijandh35:
Hi @giohappy ,@etj, @mattiagiupponi ,

I am noting here the updated plan for changes in this PR, any suggestion and things i misunderstood let me know. The updated plan is to manage assets for resources through a dedicated action within the existing **ResourceBaseViewSet**. This ensures that creating, linking, and unlinking assets are atomic operations tied directly to a specific resource.

The key flows are:
1. **Creation**: Atomically create an asset and link it to a resource.
2. **Deletion**: Unlink an asset from a resource. If the asset is no longer linked to any other resources, it should be deleted, but with specific exceptions.
3. **Protected Assets**: Certain assets (e.g., the "Original" or migrated "Files" asset) must be protected from deletion.
4. **Implementation**: The core logic should reside in `geonode/assets/utils.py` and be called from the ResourceViewSet through handlers.

## Proposed Update for the Pull Request

Here is the plan to follow to refactor the PR and implement the feature as below:

### 1. API Endpoint Definition
A new action on the ResourceBaseViewSet will be available at the following URL:
* **URL**: `/api/v2/resources/<pk>/assets/`

This endpoint supports two HTTP methods:
* **POST**: To upload a new file, create an asset, and link it to the resource identified by `<pk>`.
* **DELETE**: To unlink an asset from the resource. The asset to be unlinked is identified by an `asset_id` in the request body.

### 2. Refactor `ResourceViewSet`
* Revert the currently changed `AssetsViewSet` and its corresponding Serializer that were previously added in this PR.
* In `geonode.base.api.views.ResourceBaseViewSet`, add a new method decorated with `@action(detail=True, methods=['post', 'delete'])`.

**POST Logic**:
* The method performs permission checks to ensure the requesting user has the right to modify the resource.
* It validates the uploaded file against supported types.
* File size limit will be checked already from the uploadhandler.
* It calls the existing `create_asset_and_link` function which has handler from `geonode/assets/utils.py` to handle the file storage, asset creation, and linking to the resource.

**DELETE Logic**:
* The method expects an `asset_id` in the request payload.
* It calls a new utility function, `unlink_asset(resource_id, asset_id, remove_asset=True)`, which is added to `geonode/assets/utils.py` as mentioned in [ISSUE](https://github.com/GeoNode/geonode/issues/13378#issuecomment-3234014876)

### 3. Implement Logic in `geonode/assets/utils.py`
Create a new function to handle the complex deletion logic: `unlink_asset(resource_id, asset_id, remove_asset=True)`. This function performs the following checks in order:

1. **Check for Protected Assets**:
  * If asset name is 'Original' or asset name is 'Files', the function raises an error (or exception) and does not proceed. This check noted [ISSUE](https://github.com/GeoNode/geonode/issues/13378#issuecomment-3234014876).

2. **Handle Unlinking**:
  * The function removes the link between the specified resource and asset.

3. **Handle Asset Deletion**:
  * After unlinking, it checks if the asset is still linked to any other resources.
  * If it is not linked to any other resources, the asset file itself is deleted.

## Clarification and Suggestions Needed:
**1. Is there should be a method somewhere to create an independent asset without the resource id?**
**2. Should there be an action in ResourceViewSet where it links an already created asset to a specified resource ID?**

giohappy:
>1. Is there should be a method somewhere to create an independent asset without the resource id?

For the moment no. For this AssetsViewSet will be implemented again in the future

>2. Should there be an action in ResourceViewSet where it links an already created asset to a specified resource ID?

For the moment no.

IMHO The DELETE method should work on a path that includes the asset id, instead of passing it inside the body.
I mean, DELETE should be done for the path `api/v2/resources/<resource_id>/assets/<asset_id>`.

@sijandh35 we must also clarify how errors or warnings will be returned. It should use the same approach as with the other APIs (message structure inside the response body). The two cases where messages are required are:

- the requested asset_id doesn't exist 
- the asset was unlinked but it couldn't be deleted

sijandh35:
> IMHO The DELETE method should work on a path that includes the asset id, instead of passing it inside the body. I mean, DELETE should be done for the path `api/v2/resources/<resource_id>/assets/<asset_id>`.

 Okay @giohappy will make it like this.
 
> @sijandh35 we must also clarify how errors or warnings will be returned. It should use the same approach as with the other APIs (message structure inside the response body). 

Okay will return the errors or warning like how in other APIs is done inside Response body with error message and status code.

geonode/base/api/views.py
```diff
@@ -1415,6 +1419,85 @@ def linked_resources(self, request, pk, *args, **kwargs):
...
+            )
+
+        if not permissions_registry.user_has_perm(user, resource, "change_resourcebase", include_virtual=True):
+            return Response(
+                {"message": "You do not have permission to add an asset to this resource."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        if file and not os.path.splitext(file.name)[1].lower()[1:] in settings.ALLOWED_DOCUMENT_TYPES:
+            logger.debug("This file type is not allowed")
```
etj: Please add info about what is not allowed

geonode/assets/utils.py
```diff
@@ -167,3 +201,50 @@ def rollback_asset_and_link(asset, link):
...
+        logger.info(msg)
+        return True, msg
+
+    # Check if the asset is linked to other resources
+    other_links_count = Link.objects.filter(asset=asset).exclude(pk=link.pk).count()
+
+    if other_links_count > 0:
+        # Only delete the link
+        link.delete()
+        msg = f"Asset {asset.pk} was unlinked but could not be deleted."
```
etj: The "could not" seems there are some unknown reasons. Let's log why it was not removed
```
msg = f"Asset {asset.pk} was unlinked but not deleted, because still linked to {other_links_count} resources"
```

geonode/assets/utils.py
```diff
@@ -84,10 +85,38 @@ def create_link(resource, asset, link_type=None, extension=None, name=None, mime
...
+            description=description or asset_type or "Unknown",
+            type=asset_type or "Unknown",
+            owner=owner,
+            files=files,
+            clone_files=clone_files,
+        )
+        return asset
+    except Exception as e:
+        logger.error(f"Error creating Asset: {e}", exc_info=e)
+        if asset:
```
etj: if an exception is raised within the `try` block, `asset` will neevr be assigned

geonode/assets/utils.py
```diff
@@ -98,27 +127,32 @@ def create_asset_and_link(
...
+            description=description,
+            asset_type=asset_type,
             clone_files=clone_files,
         )
 
         link = create_link(
             resource,
             asset,
-            asset_handler=asset_handler,
+            asset_handler=handler,
```
etj: you removed line 101 
```
asset_handler = handler or asset_handler_registry.get_default_handler()
``` 
when you refactored out the `create_asset' method, so here handler may be null!

geonode/base/api/views.py
```diff
@@ -1415,6 +1419,85 @@ def linked_resources(self, request, pk, *args, **kwargs):
...
+            return Response({"message": "The uploaded file type is not allowed."}, status=status.HTTP_400_BAD_REQUEST)
+        try:
+            handler = asset_handler_registry.get_default_handler()
+            asset, link = create_asset_and_link(
+                resource, user, [file], handler=handler, title=title, description=description, clone_files=False
+            )
+            if asset and link:
+                return Response(
+                    {"message": "Asset created and linked successfully.", "asset_id": asset.id},
```
etj: i would also return the `link_id`, since it has just been created as an effect of the POST call

etj (changes requested):

geonode/base/api/views.py
```diff
@@ -1415,6 +1419,85 @@ def linked_resources(self, request, pk, *args, **kwargs):
...
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        if file and not os.path.splitext(file.name)[1].lower()[1:] in settings.ALLOWED_DOCUMENT_TYPES:
+            logger.debug("This file type is not allowed")
+            return Response({"message": "The uploaded file type is not allowed."}, status=status.HTTP_400_BAD_REQUEST)
+        try:
+            handler = asset_handler_registry.get_default_handler()
+            asset, link = create_asset_and_link(
+                resource, user, [file], handler=handler, title=title, description=description, clone_files=False
```
etj: `clone_files` is not used when files are in-memory.
I would remove this useless argument in order to make things clearer

geonode/base/api/views.py
```diff
@@ -1415,6 +1419,85 @@ def linked_resources(self, request, pk, *args, **kwargs):
...
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        if not permissions_registry.user_has_perm(user, resource, "change_resourcebase", include_virtual=True):
+            return Response(
+                {"message": "You do not have permission to add an asset to this resource."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        if file and not os.path.splitext(file.name)[1].lower()[1:] in settings.ALLOWED_DOCUMENT_TYPES:
```
mattiagiupponi: Hi @sijandh35, the asset itself can be anything that is accepted in GeoNode, this includes also the dataset. Here other than checking for the `ALLOWED_DOCUMENT_TYPES`, we should also call `get_supported_datasets_file_types` (from the formats->required_ext) and be sure that the files extention is there

geonode/base/api/views.py
```diff
@@ -1415,6 +1419,85 @@ def linked_resources(self, request, pk, *args, **kwargs):
...
+        permission_classes=[UserHasPerms(perms_dict={"default": {"POST": ["base.add_resourcebase"]}})],
+        parser_classes=[JSONParser, MultiPartParser],
+        url_name="assets",
+    )
+    def asset(self, request, pk=None, *args, **kwargs):
+        """Handles uploading a new file, creating an asset, and linking it to the resource."""
+
+        if "file" not in request.FILES:
+            return Response({"message": "File not provided."}, status=status.HTTP_400_BAD_REQUEST)
+        file = request.FILES["file"]
```
mattiagiupponi: The `asset` itself can be composed by multiple files, here we should manage the file as list rather than single file, later in the code, we should evaluate that all the files uploaded are accepted by the settings

geonode/base/api/views.py
```diff
@@ -1415,6 +1419,88 @@ def linked_resources(self, request, pk, *args, **kwargs):
...
+        resource = get_object_or_404(ResourceBase, pk=pk)
+        user = request.user
+        title = request.data.get("title", None)
+        description = request.data.get("description", None)
+
+        forbidden_titles = {"Original", "Files"}
```
etj, line 1442: Please add a comment telling that this is a temporary solutions and that a proper flag will be added later

etj (approved):

PR #13436, opened 2025-08-20
[Fixes #13424] Make LocalAssetHandler deal with in-memory files

### LINKED ISSUE #13424: Make LocalAssetHandler deal with in-memory files
When creating an Asset, the LocalAssetHandler logic should be able to tell the kind of files passed as argument,
At the moment it only allows files in the filesystem; it should as a minimum also deal with in-memory uploaded files.

sijandh35:
Fixes #13424 

Changes:

- Update create method in LocalAssetHandler to support InMemoryFile
- Added test case to test in memory file support.

## Checklist

For all pull requests:

### DIFF
--- a/geonode/assets/local.py
+++ b/geonode/assets/local.py
@@ -3,11 +3,13 @@ import logging
 import os
 import shutil
 
+
 from django.conf import settings
 from django.http import HttpResponse, StreamingHttpResponse
 from django.urls import reverse
 from django_downloadview import DownloadResponse
 from zipstream import ZipStream
+from pathlib import Path
 
 from geonode.assets.handlers import asset_handler_registry, AssetHandlerInterface, AssetDownloadHandlerInterface
 from geonode.assets.models import LocalAsset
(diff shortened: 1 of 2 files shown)


geonode/assets/local.py
```diff
@@ -62,6 +62,18 @@ def create(self, title, description, type, owner, files=None, clone_files=False,
 
         if clone_files:
             files = self._copy_data(files)
+        elif isinstance(files[0], str):
```
mattiagiupponi: What if the second file in the list is a "InMemoryFile"? I think here is better to loop on the list rather then accessing to the index 0

geonode/assets/local.py
```diff
@@ -62,6 +62,18 @@ def create(self, title, description, type, owner, files=None, clone_files=False,
...
+            pass
+        else:
+            # if files are in-memory, need to handle them properly
+            new_path = self._create_asset_dir()
+            relative_dir = os.path.relpath(new_path, os.path.dirname(settings.ASSETS_ROOT))
+            processed_files = []
+            for f in files:
+                save_path = os.path.join(relative_dir, f.name)
+                self.get_storage_manager(None).save(save_path, f)
```
mattiagiupponi: @etj do u think is better to use the `save` here or the `clone_remote_files`? The `save` method is not passing by the [DataRetriever](https://github.com/GeoNode/geonode/blob/423029f503c96b39bee93e6982d59dd546ed7b68/geonode/storage/data_retriever.py#L146) so the permissions, unzipping and [transfer in chunk](https://github.com/GeoNode/geonode/blob/423029f503c96b39bee93e6982d59dd546ed7b68/geonode/storage/data_retriever.py#L86) is not used

mattiagiupponi (changes requested):

geonode/assets/local.py
```diff
@@ -62,6 +62,18 @@ def create(self, title, description, type, owner, files=None, clone_files=False,
 
         if clone_files:
```
etj: `clone_files` should only be checked if the item in 'files` is a file or a directory.

geonode/assets/local.py
etj: We may assume that all the `files` are the same type (filesystem objects, inmemoryfile, etc).
Since we are adding in this PR the logic about handling different input, we may want to check for this assumption as a first step in the method.

geonode/assets/local.py
etj: `clone_remote_files` for sure, ohterwise we're missing the point of the issue.

PR #13535, opened 2025-09-23
[Fixes #13520] Improved Error Robustness for special characters and whitespaces in filenames and attribute tables

### LINKED ISSUE #13520: Improve Error Robustness for special characters and whitespaces in filenames and attribute tables
There are some problems when uploading files that contain special characters in the filename or in the attribute columns. These cases need to be properly handled, or the user should be shown a clear reason for the failure or error:

**Attribute Table (Column Names)**
Vector files (GeoJSON or GeoPackage) can also have issues if attribute (column) names contain whitespaces or special characters. This can lead to problems later when using filters, editing attributes, or updating data.
Examples of problematic characters include: +, @, #, %, á, é, Ö, etc.

**Filenames**
Currently, there are problems with files that contain various special characters. Failed uploads do not have a good error message. For users it should be clear which ones can and cannot be uploaded and in case something cannot be uploaded a clear error message should occur.
Characters like Á, É, í, Ö, Ä, Ü, §,´ can cause issues. Some of these are partially handled during upload using the [ `fixup_name`](https://github.com/GeoNode/geonode/blob/e21da3c0d8a628133ec6c6e9ebae0d5e8b50fdf9/geonode/upload/handlers/base.py#L217) method, but this needs to be extended more or accompanied by proper error messages.

sijandh35:
Fixes #13520 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/upload/handlers/base.py
+++ b/geonode/upload/handlers/base.py
@@ -234,9 +234,8 @@ class BaseHandler(ABC):
         We use replace because it looks to be one of the fasted options:
         https://stackoverflow.com/questions/3411771/best-way-to-replace-multiple-characters-in-a-string
         """
-        prefix = name[0]
-        if prefix.isnumeric():
-            name = name.replace(name[0], "_")
+        if name and name[0].isnumeric():
+            name = f"_{name[1:]}"
         return (
             name.lower()
             .replace("-", "_")
@@ -247,7 +246,30 @@ class BaseHandler(ABC):
             .replace(")", "")
             .replace("(", "")
             .replace(",", "")
-            .replace("&", "")[:62]
+            .replace("&", "_")
+            .replace("!", "")
+            .replace('"', "")
+            .replace("$", "")
+            .replace("%", "")
+            .replace("'", "")
+            .replace("*", "")
+            .replace("+", "")
+            .replace("/", "")
+            .replace(":", "")
+            .replace(";", "")
+            .replace("<", "")
+            .replace("=", "")
+            .replace(">", "")
+            .replace("?", "")
+            .replace("@", "")
+            .replace("[", "")
+            .replace("]", "")
+            .replace("^", "")
+            .replace("`", "")
+            .replace("{", "")
+            .replace("|", "")
+            .replace("}", "")
+            .replace("~", "")[:62]
         )
 
     def extract_resource_to_publish(self, files, layer_name, alternate, **kwargs):
--- a/geonode/upload/handlers/base.py
+++ b/geonode/upload/handlers/base.py
@@ -22,6 +22,7 @@ import os
 from pathlib import Path
 from typing import List
 import zipfile
+import re
 
 from geonode.assets.utils import create_asset_and_link
 from geonode.resource.enumerator import ExecutionRequestAction as exa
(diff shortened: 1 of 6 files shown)


geonode/upload/handlers/common/vector.py
```diff
@@ -628,6 +638,39 @@ def create_dynamic_model_fields(
...
+        vrt_content += """  </OGRVRTLayer>
+        </OGRVRTDataSource>
+        """
+
+        # Write to a temporary file that ogr2ogr will use will be deleted after use
+        vrt_fd, vrt_filename = tempfile.mkstemp(suffix=".vrt")
+        with os.fdopen(vrt_fd, "w") as f:
+            f.write(vrt_content)
+
+        return vrt_filename, vrt_layer_name
```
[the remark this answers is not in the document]
sijandh35: Updated.

geonode/upload/handlers/common/vector.py
```diff
@@ -1329,6 +1372,10 @@ def import_with_ogr2ogr(
 
         process = Popen(" ".join(commands), stdout=PIPE, stderr=PIPE, shell=True)
         stdout, stderr = process.communicate()
+
+        if files.get("vrt_file"):
```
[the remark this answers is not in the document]
sijandh35: Updated.

geonode/upload/handlers/common/vector.py
```diff
@@ -260,7 +262,9 @@ def create_ogr2ogr_command(files, original_name, ovverwrite_layer, alternate):
                 _datastore["USER"],
                 _datastore["PASSWORD"],
             )
-        options += f'"{files.get("base_file")}"' + " "
+
+        input_file = files.get("vrt_file") or files.get("base_file")
```
mattiagiupponi: i didnt get this @sijandh35 why the vrt file and the base file are interchangeable?

geonode/upload/handlers/common/vector.py
```diff
@@ -628,6 +639,40 @@ def create_dynamic_model_fields(
...
+        layer_defn = layer.GetLayerDefn()
+        for i in range(layer_defn.GetFieldCount()):
+            field_defn = layer_defn.GetFieldDefn(i)
+            original_name = field_defn.GetName()
+            sanitized_name = self.fixup_name(original_name)
+            field_type = field_defn.GetTypeName()
+            vrt_content += f'    <Field name="{escape(sanitized_name)}" src="{escape(original_name)}" type="{escape(field_type)}" />\n'
+
+        vrt_content += """  </OGRVRTLayer>
+        </OGRVRTDataSource>
```
mattiagiupponi: Did you evaluate whether to define an XML file as a Django template? If so, what is the reason for this decision?

geonode/upload/handlers/base.py
```diff
@@ -247,7 +247,30 @@ def fixup_name(self, name):
             .replace(")", "")
             .replace("(", "")
             .replace(",", "")
-            .replace("&", "")[:62]
+            .replace("&", "_")
```
mattiagiupponi: can we evaluate a better alternative than replace since we have so many chars now? maybe a `maketrans` or a `re.sub` si more efficient

geonode/upload/handlers/common/vector.py
```diff
@@ -628,6 +639,40 @@ def create_dynamic_model_fields(
...
+          <OGRVRTLayer name="{escape(vrt_layer_name)}">
+            <SrcDataSource>{escape(source_filepath)}</SrcDataSource>
+            <SrcLayer>{escape(layer.GetName())}</SrcLayer>
+        """
+
+        # Map original field names to sanitized names using fixup_name
+        layer_defn = layer.GetLayerDefn()
+        for i in range(layer_defn.GetFieldCount()):
+            field_defn = layer_defn.GetFieldDefn(i)
+            original_name = field_defn.GetName()
```
mattiagiupponi: is there a way to retrieve all the field names and type in a easier way like we do fo the dynamic_model? is easier to read https://github.com/GeoNode/geonode/blob/dafd236b858ba2fcb0410b8e2aa119e14b622800/geonode/upload/handlers/common/vector.py#L597

geonode/upload/handlers/common/vector.py
sijandh35: Hi @mattiagiupponi, I added that logic as a fallback in case the vrt_file isn't passed due to an error during its creation in _create_vrt_file, just to allow the upload to proceed using the existing base_file flow. But you're right,I've now realized I should also handle the exception properly and avoid passing the vrt_file key when its creation fails, to ensure the flow continues with base_file.

mattiagiupponi: Let's avoid to check if the user send a vrt file for now. Let's just rely on the one we manage

geonode/upload/handlers/common/vector.py
sijandh35: Updated to be used XML file as a Django template, and it is now called from the create_vrt_file function inside utils.

geonode/upload/handlers/base.py
sijandh35: Updated.

geonode/upload/handlers/common/vector.py
sijandh35: Updated.

geonode/upload/handlers/common/vector.py
```diff
@@ -414,6 +418,11 @@ def import_resource(self, files: dict, execution_id: str, **kwargs) -> str:
                     )
                     # and layer.GetGeometryColumn() is not None
                 ):
+                    vrt_filename, vrt_layer_name = create_vrt_file(layer, files.get("base_file"))
```
mattiagiupponi: @sijandh35 let's keep the generation of the vrt file as optional in the workflo. We can evaluate the attributes, if one of them needs to be modified, we can generate the vrt file, otherwise we can proceed with the normal workflow

mattiagiupponi (changes requested):

geonode/upload/handlers/common/vector.py
sijandh35: Updated.

mattiagiupponi (approved):

Issue #13541, opened 2025-09-24
Upsert improvments / refactor


mattiagiupponi:
Following the initial development started with https://github.com/GeoNode/geonode/issues/13361 for the upsert functionality, we would like to asset the following changes / improvments:

The current approach —partial inserts with early failure on validation error— is not ideal. The user will need to retry the upserts several times to discover all possible errors

- Proposal: **Two-step process** to ensure data consistency and improve user experience:
   1. **Pre-validation Phase**:
      - Validate the entire file before any insertion.
      - If errors are found, generate a **limited error report** (CSV or Excel), capped at ~100–1000 entries, where for each invalid entry, a row in the report is created explaining what was wrong.
      - User can fix the problematic rows and retry.
   2. **Atomic Insert Phase**:
      - If validation passes, insert the full dataset in one operation, otherwise the insert will not be executed
      - Avoid row-by-row processing for efficiency and robustness
 
 - Benefits:
   - **Consistent state** (no partial inserts).
   - Clear feedback loop for users.
   - Avoid trial-and-error resubmissions.

PR #13542, opened 2025-09-24
[Fixes #13451] Upsert improvments / refactor

### LINKED ISSUE #13451: broken upload datasets when asset have duplicated name from different users.
When uploading a raster file, it is processed and stored correctly.
However, if a user uploads again a file with the same name, the upload still succeeds.
The problem occurs when another user tries to upload a file with a name that has already been uploaded before (whether it is the exact same file or a different one with the same name).

This happens on geonode 4.4.x branch with geonode importer 1.1.x. branch, installed using geonode-project

In this case:

The system raises an error saying the file already exists in the workspace.

It also deletes the file from the filesystem, breaking the existing dataset.

If the user retries the upload, a second error is raised indicating that the resource is duplicated.

Steps to reproduce

User A uploads a raster file example.tif → success.

User A uploads again example.tif → success.

User B uploads example.tif (same name, same or different file) → error occurs, existing dataset is broken.

User B retries the upload → duplicate error

after the first error the dataset is removed from geoserver for the formerly uploaded dataset from user A.

Expected behavior

The system should handle raster uploads with existing names gracefully.

Uploads from different users with the same filename should not break existing datasets or remove files from the filesystem.

Image

Image

Image

Image

mattiagiupponi:
ref #13541 
## Checklist

For all pull requests:


mattiagiupponi:
> @mattiagiupponi when I'm testing an upsert failure using a shapefile, and then I'm trying to download the CSV log file, instead of the log file, it downloads the dataset itself (the shapefile)

fixed

geonode/upload/handlers/common/vector.py
```diff
@@ -1172,6 +1178,7 @@ def _process_features(
         # looping over the chunk data
         for feature in data_chunk:
             feature_as_dict = feature.items()
+            errors.append(feature_as_dict | {"reason": "The value X is invalid"})
```
Gpetrak: @mattiagiupponi I guess we should remove this line right ? Beyond this, now the assets list, includes the downloadable error log in CSV

Gpetrak (commented):
Please check the comment.

Gpetrak (commented):
@mattiagiupponi now it works fine to me. But I think it worths to add a test which tests the new functionality. What do you think ?

mattiagiupponi:
> @mattiagiupponi now it works fine to me. But I think it worths to add a test which tests the new functionality. What do you think ?

yes i'm going to! :)

Issue #13554, opened 2025-10-01
Allow the upsert process to insert only the valid feature


sijandh35:
Once the `FeatureConstraingRegistry` is completed, the registry can be called in during the upsert method in the importer.
- The implementation should be available only for vector filetypes.

The workflow will be:
- User upload the data for the upsert
- During the processing of the feature the method should also call the registry and then the validate_feature method to ensure that the row is valid. 

If the row is valid, it can save the feature in the DB (`bulk_create` or `bulk_update`)

Issue #13553, opened 2025-10-01
Implement the FeatureConstraingRegistry


sijandh35:
Implement a registry where is possible to plug the customized handlers in order to enrich the feature validation during the upsert process.

The registry should implement a method named "validate" which will iterate on all the handler implemented, where the feature is passed and validated.
An exception should be raised if the validation is not passed.

PR #13573, opened 2025-10-07
[Fixes #13552] Implementation for validation for upserts

### LINKED ISSUE #13552: Implement validation for upserts
One of the goals of the parent epic is to implement validation when data inside a vector dataset is updated.
Validation is guided by constraints that are set on the vector fields through Geoserver.
When a restriction has been defined for an attribute in Geoserver, the configuration for that constraint is returned inside the DescripbeFeatureType response.
GeoNode will implement a generic validation pass when a row is about to be upserted/inserted in an existing vector dataset.
The validation will cycle over implemented validators. In this case, the featuretype validator will be the first and only one, but could have additional (and custom) validators in the future. 

At a high level, the validator will retrieve the constraints from DescripbeFeatureType and check that the field values for the current row respect them

The types of constraints are two:

- range values, for numeric fields
- enumerators, for numeric and string fields.

If the validation fails for a row, the upsert operations will stop and report to the client the validation error for the row (and, eventually, the number of rows that were successfully inserted up to that point).

sijandh35:
This PR fixes parent #13552, which includes two sub-issues: #13553 and #13554. Both have been addressed in this PR, as #13554 is dependent on the earlier sub-issue.

## Checklist

For all pull requests:

### DIFF
--- a/geonode/upload/registry.py
+++ b/geonode/upload/registry.py
@@ -19,12 +19,12 @@
 
 from django.conf import settings
 from django.utils.module_loading import import_string
-from geonode.upload.feature_constraint_handlers import BaseFeatureConstraintHandler
+from geonode.upload.feature_validators import BaseFeatureValidator
 
 
-class FeatureConstraintRegistry:
+class FeatureValidatorsRegistry:
     """
-    A registry for feature constraint handlers.
+    A registry for feature validation.
     """
 
     REGISTRY = []
@@ -59,7 +59,7 @@ class FeatureConstraintRegistry:
 
     @classmethod
     def get_registry(cls):
-        return FeatureConstraintRegistry.REGISTRY
+        return FeatureValidatorsRegistry.REGISTRY
 
     def sanity_checks(self):
         for item in self.REGISTRY:
@@ -74,14 +74,14 @@ class FeatureConstraintRegistry:
 
     def __check_item(self, item):
         """
-        Ensure that the handler is a subclass of BaseFeatureConstraintHandler
+        Ensure that the handler is a subclass of BaseFeatureValidator
         """
-        if not issubclass(item, BaseFeatureConstraintHandler):
-            raise Exception(f"Handler {item} is not a subclass of BaseFeatureConstraintHandler")
+        if not issubclass(item, BaseFeatureValidator):
+            raise Exception(f"Handler {item} is not a subclass of BaseFeatureValidator")
 
     def _register(self):
-        for module_path in settings.FEATURE_CONSTRAINT_HANDLERS:
+        for module_path in settings.FEATURE_VALIDATORS:
             self.add(module_path)
 
 
-feature_constraint_registry = FeatureConstraintRegistry()
+feature_validators_registry = FeatureValidatorsRegistry()
--- a/geonode/settings.py
+++ b/geonode/settings.py
@@ -2304,8 +2304,8 @@ PERMISSIONS_HANDLERS = [
     "geonode.security.handlers.AdvancedWorkflowPermissionsHandler",
 ]
 
-FEATURE_CONSTRAINT_HANDLERS = [
-    "geonode.upload.feature_constraint_handlers.GeoserverFeatureConstraintHandler",
+FEATURE_VALIDATORS = [
+    "geonode.upload.feature_validators.GeoserverFeatureValidator",
 ]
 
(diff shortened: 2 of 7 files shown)


geonode/upload/feature_constraint_handlers.py
```diff
@@ -0,0 +1,110 @@
...
+        if min_val is not None and float(value) < float(min_val):
+            raise ValidationError(
+                f'Invalid value for attribute "{attribute_name}". '
+                f"Value {value} is less than the minimum allowed value of {min_val}."
+            )
+        if max_val is not None and float(value) > float(max_val):
+            raise ValidationError(
+                f'Invalid value for attribute "{attribute_name}". '
+                f"Value {value} is greater than the maximum allowed value of {max_val}."
+            )
```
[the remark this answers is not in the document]
sijandh35: Updated.

mattiagiupponi:
@sijandh35 when the geoserver-restconfig will be merged, we need to update the requirements.txt and setup.cfg by let it point to the master branch `git+https://github.com/GeoNode/geoserver-restconfig.git@master#egg=geoserver-restconfig`

sijandh35:
> @sijandh35 when the geoserver-restconfig will be merged, we need to update the requirements.txt and setup.cfg by let it point to the master branch `git+https://github.com/GeoNode/geoserver-restconfig.git@master#egg=geoserver-restconfig`

Okay @mattiagiupponi will update the PR after merge.

> `['[\'Invalid value for attribute "ogc_fid". Value 90 is greater than the maximum allowed value of 10.0.\']', '[\'Invalid value for attribute "ra_cira". Value "6" is not in the allowed enumeration: [\\\'1\\\'].\']', '[\'Invalid value for attribute "ra_nome". Value "UPSERTED NAME" is not in the allowed enumeration: [\\\'abc\\\', \\\'bece\\\'].\']']`
> 
> as you can see is a list of lists translated in string. I would suggest to flatten the list before return the errors, maybe with a `|` as a separator in order to not be confused during the CSV parsing (text to column)

Updated, now should looks like this eg:
`Invalid value for attribute "st_area_shape". Value "132456.35" is not in the allowed enumeration: ['1.0']. | Invalid value for attribute "st_length_shape". Value 134.456 is greater than the maximum allowed value of 2.0.`

geonode/upload/feature_constraint_handlers.py
```diff
@@ -0,0 +1,119 @@
...
+
+import logging
+
+from geonode.geoserver.helpers import gs_catalog
+from django.core.exceptions import ValidationError
+
+logger = logging.getLogger(__name__)
+
+
+class BaseFeatureConstraintHandler:
```
giohappy: I would name the class `BaseFeatureValidator`.
This will make room for different kinds of validations.

geonode/upload/feature_constraint_handlers.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def validate(self, feature):
+        """
+        Validates the feature.
+        Should raise an exception if validation fails.
+        """
+        raise NotImplementedError
+
+
+class GeoserverFeatureConstraintHandler(BaseFeatureConstraintHandler):
```
giohappy: And this will be `GeoserverFeatureValidator`

geonode/upload/registry.py
```diff
@@ -0,0 +1,87 @@
...
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+#########################################################################
+
+from django.conf import settings
+from django.utils.module_loading import import_string
+from geonode.upload.feature_constraint_handlers import BaseFeatureConstraintHandler
+
+
+class FeatureConstraintRegistry:
```
giohappy: Let's rename this to `FeatureValidatorsRegistry`

geonode/upload/feature_constraint_handlers.py
sijandh35: Renamed the class to BaseFeatureValidator to support various types of validations. Updated newly created file names to align with the class name and adjusted new tests to match the renamed class.

geonode/upload/feature_constraint_handlers.py
sijandh35: Updated.

geonode/upload/registry.py
sijandh35: Updated.

Issue #13578, opened 2025-10-08
Implement permissions to control public and registered members permissions management


sijandh35:
In view of the removal of ADW from GeoNode (https://github.com/GeoNode/geonode/issues/13472) we want to replace its publishing and approval mechanism by implementing the option to control if the current GeoNode user can change or not the permissions for the two "special" groups, Anonymous, and Registered Members.

### Proposed Changes:

 - Introduce two new permission types
   - `can_manage_anonymous_permissions`
   - `can_manage_registered_member_permissions`
- These two permissions will be treated the same as the other permissions, through the Permissions Registry and the default permisisons handler and will be returned by the permissions serializers inside the API responses
-  By default the the resource owner and any user with edit permissions on the resource will be assigned both permissions
- Two new settings will allow for restricting the permissions. If set to False, only an administrator or a user with `user.is_staff` set will be assigned those permissions
  - `EDITORS_CAN_MANAGE_ANONYMOUS_PERMISSIONS`= False # default value
  - `EDITORS_CAN_MANAGE_REGISTERED_MEMBERS_PERMISSIONS` = False # default value

PR #13592, opened 2025-10-10
[Fixes #13581] Forbid API changing permissions for anonymous and registered members users if user is not allowed to

### LINKED ISSUE #13581: API will forbid changing permissions for anonymous and registered members users if user is not allowed to
Based on the implementation of #13578 ,the permissions endpoints will forbid users to change anoynmous and/or registered members permissions if they don't have can_manage_anonymous_permissions and/or can_manage_registered_member_permissions permissions.

sijandh35:
This PR Fixes #13581 

This PR should only be merged after the issue(#13578 ) is merged.

## Checklist

For all pull requests:


sijandh35:
> @sijandh35 Instead of doing all this logic (including retrieving the Group objects from the DB), shouldn't we just discard the anonymous and registered_members permissions from the perms payload in case the user doesn't have permissions to change it?

> 

> I mean:

> 

> * We check if the user doesn't have can_manage_anonymous_permissions or can_manage_registered_member_permissions

> * In that case, we remove the two groups from request.data.get("groups") without throwing any exception

> * The other changes will go through, but the permissions for the two groups remain unchanged

Hi @giohappy Updated it.

> Maybe we can also avoid the check with the grups id, and just rely on the group name? This would save a DB query.

Yes, using the group name would reduce the need for the query to fetch the group object. However, from what I see on the client side, when a permission is changed, a PUT request is made, and in the groups key it only sends the ID and permissions — for example: [{'id': 3, 'permissions': 'view'}, {'id': 2, 'permissions': 'edit'}]. That’s why I used the ID.

But if we can pass the group name also from client, that would be better.

geonode/base/api/views.py
```diff
@@ -615,29 +616,19 @@ def resource_service_permissions(self, request, pk, *args, **kwargs):
...
+            elif request.method in ["PUT", "PATCH"]:
+                user_perms = permissions_registry.get_perms(instance=resource, user=request.user)
+                if request.data.get("groups"):
+                    excluded_ids = []
+                    if "can_manage_anonymous_permissions" not in user_perms:
+                        anonymous_group = Group.objects.get(name="anonymous")
+                        excluded_ids.append(anonymous_group.id)
+                    if "can_manage_registered_member_permissions" not in user_perms:
+                        registered_group = Group.objects.get(name=groups_settings.REGISTERED_MEMBERS_GROUP_NAME)
+                        excluded_ids.append(registered_group.id)
```
mattiagiupponi, line 631: I was testing it, will not be better to raise an error if the user tries to change those two perms if it can't? Or at least having a log to notify this.
I'm looking in the log run we might confuse why some users are not able to change some perms.
The FE can easily lock the selection, but from API side we should notify the user somehow

giohappy: If the client always sends the permissions, even when the user doesn't change them, you would always receive an error.
Either the client does not include the permissions, when the user cannot change them, or we just let them through. Maybe we can just add an INFO log.

giohappy: @sijandh35 please add the log and let's close this

mattiagiupponi (approved):

Issue #13648, opened 2025-10-21
Multilang: handle i18n queriable fields in pycsw


etj:
CSW does not provide a way for performing multilang queries, so we may leave the query part as is; it means that the original fields (where the value of the field in the default language is copied) will be the one used for filtering.

In order to perform searches on all the languages at once with a decent performance, a new table CswFilters will be created, containing all the queriable values. The multilang fields (e.g. `title`) would contain the title in all the languages merged into a single text string.
There should be a db view joining ResourceBases and CswFilters, and pyCSW should be querying such view (it is possibile since pycsw3).

Everytime metadata are updated, a procedure should update the entries in CswFilters, coalescing the multilang values.

This issue is **optional** for the multilang feature.

Please note that the implementation of this issue would also fix the [bad configuration mapped on pycsw](https://github.com/GeoNode/geonode/blob/master/geonode/catalogue/backends/pycsw_local_mappings.py), since many required queriable elements are mapped on non-existent fields.

Issue #13647, opened 2025-10-21
Multilang: parse multilang entries in ISO19115


etj:
Parse the entries encoded in #13646.

The entries should be stored into jsonschema elements as described and implemented in #13642.

The created jsonschema will be stored in the change implemented at https://github.com/GeoNode/geonode/pull/13591/commits/7bdedaca1c2859f27bdd7e62dd0bc90abc1135d1

This issue should include parser refactoring as documented in https://github.com/GeoNode/geonode/issues/12902

Issue #13646, opened 2025-10-21
Multilang: use multilang entries in ISO19115


etj:
ISO19115 allows for multilingual text, so it will possible to encode the related ISO fields into the exported XML document

e.g.
```xml
<gmd:title xsi:type="gmd:PT_FreeText_PropertyType">
  <gco:CharacterString>Ilmatieteen laitos</gco:CharacterString>
  <gmd:PT_FreeText>
    <gmd:textGroup>
      <gmd:LocalisedCharacterString locale="#locale-swe">Meteorologiska institutet</gmd:LocalisedCharacterString>
    </gmd:textGroup>
    <gmd:textGroup>
      <gmd:LocalisedCharacterString locale="#locale-en">Finnish Meteorological Institute</gmd:LocalisedCharacterString>
    </gmd:textGroup>
  </gmd:PT_FreeText>
</gmd:title>
```

It will be needed:
- a filter/templatetag that returns if a field is multilang
- a filter/templatetag that returns the list of languagecode+content for a multilang field

(Rememeber that the template has a complete jsonscheminstance available as `metadata`)

In the official GeoNode iso19115 xml template we may only provide multilang checks for `title` and `abstract`. 
Projects including other multilang fields will provide their own template.

Issue #13643, opened 2025-10-21
Multilang: client metadata editor i18n


etj:
We need a way to pack multilang entries in a visual control that makes it clear that the various text fields belong to a single context.

On client side, we may take example from how GeoNetwork deals with multilang translations.

- In this case, "English" is selected:
   <img width="448" height="90" alt="Image" src="https://github.com/user-attachments/assets/9ff57197-231a-4f7c-ad5d-8d0ae7b4f303" />
- Clicking on other languages, the related translation is shown:
   <img width="452" height="92" alt="Image" src="https://github.com/user-attachments/assets/06d66ee6-c981-4152-8951-b900b2c0a642" />
- Clicking on "all", all the translations are presented at once:
  <img width="450" height="270" alt="Image" src="https://github.com/user-attachments/assets/7fb6d50f-d30f-4862-9b08-42d309c95781" />
- By selecting "One" you get back to the single language textfield.

As reported in #13642 , this is the jsonschema related to a multilang field:

```json
    "title": {
      "type": "string",
      "title": "Title",
      "description": "name by which the cited resource is known",
      "maxLength": 255,
      "geonode:handler": "base",
      "geonode:required": true,
      "geonode:multilang": true,

      "ui:widget": "hidden"
    },
    "title_multilang_en": {
      "type": "string",
      "title": "Title [English]",
      "description": "name by which the cited resource is known",
      "maxLength": 255,
      "geonode:handler": "sparse",

      "geonode:multilang-lang": "en",
      "geonode:multilang-group": "title",
    },
    "title_multilang_it": {
      "type": "string",
      "title": "Titolo [Italiano]",
      "description": "nome attraverso cui la risorsa citata è conosciuta",
      "maxLength": 255,
      "geonode:handler": "sparse",

      "geonode:multilang-lang": "it",
      "geonode:multilang-group": "title",
    },
```

- `"geonode:multilang": true` in `title` tells that this is the main entry, to which the other entries will reference. 
   - This is needed server-side.
- `"geonode:multilang-group":"title"` tells to which main entry the subschema is referring to, useful for grouping translation subschemas. Having this explicit info, the client does not need to make assumptions about which fields belong to the translation group using the fieldname or whatever.
   - this is just a proposal
   - **is it useful for the client?**
- "geonode:multilang-lang": "it" tells the client to which language this entry refers. 
   - this is just a proposal
   - **is it useful for the client?**

See also:
- #13642

giohappy:
[...]

PR #13722, opened 2025-11-12
PR for all GNIP-104 sub-issues

### LINKED ISSUE #2266: Multilang: use indexed fields in client queries
According to https://github.com/GeoNode/geonode/issues/13644, the backend now supports text search based on database text indexes.

The client at the moment uses as query params `search_field`+`search` to search some text in the given fields.
The client should now use one search index using the params `search_index`+`search`, where the default indexes are `title`, `title_abstract`, `all`, defined in this way:
```py
METADATA_INDEXES = {
    "title": ["title"],
    "title_abstract": ["title", "abstract"],
    "all": ["title", "abstract", "supplemental_information"],
}
```
Note that the old search `search_field`+`search` will still be supported.

### LINKED ISSUE #13641: Multilang: index
We need an `index` sub-app to handle resource indexing.
This is a general improvement to geonode, but becomes required since we are introducing multilang metadata, so some logic that understand how to retrieve and index the various metadata fields is needed.

We are going to leverage the [full text search](https://www.postgresql.org/docs/current/datatype-textsearch.html) in Postgres and the [related django library](https://docs.djangoproject.com/en/6.0/ref/contrib/postgres/search/).

### Model

The basic model class for indexing is the `ResourceIndex`, having fields:
- `resource`: FK to the ResourceBase
- `lang`: nullable 2-letters language code
- `name` the index name 
- `vector` the content's tsvector, a `SearchVectorField`

The tuple (`resource`, `lang`, `name`) is unique.

The language code must be nullable, since this indexing will be used also in GeoNode instances where multilang is not enabled.

We may [want to choose](https://www.postgresql.org/docs/9.1/textsearch-indexes.html#:~:text=GIN%20indexes%20take%20about%20three,times%20larger%20than%20GiST%20indexes) [GIN](https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/indexes/#ginindex) over GiST index, since reads will be more frequent than writes.
This may need a new postgres docker image with the GIN extension available.
We may add the extension via a django migration, according to the [GIN documentation](https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/operations/#btreeginextension)

### Logic

The main logic manager will be the `IndexManager`.

The method `update_index()` will take as input:
- `resource`: the ResourceBase to be indexed
- `jsonschema`: the related jsonschema. The logic will take from here the info about multilang fields
- `jsoninstance`: the metadata used to create the indexed data

### Settings

The used settings are:
- `METADATA_MULTILANG`, array containing the fields with multilang data
- `METADATA_INDEXES`, defines which metadata fields will be composing the indexed text
  It is a dict defined this way:
  - key: the name of the index field
  - value: a list of schema fields composing the index

Example:
```JSON
{
   "title" : ["title"],
   "title_abstract": ["title", "abstract"],
   "all":["title", "abstract", "supplemental_info"]
}
```
The combined fields allows better search performance than having each field vectorized.

#### No Multilang

If `METADATA_MULTILANG` is empty, the index record language will always be null,  so that :

- For `title` index:
  - ```
       lang = null
       name = title
       vector = tsvector of jsonschema["title")]
     ```
- For `title_abstract` index:
  - ```
       lang = null
       name = title_abstract
       vector = tsvectorof jsonschema["title"] + jsonschema["abstract"]
    ```
- For `all` index:
  - ```
       lang = null
       name = all
       vector  = tsvectorof jsonschema["title"] + jsonschema["abstract"] + jsonschema["supp_info"]
     ```

#### Multilang enabled

For each entry in the METADATA_INDEXES dict and for each language, an index record will be created with the content of related fields in the localized language, or the plain text content if it's not localized.

Some examples:
```json
   "title" : ["title"],
```

if title is not localized, an index record will be created with
- ```
       lang = null
       name= title
       vector = tsvector of jsonschema["title")
   ```

If title is localized, languages are 'en' and 'it', there will be 2 entries:
- Entry for 'en':
  - ```
       lang = 'en'
       name= title
       vector = tsvector of jsonschema["title_multilang_en")
    ```
- Entry for 'it':
  - ```
       lang = 'it'
       name= title
       vector = tsvector of jsonschema["title_multilang_it")
    ```

In the case of multiple values:
```json
   "title_abstract": ["title", "abstract"],
```

let's say `title` is multilang and `abstract` is not (the logic may check it using the passed jsonschema):

- Entry for 'en':
  - ```
       lang = 'en'
       name= title_abstract
       vector = tsvector of js["title_multilang_en"] +  js["abstract"] 
    ```
- Entry for 'it':
  - ```
       lang = 'it'
       name= title_abstract
       vector = tsvector of js["title_multilang_it"] +  js["abstract"] 
    ```

```json
    "all":["title", "abstract", "supplemental_info"]
```

let's say `title` and `abstract` are multilang, and `supplemental_info` is not:
- Entry for 'en':
  - ```
       lang = 'en'
       name= all
       vector = tsvector of js["title_multilang_en"] + js["abstract_multilang_en"]  + js["supplemental_info"] 
    ```
- Entry for 'it':
  - ```
       lang = 'it'
       name= all
       vector = tsvector of js["title_multilang_it"] + js["abstract_multilang_it"]  + js["supplemental_info"] 
    ```

#### Missing translations

If a Resource is missing the translation for a searchable field, such Resource may not be findable by a textual search.
We may want to make it foundable by any content in the other languages.
For instance:

- title_en = text1 --> index en = text1
- title_it = text2 --> index it = text2
- title_es *missing* --> index_es = text1 + text2

That is:

- Entry for 'en':
  - ```
       lang = 'en'
       name = title_abstract
       vector = tsvector  of js["title_multilang_en"] +  js["abstract"] 
    ```
- Entry for 'it':
  - ```
       lang = 'it'
       name= title_abstract
       vector = tsvector of js["title_multilang_it"] +  js["abstract"] 
    ```
- Entry for 'es':
  - ```
       lang = 'es'
       name= title_abstract
       vector = tsvector of js["title_multilang_en"] + js["title_multilang_it"] + js["abstract"] 
    ```

### Calling the indexer

Since the only point where the metadata can be edited is the metadata manager, the end of the update instance is the place where the indexing should be called.

### Initial data -- Title

The title is the only potential multilang field that needs an initialization when a Resource is created, so we may only take care of its special initialization case.

When creating a title for a brand new resource, GeoNode usually uses the file name of the uploaded resource or something derived from it.
If we do not create indexes for it, the resource will be only visibile when no text filters are present. It may be difficult to find it if the metadata editing is a process done by someone else wrt who loaded the resource, so we want to make it searchable in every language configured in the system.
We may create index entries by putting the default title in every language index, so it will be possibile to find the resource in any language.

As reported above, the metadatamanager is generally the only caller of the indexing procedures.
This means that most of its metadta will be unedited yet, and the title can not be multilang (except cases when an XML document is uploaded along with the data).
Since we need to create a record index related to the ResourceBase to make it searchable by title, we may either:
- force a [partial update](#13540) on title in resource.save(), so that the metadatamanager creates the needed index
- invoke explicitly, only on resource creation, the resource indexing. I'd rather avoid this solution in that it may spread the usage of this call also elsewhere. Also, callers of resourcemanager.create() always also calls resourcemanager.update(), so we may want to have a single caller point.

### Migrations

There should be a migration that creates the indexes for all the existing ResourceBase.
We should make sure the [modeltranslation](https://pypi.org/project/modeltranslation/) is not really working; in the unlikely case it is, we should take the strings from the localized fields.

### Commands

It would be useful to have a management command that (re)creates indexes.
It may take optional id/uuid to only limit the operation on a subset of resources.

### LINKED ISSUE #13642: Multilang: metadata manager i18n
In the metadata json schema any text field can be defined as multilang.

It means that it will be possibile, for such field:
- being edited in multiple languages
- being queried in the requested language
- being returned in the requested language

Since also base field can be made multilang, we do not want the schema to be heavily altered, for instance by transforming that field from `text` to `object', since this change would need a migration of all the existing data.
Also, the field may be a wellknown field (e.g. "title") which fieldname/content should not be tampered with.

A field can be set as a multilang field by two means:
- Adding the field name in the setting `MULTILANG_FIELDS`.
   This method can be useful for base fields, so that no change in the scheleton schema is needed
- Adding the annotation `'geonode-multilang':true` in the field subschema.
  Use this method in sparse fields in custom schemas, so that the whole definition will be in the scheleton schema.

Making a text field multilanguage will add in the schema new subschemas, one for each language, for each multilang field.

As an example, set the mutlilang for the `title` field, wither by adding the `'geonode-multilang':true` annotation the `title` field of our scheleton json schema or defining it in the setting.
The GeoNode schema processor (in the metadata manager) will:
- set the "read-only" and "hidden" annotation in the `title` subschema, because we don't want to use it for editing, but anyway we want to publish the title in the default language for that field
- add a `title-imultilang-LANGCODE` schema entry for each configured language.
- the new i18n fields will be sparse fields. We can keep the langcode in a non-normalized form because it will not be used directly, since search will be performed on other tables (see the "Search" paragraph).

The client should be aware of the `geonode-multilang` annotation, since it will gather all the `title-i18n-*` properties to create a brand new control (see the "Metadata editing" paragraph)

It will be up to the i18n handler to store the value of the default language into the unlocalized field ("title" in this case).

The new `MultiLangHandler` shall implement:
- `def update_schema(self, jsonschema: dict, context: dict, lang=None)`
  - Defined as last Handler, it should loop on all the existing fields, and check if in the subschema `'geonode-multilang':true` is defined. In that case, the "hidden" annotation should be added. 
    *TBD: check if the client needs the related fields defined here, of if can work it out by itself.*
  - Add the related SparseFields. 
    At the moment the SparseHandler needs the Sparse field to be registered beforehand, so that the SparseHandler itself will add the fields. 
    The method `SparseHandler.update_schema()` needs to be refactored by adding a new method 'add_sparse_field()', so that the logic that perform subschema augmentation can be called from the outside.
 - `def post_save(self, resource: ResourceBase, json_instance: dict, context: dict, errors: dict, **kwargs)`
    Should set the base field with the multilang field in the default language. If the field in the default lang is not set, 

Make sure to have the `MultiLangHandler` defined *after* the `SparseHandler`, since the MLH needs to find the sparse fields declared as multilang already in the schema.

TBD: Also note that on first call of the metadata editor, all the localized field will be empty.
Since the 'title` field may contain an initial content, we may choose to copy such value in all the localized items, or maybe just in the default lang.
 
Note that this solution only works with text field in the first level of the schema, and not for fields embedded in subschemas. Since this case includes a broad set of use cases, and the first required multilang field is `title', we'll go this way.

Subschema example:

If we have this skeleton schema:
```json
    "title": {
      "type": "string",
      "title": "Title",
      "description": "name by which the cited resource is known",
      "maxLength": 255,
      "geonode:handler": "base",
      "geonode:required": true,
      "geonode:multilang": true
    },
```

it will be augmented as:

```json
    "title": {
      "type": "string",
      "title": "Title",
      "description": "name by which the cited resource is known",
      "maxLength": 255,
      "geonode:handler": "base",
      "geonode:required": true,
      "geonode:multilang": true,

      "ui:widget": "hidden"
    },
    "title_multilang_en": {
      "type": "string",
      "title": "Title [English]",
      "description": "name by which the cited resource is known",
      "maxLength": 255,
      "geonode:handler": "sparse",

      "geonode:multilang-lang": "en",
      "geonode:multilang-group": "title",
    },
    "title_multilang_it": {
      "type": "string",
      "title": "Titolo [Italiano]",
      "description": "nome attraverso cui la risorsa citata è conosciuta",
      "maxLength": 255,
      "geonode:handler": "sparse",

      "geonode:multilang-lang": "it",
      "geonode:multilang-group": "title",
    },
```

- `title` and `description` will be added for generic clients that do not know how to group multilang fields.
- TBD `geonode:multilang-lang` could be needed by GeoNode client to display the language name
- TBD `geonode:multilang-group` could be needed by GeoNode client to group the subfields related to the same multilang field.

### LINKED ISSUE #13644: Multilang: update API query using index
According to GeoNode/geonode#13641 , the text filtering should no longer be performed on fields in ResourceBase, but should use a different model, that leverages on postgres full text search.

In class `geonode.base.api.views.ResourceBaseViewSet` the field [`filter_backends`](https://github.com/GeoNode/geonode/blob/4.4.3/geonode/base/api/views.py#L311) includes the [`DynamicSearchFilter`]( https://github.com/GeoNode/geonode/blob/master/geonode/base/api/filters.py#L35-L37) which at the moment deals with the text query.

We may retain such filter in order to make the params `search_field`+`search` still work (so that the client can be migrated later, but also for backward compatibility), and add another filter that enables the search by the index, for instance `search_index`+`search`.
Also note that DynamicSearchFilter will not alter the queryset if `search_field` is not given, so the two search methods can co-exist side by side.

### Settings

The used settings are:
- `METADATA_INDEXES`, defines which metadata fields will be composing the indexed text

The settings are detailed in GeoNode/geonode#13641 .

### Procedure 

Like the DynamicSearchFilter, this filter will only be applied if both the `search_index` and `search` query params are defined.

The keys in the `METADATA_INDEXES` dict are used to check the correctness of the `search_index` single param.

The requested language will be read from (in sequence, the first found wins)
- the query param `lang` (so that we can target a specific language explicitly)
- the `Accept-Language` header (usually handled by the browser automatically)
- the default language in the settings

The search query param will be transformed into a [search vector](https://docs.djangoproject.com/en/6.0/ref/contrib/postgres/search/#searchvector) and searched within the `ResourceIndex`
The filter on `lang` should include both the requested language and `lang=null`; the null language will include indexes in a non-multilang geonode instance and indexes made of non multilingual fields. 
For each resource and indexname, no more that one record is expected (either the index is localized or it is not).

### LINKED ISSUE #13645: Multilang: update API output payload
API should be updated so that calls returning metadata fields uses the multilang data when needed.

As a first step we need to collect all the endpoint returning metadata.

The main endpoint used by the geonode index page is `/api/v2/resources` implemented by `geonode.base.api.views.ResourceBaseViewSet`.

The ViewSet uses the `geonode.base.api.serializers.ResourceBaseSerializer` to serialize the resources.
We need a new `MultiLangSerializer` for the `title` field and for any other text field that can be made multilang.

The Serializer should have the original request in the serializer context, so the requested language can be retrieved from there. 
Please check issue #13644 to find out the ways in which the lang info can be given and their priority.

The serializer should use the jsonschema to find out whether the field is multilang.
In that case, the related localized SparseField should be returned.
If such SparseField is not found, the value of the unlocalized field should be used: if it's a base field, it should be returned as is; if it's SparseField, null can be returned.

### LINKED ISSUE #13649: Multilang: remove modeltranslation
As described in https://github.com/GeoNode/geonode/issues/13640, modeltranslation is no longer supported in GeoNode.

etj:
PR for GeoNode/geonode#13640, including one commit for each sub-issues:

- Multilang: remove modeltranslation GeoNode/geonode#13649
- Multilang: index GeoNode/geonode#13641
- Multilang: update API query using index GeoNode/geonode#13644
- Multilang: metadata handler GeoNode/geonode#13642
- Multilang: update API output payload GeoNode/geonode#13645

Not included:
- to be implemented later:
  - Multilang: use multilang entries in ISO19115 GeoNode/geonode#13646
  - Multilang: parse multilang entries in ISO19115 GeoNode/geonode#13647
- to be discussed
  - Multilang: handle i18n queriable fields in pycsw GeoNode/geonode#13648
- client related:
  - Multilang: client metadata editor i18n GeoNode/geonode#13643
  - Multilang: use indexed fields in client queries GeoNode/geonode-mapstore-client#2266

## Checklist

For all pull requests:


mattiagiupponi (approved):

etj:
Replaced by https://github.com/GeoNode/geonode/pull/13769

PR #13745, opened 2025-11-18
Admin theme alignment

### LINKED ISSUE #2245: Django Admin revamp
The plans for GeoNode 5 include a new central role for the Django Admin section. Some GeoNode functionality currently implemented with legacy views and templates should be moved here, and new planned admin features should be located under the admin area (batch actions, custom views, etc.).

We would like the new admin to also be the section where users can manage their profiles and groups in place of the current views.
This means that access to the admin should have different access levels depending on the user's role (registered member, group manager, admin).

Given the new centrality of this section, we want to restyle and rename it to something more meaningful. Maybe "GeoNode Admin", or "GeoNode Backoffice"?
We are evaluating the [unfold](https://unfoldadmin.com/), which provides a nice-looking UI and pretty useful features for the layout of the admin tools. However, we prioritize the following over pleasantness:
- compatibility with the current GeoNode admin widgets (e.g., TinyMCE), filters, and custom pages.
- maintenance, since we want to avoid introducing a dependency that adds too big a burden to maintain and keep aligned

giohappy:


giohappy:
Moved to the client https://github.com/GeoNode/geonode-mapstore-client/pull/2246

PR #13830, opened 2025-12-17
[Fixes #13828] Return dirty resources inside endpoints

### LINKED ISSUE #13828: Return dirty resources inside endpoints
I suspect this is a regression: resources in a dirty state are not returned by the APIs, which prevents the client from showing and managing them according to their current state. 
In general, hiding resources in a dirty state makes them disappear from any catalogue listing, which is confusing at the very least.

For example, when a resource is cloned, it's marked as dirty to prevent changes to the resource until the clone operation is completed. If the operation takes longer, the user will see the original resources disappear, while the cloned resource is not created yet.
We want instead to return the resource, along with the "executions" info, so that the client can apply the expected labels and constraints, and provide users the expected feedback.

sijandh35:
Fixes #13828 

This PR returns dirty state resources inside endpoints and also added execution info for original dirty state resource.

**Note:**
During cloning process, if we quickly refresh it after cloning it shows the following error and refreshing again after some period it will go, this behaviour seen when checked from admin.
image

It might be due to resource_type null for newly created resource on dirty state, so might need be handled from the client side for this error.

## Checklist

For all pull requests:

### DIFF
--- a/geonode/upload/celery_tasks.py
+++ b/geonode/upload/celery_tasks.py
@@ -602,7 +602,7 @@ def copy_geonode_resource(self, exec_id, actual_step, layer_name, alternate, han
 
         _exec = orchestrator.get_execution_object(exec_id)
 
-        # Update input_params with original resource's PK immediately so the API
+        # Update input_params with original resource's uuid immediately so the API
         # can find this execution request for original the dirty resource
         orchestrator.update_execution_request_status(
             execution_id=str(_exec.exec_id),
--- a/geonode/security/registry.py
+++ b/geonode/security/registry.py
@@ -261,7 +261,7 @@ class PermissionsHandlerRegistry:
         admin_approval_required=False,
         unpublished_not_visible=False,
         private_groups_not_visibile=False,
-        include_dirty=False,
+        include_dirty=True,
     ):
         # Get the list of objects the user has access to
         from geonode.groups.models import GroupProfile


geonode/upload/celery_tasks.py
```diff
@@ -602,6 +602,13 @@ def copy_geonode_resource(self, exec_id, actual_step, layer_name, alternate, han
 
         _exec = orchestrator.get_execution_object(exec_id)
 
+        # Update input_params with original resource's PK immediately so the API
+        # can find this execution request for original the dirty resource
+        orchestrator.update_execution_request_status(
+            execution_id=str(_exec.exec_id),
+            input_params={**_exec.input_params, **{"uuid": resource.uuid}},
```
mattiagiupponi, line 609: In the comment is mentioned the `PK`, but i see that the `UUID` is passed, which one is the expected one?

geonode/security/utils.py
```diff
@@ -56,7 +56,7 @@ def get_visible_resources(
     admin_approval_required=False,
     unpublished_not_visible=False,
     private_groups_not_visibile=False,
-    include_dirty=False,
+    include_dirty=True,
```
mattiagiupponi, line 59: i think also the util should use the True by default [get_visible_resources](https://github.com/GeoNode/geonode/blob/bde9c9dadcb79d97ccc17254a917cca9ec837866/geonode/security/utils.py#L59)

geonode/upload/celery_tasks.py
sijandh35: Updated the comment, [uuid](https://github.com/GeoNode/geonode/blob/bde9c9dadcb79d97ccc17254a917cca9ec837866/geonode/base/api/serializers.py#L506) is expected.

geonode/security/utils.py
sijandh35: Updated PR with making [utils](https://github.com/GeoNode/geonode/blob/bde9c9dadcb79d97ccc17254a917cca9ec837866/geonode/security/utils.py#L51) and [registry](https://github.com/GeoNode/geonode/blob/bde9c9dadcb79d97ccc17254a917cca9ec837866/geonode/security/registry.py#L255) `get_visible_resources` `include_dirty` default to True.

PR #13887, opened 2026-01-16
[Fixes #13885] Support for COG Remote Layers

### LINKED ISSUE #13885: Support for Cloud Optimized GeoTIFF (COG) remote  layers
**Description**

This issue proposes adding support for Cloud Optimized GeoTIFF (COG) datasets as remote layers. Users would be able to register a remote COG by providing its URL, allowing GeoNode to access large raster datasets hosted externally without downloading them locally.

The implementation should leverage the HTTP range request capabilities of the COG format to retrieve essential raster metadata such as CRS and spatial extent directly from the remote source. If the remote server does not support range requests, dataset creation should fail gracefully with a clear, user-friendly message explaining the limitation. When metadata retrieval succeeds, the dataset is registered as a remote raster resource backed by the COG URL, enabling efficient and cloud-native raster access.

sijandh35:
Fixes #13885 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/upload/handlers/remote/cog.py
+++ b/geonode/upload/handlers/remote/cog.py
@@ -104,8 +104,8 @@ class RemoteCOGResourceHandler(BaseRemoteResourceHandler):
         logger.debug(f"Attempting to open COG with GDAL: /vsicurl/{url}")
         try:
             # Set GDAL config options for faster failure
-            gdal.SetConfigOption("GDAL_HTTP_TIMEOUT", "15")
-            gdal.SetConfigOption("GDAL_HTTP_MAX_RETRY", "1")
+            gdal.SetThreadLocalConfigOption("GDAL_HTTP_TIMEOUT", "15")
+            gdal.SetThreadLocalConfigOption("GDAL_HTTP_MAX_RETRY", "1")
 
             vsiurl = f"/vsicurl/{url}"
             ds = gdal.OpenEx(vsiurl)
--- a/geonode/upload/handlers/remote/tests/test_cog.py
+++ b/geonode/upload/handlers/remote/tests/test_cog.py
@@ -68,8 +68,8 @@ class TestRemoteCOGResourceHandler(TestCase):
         with self.assertRaises(ImportException):
             self.handler.is_valid_url(self.valid_url)
 
-    @patch("geonode.upload.handlers.remote.cog.gdal.Open")
-    def test_create_geonode_resource(self, mock_gdal_open):
+    @patch("geonode.upload.handlers.remote.cog.gdal.OpenEx")
+    def test_create_geonode_resource(self, mock_gdal_openex):
         # Mock GDAL dataset
         mock_ds = MagicMock()
         mock_srs = MagicMock()
@@ -79,7 +79,7 @@ class TestRemoteCOGResourceHandler(TestCase):
         mock_ds.GetGeoTransform.return_value = [0, 1, 0, 0, 0, -1]
         mock_ds.RasterXSize = 100
         mock_ds.RasterYSize = 100
-        mock_gdal_open.return_value = mock_ds
+        mock_gdal_openex.return_value = mock_ds
 
         exec_id = orchestrator.create_execution_request(
             user=self.owner,


mattiagiupponi (changes requested):
Hi @sijandh35 
I teste using the images here: https://cogeo.craig.fr/opendata/ortho/

1) the images are not visible, i can just see the basemap (example https://cogeo.craig.fr/opendata/ortho/orthocraig3_vichy_2021.cog.tif)
2) i'm getting the following error for the coordinate, can you doublecheck if we are correctly managing them?

Image

3) The thumbnail is not correctl generated, the BE raise the following error:

```python
[2026-01-21T12:10:49Z] {/opt/core/geonode/geonode/thumbs/utils.py:240} ERROR - Could not find layer httpscogeocraigfropendataortho2018_25cm_ain2biseretif
  File "/opt/core/geonode/geonode/thumbs/utils.py", line 218, in get_map
    image = getmap(
            ^^^^^^^
  File "/opt/core/geonode/geonode/thumbs/utils.py", line 435, in getmap
    raise ServiceException(err_message)
owslib.util.ServiceException: Could not find layer httpscogeocraigfropendataortho2018_25cm_ain2biseretif
```

Image

geonode/upload/handlers/remote/cog.py
```diff
@@ -0,0 +1,187 @@
...
+            if ds is None:
+                logger.debug(f"GDAL failed to open dataset: {vsifile}")
+                raise ImportException(f"Could not open remote COG: {url}")
+
+            logger.debug("GDAL opened dataset. Extracting metadata...")
+            # Get Projection/CRS
+            srs = ds.GetSpatialRef()
+            if srs:
+                authority = srs.GetAuthorityName(None)
+                code = srs.GetAuthorityCode(None) or srs.GetAuthorityCode("PROJCS") or srs.GetAuthorityCode("GEOGCS")
```
mattiagiupponi: there is no way to re-use the `identify_authority` defined as in the other handlers? so we can keep choerent they way to indentify the authorities

mattiagiupponi (changes requested):

giohappy:
@mattiagiupponi my feedback on your points:

1. The https://cogeo.craig.fr/opendata/ortho/orthocraig3_vichy_2021.cog.tif has EPSG:2254, which is not a supported EPSG in the client (unless you configure it). So the issue is not in the backend, which correctly parses the CRS and bbox. You can verify the issue by adding the COG directly to a map, using the MapStore catalog
2. We will not generate thumbnails for COGs, since they're remote files that we don't want to download in the backend, and in any case we would need to implement a method to render them. We could maybe decide to use the Overviews that can optionally be contained inside the COG, but it's left for a future enhancement in case
3. Regarding the EPSG identification, I agree that we should adopt a common approach, but I have two questions:
  - We have duplicated methods [here](https://github.com/GeoNode/geonode/blob/streaming_download_handler/geonode/upload/handlers/common/vector.py#L422) and [here](https://github.com/GeoNode/geonode/blob/streaming_download_handler/geonode/upload/handlers/common/raster.py#L263). We should probably extract the method to a common utils module first
  - Why do you first try the hard way, i.e. parse the `wkt` or the `proj4` definitions first, and then try to obtain the code directly if they fail? It looks like the logic should be reversed, but I suppose there's a good reason for this order.

giohappy:
@sijandh35 isn't there a better way to avoid the creation of the thumbnails for COGs? The condition [here](https://github.com/GeoNode/geonode/pull/13887/changes/58eb93f47b0578b9fc5f96cf63c18f86ef9d7bae..e474aa8b29216c53ed27960831b09b8691dd1c3d ) is spaghetti code. What if we add other formats that shouldn't produce thumbnails? We can't keep adding hardcoded conditions like that.
How do we do for 3dtiles?
What solution do you propose to encapsulate this better, and avoid spilling out the conditions inside the thumbnail module?

geonode/upload/handlers/remote/cog.py
mattiagiupponi: @sijandh35 try to see if is possible to authority identification into the basehandler and let this handler uses the code in common with the others.

geonode/upload/handlers/remote/cog.py
```diff
@@ -0,0 +1,171 @@
...
+            logger.exception(e)
+            if isinstance(e, ImportException):
+                raise e
+            raise ImportException(f"Failed to extract metadata from COG: {url}")
+        resource = super().create_geonode_resource(layer_name, alternate, execution_id, resource_type, asset)
+        resource.set_bbox_polygon(bbox, srid)
+        return resource
+
+    def generate_resource_payload(self, layer_name, alternate, asset, _exec, workspace, **kwargs):
+        payload = super().generate_resource_payload(layer_name, alternate, asset, _exec, workspace, **kwargs)
```
mattiagiupponi, line 162: Hi @sijandh35 
This override is not strictly required:
1) The workspace for remote resources is not required since are hosted by an external service and not by the internal geoserver for geonode
2) the name (even if not defined is saved as alternate) and alternate are already defined in the `BaseRemoteResourceHandler`,

so i guess this part can be removed

geonode/upload/handlers/remote/cog.py
```diff
@@ -0,0 +1,171 @@
...
+        # Extract metadata via GDAL VSICURL
+        gdal.UseExceptions()
+        logger.debug(f"Attempting to open COG with GDAL: /vsicurl/{url}")
+        try:
+            # Set GDAL config options for faster failure
+            gdal.SetConfigOption("GDAL_HTTP_TIMEOUT", "15")
+            gdal.SetConfigOption("GDAL_HTTP_MAX_RETRY", "1")
+
+            vsifile = f"/vsicurl/{url}"
+            ds = gdal.Open(vsifile)
```
mattiagiupponi: In order to speed-up a bit the opening of the URL, i suggest to use `OpenEx` instead of `Open`, not because the `Open` is slow, but with `OpenEx` is possible to specify the driver that gdal needs to use to open the resource, in this case the driver will be COG, plus also allow to specify the attribute directly in the open call instead of setting at `ConfigOpen` level:

for example:

```
 ds = gdal.OpenEx(
    f"/vsicurl/{url}",
    allowed_drivers=['COG'],
    open_options=['HTTP_TIMEOUT=15', 'HTTP_MAX_RETRY=1']
)
```

PR #13916, opened 2026-01-30
[Fixes 13919] Fix UTF-8 handling for GeoJSON uploads with non-ASCII properties

F0rtiter:
Ensure GeoJSON files containing non-ASCII (e.g. Chinese) attribute values
are correctly detected and validated by explicitly decoding input as UTF-8
in GeoJsonFileHandler.

This prevents false negatives in handler detection and avoids incorrectly
reporting valid GeoJSON files as invalid due to implicit or system-dependent
encoding assumptions.

### DIFF
--- a/geonode/upload/handlers/geojson/handler.py
+++ b/geonode/upload/handlers/geojson/handler.py
@@ -73,7 +73,7 @@ class GeoJsonFileHandler(BaseVectorFileHandler):
             try:
                 _file = base
                 if isinstance(base, str):
-                    with open(base, "r") as f:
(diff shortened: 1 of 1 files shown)


mattiagiupponi:
Hi @F0rtiter thanks for the contribution, as soon as the Test suite pass i'll test the PR.
If possible, can you also open an issue explaining the use case? is more to keep track and a connection between ISSUE the PR assosiated, so when we prepare the release notes we can link the issue to them.

Thanks!

F0rtiter:
I've opened an issue describing the use case: #13919

PR #13917, opened 2026-01-30
[Fixes #13913] Disable default sync dataset download

### LINKED ISSUE #13913: Remove WPS sync downloads backend side
### Background and Overview of dataset downloads

The download of datasets, which is triggered by the Download button and the download icon, is driven by the object returned inside the `download_urls` property, where the `default` property is `True`. This was done to distinguish dataset downloads from assets downloads, which are also returned inside the `download_urls` property.

The URL returned inside the object depends on the configured download handler. The default one returns the /dataset_download endpoint or, if a `Link` of type "original is connected to the resource, the URL set inside the `Link` object. The concept of a Link comes from previous versions and is still supported to allow administrators to configure custom download URLs without having to implement a custom download handler.

The `/dataset_download` endpoint, when using the default handler, will make a **sync** call to Geoserver WPS Download and will stream the response from it, basically acting as a proxy to the WPS service.
This can be ok for small to medium datasets, but is problematic for bigger ones, for a list of reasons:

1. The current implementation doesn't actually stream the response. The response from WPS is loaded in memory and returned by Django
2. Even if we switch to a stream response, we found that correctly configuring the buffering and back-pressure of the whole chain (from Geosever to Nginx) is tricky to achieve, since its optimisation also depends on factors related to the specific deployment (networking, etc.).
3. The WPS sync download is also heavier on the side of Geoserver, since requests cannot be queued and optimized

We want to refactor the logic for the default downloads based on WPS to reuse the async WPS functionality that the client already uses for the "Export data"  tool.

### Plan
1. "default" download objects won't be returned anymore by the default download handler for datasets. Custom download handlers can still return them.
2. When the client application doesn't find a "default" download inside the `download_urls` array, it will:
   - Disable the Download button, since it would be a duplication of the Export data button
   - Trigger the export data functionality when the download icon is clicked inside the dataset info bar
3. In case the "default" download URL is returned, the client will behave the same as now, including the `ajax_safe` option 

Nothing will change for users, except that the Download button will be hidden for all datasets, unless a `Link` (type `original`) is configured for the resource or a custom download handler returns the default download object.

### Implementation
The download handler will be adapted to return nothing for the default download URL, but keep the current logic for the other cases.

nrjadkry:
Fixes #13913 

## Changes
- **`geonode/layers/download_handler.py`**: 
  - `DatasetDownloadHandler.download_url` now returns `None` instead of the dynamic `dataset_download` URL.
  - It only returns a URL if a link (`link_type="original"`) is explicitly configured.

- **`geonode/base/api/serializers.py`**: 
  - Updated `DownloadArrayLinkField` logic to prevent the asset from  receiving `default=True` when the download list is empty.
 
## Checklist

For all pull requests:

### DIFF
--- a/geonode/layers/download_handler.py
+++ b/geonode/layers/download_handler.py
@@ -89,4 +89,3 @@ class DatasetDownloadHandler:
                 logger.debug(e)
 
         return self._resource
-


mattiagiupponi:
Hi @nrjadkry can you give a check to the failing test?

nrjadkry:
> @nrjadkry tests are failing and we can drop the the dataset_download method inside the default hanlder. We're not using it anymore...

@giohappy just to clarify, do you mean removing the `process_dowload` (and `get_download_response` because it needs `process_download`) function inside the handler?

If yes, then should I also remove the `dataset_download` endpoint ? Because this endpoint relies entirely on `get_download_response`. I would also need to update the Dataset model  `download_url` property to stop reversing this URL.

giohappy:
@nrjadkry, my fault, I wasn't clear. We don't want to remove the dataset_download endpoint, since a custom endpoint might want to implement the `get_download_response` method.

We will just return a 404 inside the `get_download_response` for the default download handler, and we can drop the `process_download` method.

nrjadkry:
@giohappy @mattiagiupponi 
I have addressed the failing tests and have updated them according to the new status code returned by `dataset_download` endpoint

PR #13953, opened 2026-02-13
[Fixes #13952] Handle the exception when the update bbox method fails at managing the SRID for a dataset

### LINKED ISSUE #13952: Handle the exception when the update bbox method fails at managing the SRID for a dataset
## Description
When working with datasets that use custom coordinate reference systems (CRS) that cannot be managed by PROJ, the [Dataset.recalc_bbox_on_geoserver](https://github.com/GeoNode/geonode/blob/644f81babe072bcf6565362a3570f7a180bccd91/geonode/layers/models.py#L309) method throws an unhandled exception.

## Current Behavior
The assignment of the bbox polygon and SRID inside [Dataset.recalc_bbox_on_geoserver](https://github.com/GeoNode/geonode/blob/644f81babe072bcf6565362a3570f7a180bccd91/geonode/layers/models.py#L309) does not handle exceptions when the CRS cannot be processed by PROJ, causing the method to fail.

## Proposed Solution
The method should implement try/except exception handling as `geoserver.helpers.sync_instance_with_geoserver `does [here](https://github.com/GeoNode/geonode/blob/644f81babe072bcf6565362a3570f7a180bccd91/geonode/geoserver/helpers.py#L2025-L2040).

Moreover, since the two blocks are very similar, the logic can be moved to a common Dataset method with improved logic so that:
- if the SRID is known and the reprojection can be performed, the `bbox_polygon` geometry and the `srid` will be in the original CRS, and
- otherwise, `srid `is assigned EPSG:4326 and the `bbox_polygon` will be the same as `ll_bbox_polygon`

sijandh35:
Fixes #13952 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/layers/models.py
+++ b/geonode/layers/models.py
@@ -31,7 +31,7 @@ from tinymce.models import HTMLField
 
 from geonode.client.hooks import hookset
 from geonode import GeoNodeException
-from geonode.utils import build_absolute_uri, check_shp_columnnames
+from geonode.utils import build_absolute_uri, check_shp_columnnames, check_bbox_validity
 from geonode.security.models import PermissionLevelMixin
 from geonode.groups.conf import settings as groups_settings
 from geonode.security.permissions import (
@@ -356,12 +356,12 @@ class Dataset(ResourceBase):
         return True
 
     def set_bbox_and_srid(self, bbox, ll_bbox, srid):
-        if not ll_bbox or len(ll_bbox) < 4:
-            raise GeoNodeException("Lat/Lon BBox was not provided")
+        if not check_bbox_validity(ll_bbox):
+            raise GeoNodeException("Lat/Lon BBox was not provided or is invalid")
 
         ll_bbox_gn = [ll_bbox[0], ll_bbox[2], ll_bbox[1], ll_bbox[3]]
         # Try to use native bbox if available
-        if bbox and len(bbox) >= 4 and srid:
+        if check_bbox_validity(bbox) and srid:
             try:
                 native_bbox_gn = [bbox[0], bbox[2], bbox[1], bbox[3]]
                 self.set_bbox_polygon(native_bbox_gn, srid)
--- a/geonode/layers/models.py
+++ b/geonode/layers/models.py
@@ -30,6 +30,7 @@ from django.utils.translation import gettext_lazy as _
 from tinymce.models import HTMLField
 
 from geonode.client.hooks import hookset
+from geonode import GeoNodeException
 from geonode.utils import build_absolute_uri, check_shp_columnnames
 from geonode.security.models import PermissionLevelMixin
 from geonode.groups.conf import settings as groups_settings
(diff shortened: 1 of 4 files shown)


geonode/layers/models.py
```diff
@@ -348,15 +349,33 @@ def recalc_bbox_on_geoserver(self, force_bbox=None):
...
         with transaction.atomic():
-            self.set_bbox_polygon([bbox[0], bbox[2], bbox[1], bbox[3]], srid)
-            self.set_ll_bbox_polygon([ll[0], ll[2], ll[1], ll[3]])
-            self.srid = srid or self.srid
+            self.set_bbox_and_srid_from_geoserver(bbox=bbox, ll_bbox=ll, srid=srid)
             self.save(update_fields=["srid"])
 
         return True
 
+    def set_bbox_and_srid_from_geoserver(self, bbox, ll_bbox, srid):
```
giohappy: @sijandh35 remove the "from_geoserver" suffix. It's a generic method, not strictly related to the Geoserver context.

giohappy (changes requested):

geonode/layers/models.py
```diff
@@ -348,15 +349,33 @@ def recalc_bbox_on_geoserver(self, force_bbox=None):
...
-            self.set_bbox_polygon([bbox[0], bbox[2], bbox[1], bbox[3]], srid)
-            self.set_ll_bbox_polygon([ll[0], ll[2], ll[1], ll[3]])
-            self.srid = srid or self.srid
+            self.set_bbox_and_srid(bbox=bbox, ll_bbox=ll, srid=srid)
             self.save(update_fields=["srid"])
 
         return True
 
+    def set_bbox_and_srid(self, bbox, ll_bbox, srid):
+        if not ll_bbox or len(ll_bbox) < 4:
```
mattiagiupponi: @sijandh35 i propose to impove a bit the check here, if the function is called with ll_bbo='abcd' the check is considered valid, while it should not. i would verify that is at least a list with some numeric values, something like:
```
def check_bbox_validity(value):
    if not value or not isinstance(value, list) or not all(isinstance(x, (int, float)) for x in value):
          return False
    return True


if not check_bbox_validity(ll_bbox):
    raise ...
```

geonode/layers/models.py
```diff
@@ -348,15 +349,33 @@ def recalc_bbox_on_geoserver(self, force_bbox=None):
...
+            raise GeoNodeException("Lat/Lon BBox was not provided")
+
+        ll_bbox_gn = [ll_bbox[0], ll_bbox[2], ll_bbox[1], ll_bbox[3]]
+        # Try to use native bbox if available
+        if bbox and len(bbox) >= 4 and srid:
```
mattiagiupponi: same as above

mattiagiupponi (changes requested):

mattiagiupponi (approved):

giohappy (approved):
@sijandh35 I think we can remove duplications further.

You designed the set_bbox_and_srid method to receive bbox and ll_bbox.
Why don't just receive bbox and perform

PR #13967, opened 2026-02-17
[Fixes #13966] Harvesters: Integrity error when a title is updated in the remote WMS service 

### LINKED ISSUE #13966: Harvesters: Discovery sessions fail because of a "duplicate key" database error
During discovery sessions of a harvester, if a resource's title has been changed in the WMS remote service, the next time that the harvesting session will try to discover the corresponding resource will fail with a "duplcate error" from the database since there is a metadata mismatch between the already registered harvestable resources (in the database) and the corresponding resources that are retrived from the WMS service. 

The error happens because the current logic searches for resources using both the unique ID and the title: https://github.com/GeoNode/geonode/blob/master/geonode/harvesting/tasks.py#L744-L753 
If the remote WMS service updates a resource's title, the system fails to find the existing record and tries to create a new one. This is blocked by the database because the unique ID already exists. 

**Proposed solution**:
We will update the logic to look up resources by unique ID only. Any changes to titles will be handled as an update to the existing record rather than a new insertion. 
Moreover, we will also add a "retry" mechanism to handle potentially simultaneous worker collisions gracefully.

Gpetrak:
This PR is created according to this issue #13966 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/harvesting/tasks.py
+++ b/geonode/harvesting/tasks.py
@@ -757,16 +757,15 @@ def _update_harvestable_resources_batch(self, refresh_session_id: int, page: int
                     # RACE CONDITION: Another worker created this between our SELECT and INSERT.
                     # We catch the error and simply fetch the one they created.
                     resource = models.HarvestableResource.objects.get(
-                        harvester=harvester,
-                        unique_identifier=remote_resource.unique_identifier
+                        harvester=harvester, unique_identifier=remote_resource.unique_identifier
                     )
                     created = False
-                # If the resource wasn't just created, check if the title changed 
+                # If the resource wasn't just created, check if the title changed
                 # (e.g. from 'copy title' to '28409') and update it.
                 if not created:
                     resource.title = remote_resource.title
                     resource.remote_resource_type = remote_resource.resource_type
-                
+
                 processed += 1
                 # NOTE: make sure to save the resource because we need to have its
                 # `last_updated` property be refreshed - this is done in order to be able
--- a/geonode/harvesting/tasks.py
+++ b/geonode/harvesting/tasks.py
@@ -742,17 +742,16 @@ def _update_harvestable_resources_batch(self, refresh_session_id: int, page: int
             processed = 0
             for remote_resource in found_resources:
                 try:
-                    with transaction.atomic():
-                        resource, created = models.HarvestableResource.objects.get_or_create(
-                            harvester=harvester,
-                            unique_identifier=remote_resource.unique_identifier,
-                            defaults={
-                                "title": remote_resource.title,
-                                "should_be_harvested": harvester.harvest_new_resources_by_default,
-                                "remote_resource_type": remote_resource.resource_type,
-                                "last_refreshed": timezone.now(),
-                            },
-                        )
+                    resource, created = models.HarvestableResource.objects.get_or_create(
+                        harvester=harvester,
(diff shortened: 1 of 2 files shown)


geonode/harvesting/tasks.py
```diff
@@ -741,16 +741,31 @@ def _update_harvestable_resources_batch(self, refresh_session_id: int, page: int
...
-                    unique_identifier=remote_resource.unique_identifier,
-                    title=remote_resource.title,
-                    defaults={
-                        "should_be_harvested": harvester.harvest_new_resources_by_default,
-                        "remote_resource_type": remote_resource.resource_type,
-                        "last_refreshed": timezone.now(),
-                    },
-                )
+                try:
+                    with transaction.atomic():
```
giohappy: why so we need transaction.atomic here @Gpetrak ?

Gpetrak: @giohappy I added the atomic block to handle concurrency safety for our 10 parallel harvesting workers (as an extra safety level). While the primary bug was caused by a metadata mismatch, this addition prevents the discovery session from crashing because of the multiple workers.

mattiagiupponi: In this case it might be redundant the atomic block as specified in the django doc:

Image
https://docs.djangoproject.com/en/5.2/ref/models/querysets/#get-or-create

In this case the unique of the two fields is present, so it should be automatically used as an atomic transaction

Gpetrak: @mattiagiupponi you are right and thank you that mention it, but I don't think that it's 100% reduntant. The problem is not the Django itself but the whole database transaction inside the task execution  (Celery) workflow. In a `UniqueViolation` error, the database aborts the current transaction. While Django's `get_or_create` handles the logic, it doesn't clear the 'Aborted' state from the database connection. By wrapping it in a manual atomic block, we create a savepoint. This allows us to roll back only the collision, clean the connection, and safely call `.save()` and continue the rest of the loop without crashing the whole Celery task. I think the problem here is not Django but Celery / PostgreSQL.
Anyway, if we want to prioritize code simplicity, I’m fine with removing it for now. We can monitor the logs for Transaction-aborted errors and re-add the savepoint if the high-concurrency proves it's necessary.

giohappy: I see you already removed the transaction.atomic() context in your last commit, but for clarity I want to highlight that the transaction is already handled inside the [get_or_create](https://github.com/django/django/blob/stable/5.2.x/django/db/models/query.py#L953-L955) method. Wrapping the method inside another atomic transaction would be useless.

```python
def get_or_create(self, defaults=None, **kwargs):
        """
        Look up an object with the given kwargs, creating one if necessary.
        Return a tuple of (object, created), where created is a boolean
        specifying whether an object was created.
        """
        # The get() needs to be targeted at the write database in order
        # to avoid potential transaction consistency problems.
        self._for_write = True
        try:
            return self.get(**kwargs), False
        except self.model.DoesNotExist:
            params = self._extract_model_params(defaults, **kwargs)
            # Try to create an object using passed params.
            try:
                with transaction.atomic(using=self.db):
                    params = dict(resolve_callables(params))
                    return self.create(**params), True
            except IntegrityError:
                try:
                    return self.get(**kwargs), False
                except self.model.DoesNotExist:
                    pass
                raise
```

mattiagiupponi (approved):

PR #13985, opened 2026-02-23
[Fixes #13971] Added optional query parameter to return all translated fields in resource APIs

### LINKED ISSUE #13971: Add optional query parameter to return all translated fields in resource APIs
If a metadata field is configured to support translations, we want all translated fields to be optionally returned in the API response for any resource type, such as datasets/, documents/, or maps/.

A new optional query parameter, include_i18n, will be introduced. When set to true, the response should include all available translations for translatable fields, for example:

```
title: (translated title as we have now),
title_en: (engligh title),
title_ru: (russian),
(...)
```

This feature can be implemented via the [MultiLangViewMixin](https://github.com/GeoNode/geonode/blob/ec146e5bb7cd2e4a9d93a02637ceea08634f69c6/geonode/metadata/multilang/views.py#L12) which already handles localized field resolution

**Note**: Default behavior must remain unchanged for backward compatibility, and translations should only be included when the include_i18n parameter is explicitly requested.

nrjadkry:
Fixes #13971 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/documents/api/tests.py
+++ b/geonode/documents/api/tests.py
@@ -32,6 +32,7 @@ from geonode.base.populate_test_data import create_models
 from geonode.base.enumerations import SOURCE_TYPE_REMOTE
 from geonode.documents.models import Document
 from geonode.metadata.models import SparseField
+from geonode.security.registry import permissions_registry
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


geonode/metadata/multilang/serializers.py
```diff
@@ -12,12 +14,21 @@ class MultiLangOutputMixin(serializers.BaseSerializer):
     def to_representation(self, instance):
 
         representation = super().to_representation(instance)
+        request = self.context.get("request")
+        params = getattr(request, "query_params", None) or getattr(request, "GET", {})
+        include_i18n = params.get("include_i18n", "false").lower() == "true" if request else False
```
giohappy, line 19: I would ignore the value; we only need to check if the query parameter is set. A  `params.get("include_i18n")` should suffice.

geonode/metadata/multilang/utils.py
```diff
@@ -48,3 +48,14 @@ def get_language(request):
...
+
+def get_all_multilang_fields():
+    all_fields = []
+    langs = get_2letters_languages()
+
+    for field in settings.MULTILANG_FIELDS:
+        for lang in langs:
+            all_fields.append(get_multilang_field_name(field, lang))
+
+    return all_fields
```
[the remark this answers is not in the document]
etj: You can implement `get_all_multilang_fields` as suggested by gemini.
Furthermore, it would be good to return a dict (field,lang):fieldname.
With the tuple you can easily add more lang info in the response, without need to parse the field name.

geonode/metadata/multilang/serializers.py
```diff
@@ -12,12 +14,21 @@ class MultiLangOutputMixin(serializers.BaseSerializer):
...
 
         if settings.MULTILANG_FIELDS and hasattr(instance, "_multilang_sparse_prefetch"):
             for sparse in instance._multilang_sparse_prefetch:
```
etj: using `get_all_multilang_fields`, you may loop on the returned fields instead of `_multilang_sparse_prefetch`

geonode/metadata/multilang/serializers.py
giohappy: @nrjadkry after speaking with @etj he thinks your implementation is more intuitive.
So, ok, forget my comment and leave it as it is

geonode/metadata/multilang/serializers.py
nrjadkry: @etj 
I’ve applied the suggested changes to `get_all_multilang_fields` (returning a dict) and updated the serializer as well

PR #14003, opened 2026-02-27
[Fixes #13996] Permissions cache reset when changing from Readonly to normal

### LINKED ISSUE #13996: Permissions cache is not reset when changing from Readonly to normal
**Use Case:**

- admin set the system in read-only
- the env variable PERMISSION_CACHE_EXPIRATION_TIME is not set by default in the project
- admin navigates in the system, the permissions_registry cached the permissions to read-only
- the admin removed the read-only
- the admin still have the permissions cached earlier since the cache expiry time was not reached
- a restart is required

**Expected:**

the cache should be cleared if the system is set in read-only and re-cleared when the system is set to normal.
Plus the variable was missing from the geonode-project structure. we just added it via https://github.com/GeoNode/geonode-project/pull/602

this should be backported to 5.0.x

sijandh35:
Fixes #13996 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/geoserver/helpers.py
+++ b/geonode/geoserver/helpers.py
@@ -75,6 +75,7 @@ from geonode.utils import (
     get_legend_url,
     is_monochromatic_image,
     set_resource_default_links,
+    normalize_bbox_to_float_list,
 )
 
 from .geofence import GeoFenceClient, GeoFenceUtils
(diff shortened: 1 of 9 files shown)


geonode/security/registry.py
```diff
@@ -462,6 +462,32 @@ def delete_resource_permissions_cache(self, instance, user_clear_cache=True, gro
         else:
             pass
 
+    def clear_permissions_cache(self):
+        """
+        Clear all known permission cache entries.
```
mattiagiupponi: Hi @sijandh35 
would not be esier to do like this:

```python
from django.core.cache import cache

# This wipes everything in the default cache
cache.clear()
```

I know all the cache is cleared, but is a rarely used operation. Plus we might have a specific cache-named for the permissions like we have for other parts of GeoNode in the future

mattiagiupponi (changes requested):

geonode/security/registry.py
sijandh35: Updated.

PR #14096, opened 2026-04-03
[Fixes #14095] Upload thumbnails through the UI for the remote spatial datasets

### LINKED ISSUE #14095: Users are not able to upload thumbnails through the UI for the remote spatial datasets
# Description
Users are not able to upload thumbnails through the UI for the remote spatial datasets which include the following formats: 3dtiles, "cog", "flatgeobuf". 

This is because of the method `can_have_thumbnail` where is used in the `resource_manager` here: https://github.com/GeoNode/geonode/blob/master/geonode/resource/manager.py#L959
while the method is this one: https://github.com/GeoNode/geonode/blob/master/geonode/base/models.py#L972-L974

As a result, in the current workflow users are not able to have thumbnails in the remote spatial datasets. 

### Proposed solution
The thumbnail should not be generated during upload (as it is already applied), but the users should instead have the ability to apply or modify the thumbnail afterward through  the UI.

sijandh35:
Fixes #14095 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/resource/manager.py
+++ b/geonode/resource/manager.py
@@ -959,7 +959,7 @@ class BaseResourceManager(ResourceManagerInterface):
         map_thumb_from_bbox: bool = False,
     ) -> bool:
         _resource = instance or BaseResourceManager._get_instance(uuid)
-        if _resource and (thumbnail or _resource.can_have_thumbnail):
+        if _resource and _resource.can_set_thumbnail(thumbnail):
             try:
                 with transaction.atomic():
                     if thumbnail:
--- a/geonode/base/models.py
+++ b/geonode/base/models.py
@@ -969,8 +969,16 @@ class ResourceBase(PolymorphicModel, PermissionLevelMixin, ItemBase):
     def can_have_style(self):
         return self.subtype not in {"tileStore", "remote"}
 
-    @property
-    def can_have_thumbnail(self):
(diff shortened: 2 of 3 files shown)


geonode/resource/manager.py
```diff
@@ -959,7 +959,7 @@ def set_thumbnail(
         map_thumb_from_bbox: bool = False,
     ) -> bool:
         _resource = instance or BaseResourceManager._get_instance(uuid)
-        if _resource and _resource.can_have_thumbnail:
+        if _resource and (thumbnail or _resource.can_have_thumbnail):
```
Gpetrak: @sijandh35 although the fix works fine, we could further improve the code in terms of sustainability and readability.
At first I would  rename the property from `can_have_thumbnail` to the `can_set_thumbnail` and I would move the logic there as below:

```py
def can_set_thumbnail(self, thumbnail_data=None):
    """
   a simple description
    """
    if thumbnail_data:
        return True
    
    # a comment for clarification
    return self.subtype not in {"3dtiles", "cog", "flatgeobuf"}
```
 and then in the manager we could simply have: 
```py
if _resource and _resource.can_set_thumbnail(thumbnail):
```
As I saw the property `can_have_thumbnail` is used only in this part.

sijandh35: Updated.

PR #14108, opened 2026-04-06
[Fixes #14107] Add configurable sync strategies for Social Account group memberships

### LINKED ISSUE #14107: Prevent destructive group membership wipes during OIDC synchronization
Currently, GeoNode employs a strict, reconciliation-based approach to synchronizing user groups during the social login flow (OIDC/Azure). While this ensures a 1:1 mirror with the Identity Provider (IdP), it lacks the flexibility required for environments where group memberships may be managed partially or entirely within GeoNode.

More specifically, the issue comes from the [_update_user_groups_from_social](https://github.com/GeoNode/geonode/blob/master/geonode/people/adapters.py#L282-L302) method during the `pre_social_login` flow in the `GenericOpenIDConnectAdapter` class. 

Currently, this logic is destructive: regardless of what Azure returns, the code removes the user from all local GeoNode groups before attempting to re-sync. If the provider does not send explicit group claims, the user is left with no groups, effectively wiping any manual assignments made by the GeoNode Admin.

### Proposed solution

To resolve the destructive nature of the current OIDC group sync, we propose introducing a configurable setting, `SOCIALACCOUNT_SYNC_USER_GROUPS_ON_LOGIN`, allowing administrators to choose the logic that fits their organizational policy:

- Choice 1: `FULL_SYNC` (The current and default approach) it maintains the existing behavior for backward compatibility,  wiping all local groups on every login to match the provider's token exactly.
- Choice 2: `SAFE_SYNC` : This approach acts as a safety net. If Azure sends no group data, GeoNode skips the synchronization entirely. However, if Azure sends at least one group, a full "wipe-and-sync" occurs to ensure the identity provider remains the absolute source of truth. 
- Choice 3: `NO_SYNC` : This approach, completely decouples GeoNode group memberships from the Social Provider. In this mode, Azure/OIDC is used exclusively for authentication (identifying the user), while groups are managed by GeoNode. No groups are added or removed during the login flow, ensuring total stability and manual control over user permissions.

By moving to a method-based setting, we provide a more flexible approach to the Social / GeoNode sync in terms of the groups. This allows us to solve the immediate data-loss issue for users with missing OIDC claims while still supporting high-security environments that require strict 1:1 group mirroring.

As the default configuration, the current approach `FULL_SYNC` will be persisted.

Gpetrak:
This PR was created according to this issue: #14107 

## Checklist

For all pull requests:


geonode/people/adapters.py
```diff
@@ -280,23 +280,36 @@ def handle_user_invalid_email(user):
...
     extractor = get_data_extractor(sociallogin.account.provider)
     group_role_mapper = get_group_role_mapper(sociallogin.account.provider)
     try:
         groups = extractor.extract_groups(sociallogin.account.extra_data) or extractor.extract_roles(
             sociallogin.account.extra_data
         )
 
-        # check here if user is member already of other groups and remove it form the ones that are not declared here...
+        # If Azure returns no group data, skip synchronization entirely
+        if sync_strategy == "SAFE_SYNC" and (groups is None or groups == ""):
```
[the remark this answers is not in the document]
Gpetrak: Done

geonode/settings.py
```diff
@@ -2012,6 +2012,9 @@ def get_geonode_catalogue_service():
     SOCIALACCOUNT_OIDC_PROVIDER: SOCIALACCOUNT_PROVIDERS_DEFS.get(_SOCIALACCOUNT_PROVIDER),
 }
 
+# Default to FULL_SYNC to maintain backward compatibility
+SOCIALACCOUNT_SYNC_USER_GROUPS_ON_LOGIN = os.getenv("SOCIALACCOUNT_SYNC_USER_GROUPS_ON_LOGIN", "FULL_SYNC")
```
mattiagiupponi, line 2020: i would add here as a comment the three possibility and what they actually do

geonode/people/adapters.py
```diff
@@ -280,23 +280,40 @@ def handle_user_invalid_email(user):
...
     group_role_mapper = get_group_role_mapper(sociallogin.account.provider)
     try:
-        groups = extractor.extract_groups(sociallogin.account.extra_data) or extractor.extract_roles(
-            sociallogin.account.extra_data
-        )
+        # Try groups first
+        groups = extractor.extract_groups(sociallogin.account.extra_data)
+
+        # If groups are missing, try roles
+        if groups == "":
```
mattiagiupponi: maybe `if not groups` is better, so if the code change and returns `False` / `None` or `[]` here is still compatible

mattiagiupponi (changes requested):

geonode/settings.py
Gpetrak: Done

geonode/people/adapters.py
Gpetrak: @mattiagiupponi I hear you but in this case we need to differentiate the case of `False / None / ""` with an empty list `[]` because an empty list is a valid case. It means that we have the group key but withtout assigned groups for this member. Thus, in case of an empty list the `SAFE_SYNC` we should remove all the local groups. Just to make it clearer, I replaced the `if groups == ""` with `if not isinstance(groups, list)`

sijandh35 (approved):

mattiagiupponi (approved):

PR #14122, opened 2026-04-10
[Fixes #14117] Implement the option to configure the default permissions for registered members and add for anonymous default

### LINKED ISSUE #14117: Implement the option to configure the default permissions for registered members in GeoNode
The same as we do for anonymous users with `DEFAULT_ANONYMOUS_VIEW_PERMISSION` and `DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION`, we want to allow configuring the default permissions for registered members.

This will be the opportunity to deprecate the current ones for the anonymous user, and switch to a configuration that sets the permissions itself, instead of using two mutually-exclusive booleans. 
```
DEFAULT_ANONYMOUS_PERMISSIONS = os.getenv("DEFAULT_ANONYMOUS_PERMISSIONS", None) # Compact permission enum, "view or download"
DEFAULT_REGISTERED_MEMBERS_PERMISSIONS  = os.getenv("DEFAULT_ANONYMOUS_PERMISSIONS", None) # Compact permission enum, "view, download, edit or manage"
```

They will be `None` by default. `None` is a valid compact permission value, but it will be used by the underlying logic to assign no permissions.

sijandh35:
Fixes #14117 

## Checklist

For all pull requests:


geonode/settings.py
```diff
@@ -1868,9 +1868,26 @@ def get_geonode_catalogue_service():
...
 DEFAULT_ANONYMOUS_VIEW_PERMISSION = ast.literal_eval(os.getenv("DEFAULT_ANONYMOUS_VIEW_PERMISSION", "True"))
+# DEPRECATED: use DEFAULT_ANONYMOUS_PERMISSIONS (compact permissions)
 DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION = ast.literal_eval(os.getenv("DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION", "True"))
 
+# Compact permissions for default groups
+# Valid values:
+#  - DEFAULT_ANONYMOUS_PERMISSIONS: view | download | none
+#  - DEFAULT_REGISTERED_MEMBERS_PERMISSIONS: view | download | edit | manage | none
+DEFAULT_ANONYMOUS_PERMISSIONS = os.getenv("DEFAULT_ANONYMOUS_PERMISSIONS", None)
+DEFAULT_REGISTERED_MEMBERS_PERMISSIONS = os.getenv("DEFAULT_REGISTERED_MEMBERS_PERMISSIONS", None)
```
mattiagiupponi: If those two variables are the new default, i think the default value should not be None, but what we want to be the new default like `download`

geonode/geoserver/manager.py
```diff
@@ -244,13 +247,13 @@ def set_permissions(
                                     if not skip_registered_members_common_group(user_group):
                                         create_geofence_rules(_resource, perms, None, user_group, batch)
                                         exist_geolimits = exist_geolimits or has_geolimits(_resource, None, user_group)
-
                                 # Anonymous
-                                if settings.DEFAULT_ANONYMOUS_VIEW_PERMISSION:
+                                anonymous_compact = get_default_anonymous_compact_permission()
```
mattiagiupponi: I think this part should not be longer required. Since we have an handler in the permissions_registry, the permissions should be already internally managed without the need to re-write again the check. This is valid also in other parts of the code

geonode/geoserver/manager.py
```diff
@@ -244,13 +247,13 @@ def set_permissions(
...
+                                if anonymous_compact in (VIEW_RIGHTS, DOWNLOAD_RIGHTS):
                                     create_geofence_rules(_resource, VIEW_PERMISSIONS, None, None, batch)
                                     exist_geolimits = exist_geolimits or has_geolimits(_resource, None, None)
 
-                                if settings.DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION:
+                                if anonymous_compact == DOWNLOAD_RIGHTS:
```
mattiagiupponi: I think this part should not be longer required. Since we have an handler in the permissions_registry, the permissions should be already internally managed without the need to re-write again the check. This is valid also in other parts of the code

geonode/security/models.py
```diff
@@ -198,7 +202,8 @@ def set_default_permissions(self, owner=None, created=False, **kwargs):
         user_groups = Group.objects.filter(name__in=_owner.groupmember_set.values_list("group__slug", flat=True))
 
         # Anonymous
-        anonymous_can_view = settings.DEFAULT_ANONYMOUS_VIEW_PERMISSION
+        anonymous_compact = get_default_anonymous_compact_permission()
+        anonymous_can_view = anonymous_compact == VIEW_RIGHTS
```
mattiagiupponi, line 206: I think this part should not be longer required. Since we have an handler in the permissions_registry, the permissions should be already internally managed without the need to re-write again the check. This is valid also in other parts of the code

geonode/security/models.py
```diff
@@ -211,7 +216,7 @@ def set_default_permissions(self, owner=None, created=False, **kwargs):
                 ):
                     perm_spec["groups"][user_group] = ["view_resourcebase"]
 
-        anonymous_can_download = settings.DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION
+        anonymous_can_download = anonymous_compact == DOWNLOAD_RIGHTS
```
mattiagiupponi, line 219: I think this part should not be longer required. Since we have an handler in the permissions_registry, the permissions should be already internally managed without the need to re-write again the check. This is valid also in other parts of the code

geonode/security/utils.py
```diff
@@ -165,11 +168,11 @@ def get_user_visible_groups(user, include_public_invite: bool = False):
 class AdvancedSecurityWorkflowManager:
     @staticmethod
     def is_anonymous_can_view():
-        return settings.DEFAULT_ANONYMOUS_VIEW_PERMISSION
+        return get_default_anonymous_compact_permission() in (VIEW_RIGHTS, DOWNLOAD_RIGHTS)
```
mattiagiupponi, line 181: maybe here we should use the permissions registry? the permisisons for anonymous should be mananaged already there

geonode/security/utils.py
```diff
@@ -165,11 +168,11 @@ def get_user_visible_groups(user, include_public_invite: bool = False):
...
 
     @staticmethod
     def is_anonymous_can_download():
-        return settings.DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION
+        return get_default_anonymous_compact_permission() == DOWNLOAD_RIGHTS
```
mattiagiupponi, line 185: maybe here we should use the permissions registry? the permisisons for anonymous should be mananaged already there

geonode/context_processors.py
```diff
@@ -27,13 +27,21 @@
...
+    get_default_anonymous_compact_permission,
+)
 
 from allauth.socialaccount.models import SocialApp
 
 
 def resource_urls(request):
     """Global values to pass to templates"""
     site = Site.objects.get_current()
+    anonymous_compact = get_default_anonymous_compact_permission()
```
mattiagiupponi, line 42: maybe here we should use the permissions registry? the permisisons for anonymous should be mananaged already there

mattiagiupponi (changes requested):

geonode/settings.py
sijandh35: Updated.

geonode/geoserver/manager.py
sijandh35: Updated.

geonode/geoserver/manager.py
sijandh35: Updated.

geonode/security/models.py
sijandh35: Hi @mattiagiupponi, 
When I checked this flow, this block runs **before** the handler applies default anonymous permissions.  
First, `set_default_permissions` logic is executed ([link](https://github.com/GeoNode/geonode/blob/4327fd8befb86d7e525b64bc71752d47df0db607/geonode/security/models.py#L177)), and only later permissions are finalized through `set_permissions`, where handlers run ([link](https://github.com/GeoNode/geonode/blob/4327fd8befb86d7e525b64bc71752d47df0db607/geonode/security/models.py#L228)).
So at this earlier step we still do not have handler-enriched anonymous permissions yet.  
To keep behavior aligned with the previous implementation, I kept this part reading from settings/compact resolver at this stage.What do you think?

geonode/security/models.py
sijandh35: Same as https://github.com/GeoNode/geonode/pull/14122#discussion_r3146296142

geonode/security/utils.py
sijandh35: `is_anonymous_can_view()` currently has no `instance/uuid` context, while `permissions_registry` is resource-scoped.  
So in its current signature, it cannot reliably fetch effective anonymous permissions from the registry.
For now I switched it to the compact resolver (`get_default_anonymous_compact_permission()`), so it is aligned with the new settings model.

geonode/security/utils.py
sijandh35: Same as https://github.com/GeoNode/geonode/pull/14122#discussion_r3146302128 .

geonode/geoserver/manager.py
```diff
@@ -244,13 +245,27 @@ def set_permissions(
...
+                                    if anonymous_group
+                                    else []
+                                )
+
+                                if "view_resourcebase" in anonymous_perms:
                                     create_geofence_rules(_resource, VIEW_PERMISSIONS, None, None, batch)
                                     exist_geolimits = exist_geolimits or has_geolimits(_resource, None, None)
 
-                                if settings.DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION:
+                                if "download_resourcebase" in anonymous_perms:
```
mattiagiupponi: better to use [DOWNLOAD_PERMISSIONS](https://github.com/GeoNode/geonode/blob/9493c2c408f006e5b7620b5fbeb09a309d4d3071/geonode/security/permissions.py#L84)

geonode/geoserver/manager.py
```diff
@@ -244,13 +245,27 @@ def set_permissions(
...
+                                    ),
+                                    None,
+                                )
+                                anonymous_perms = (
+                                    effective_permissions.get("groups", {}).get(anonymous_group, [])
+                                    if anonymous_group
+                                    else []
+                                )
+
+                                if "view_resourcebase" in anonymous_perms:
```
mattiagiupponi: better to use [VIEW_PERMISSIONS](https://github.com/GeoNode/geonode/blob/9493c2c408f006e5b7620b5fbeb09a309d4d3071/geonode/security/permissions.py#L80)

geonode/geoserver/manager.py
```diff
@@ -244,13 +245,27 @@ def set_permissions(
...
                                         create_geofence_rules(_resource, perms, None, user_group, batch)
                                         exist_geolimits = exist_geolimits or has_geolimits(_resource, None, user_group)
-
                                 # Anonymous
-                                if settings.DEFAULT_ANONYMOUS_VIEW_PERMISSION:
+                                effective_permissions = permissions_registry.get_perms(instance=_resource) or {}
+                                anonymous_group = next(
+                                    (
+                                        g
+                                        for g in effective_permissions.get("groups", {})
```
mattiagiupponi: i think all this block can be simplify with:

```
anonymous_perms = permissions_registry.get_perms(instance=_resource, group=Group.objects.get(name="anonymous"))
```

geonode/security/handlers.py
```diff
@@ -98,6 +108,42 @@ def _has_edit(perms_list, u):
...
+
+        payload = perms_payload or {}
+        payload.setdefault("groups", {})
+
+        _resource_type = getattr(instance, "resource_type", None) or instance.polymorphic_ctype.name
+        _resource_subtype = (getattr(instance, "subtype", None) or "").lower()
+
+        anonymous_compact = get_default_anonymous_compact_permission()
+        anonymous_group, _ = Group.objects.get_or_create(name="anonymous")
+        payload["groups"][anonymous_group] = sorted(
```
mattiagiupponi: in this way we are overriding any permission defined previously with the value that we have in the default... i donno if is the correct way. We should give a default, yes but in this way if any hander before this removed the perms from the group, we are re-assigning them anyway... i think it should evaluate if the permission needs to be re-assigned not reassign them by default. same for the registered members

geonode/security/permissions.py
```diff
@@ -136,6 +132,59 @@
...
+VALID_REGISTERED_MEMBERS_COMPACT_PERMISSIONS = {VIEW_RIGHTS, DOWNLOAD_RIGHTS, EDIT_RIGHTS, MANAGE_RIGHTS, NONE_RIGHTS}
+
+
+def _normalize_compact_permission(raw_value, valid_values, setting_name):
+    if raw_value is None:
+        return None
+    normalized_value = str(raw_value).strip().lower()
+    if normalized_value in ("", NONE_RIGHTS):
+        return None
+    if normalized_value not in valid_values:
```
mattiagiupponi, line 152: I guess this can be replaced by the existing [compact_permission_labels](https://github.com/GeoNode/geonode/blob/9493c2c408f006e5b7620b5fbeb09a309d4d3071/geonode/base/models.py#L942) if the name is not in the dict, an error should be raised.
We can think to externalize it in case

geonode/security/permissions.py
```diff
@@ -136,6 +132,59 @@
...
+        return None
+    return _normalize_compact_permission(
+        raw_value,
+        VALID_REGISTERED_MEMBERS_COMPACT_PERMISSIONS,
+        "DEFAULT_REGISTERED_MEMBERS_PERMISSIONS",
+    )
+
+
+def get_default_anonymous_permissions_list():
+    compact_perm = get_default_anonymous_compact_permission()
```
mattiagiupponi, line 198: there is the [_to_extended_perms](https://github.com/GeoNode/geonode/blob/9493c2c408f006e5b7620b5fbeb09a309d4d3071/geonode/security/permissions.py#L216) to convert the perms back to the real list

mattiagiupponi (changes requested):

geonode/geoserver/manager.py
sijandh35: Updated.

geonode/geoserver/manager.py
sijandh35: Updated.

geonode/security/permissions.py
sijandh35: Hi @mattiagiupponi , here I’m validating against a caller-provided allowed set, because the valid compact values differ between anonymous and registered-members defaults.
That’s why I kept separate allowed-value sets and pass them into this function, instead of using compact_permission_labels, which is a fixed/global map.

geonode/security/permissions.py
sijandh35: Hi @mattiagiupponi , when I tested `_to_extended_perms("download", None, None)`, it returned only `view_resourcebase` because we don’t have resource / resource_type here, so it falls into the else branch in that condition ([link](https://github.com/GeoNode/geonode/blob/87e0715ddd880a72508a43a86a0933a48853adf9/geonode/security/permissions.py#L246-L249)).
To keep behavior aligned with the previous logic (which also included `download_resourcebase`), I kept this part as-is for now.

geonode/geoserver/manager.py
sijandh35: Hi @mattiagiupponi, I checked this approach. When I tried:

`anonymous_perms = permissions_registry.get_perms(instance=_resource, group=Group.objects.get(name="anonymous"))`

it returned an empty permission list for the group here. It looks like this path checks DB-level group permissions ([link](https://github.com/GeoNode/geonode/blob/87e0715ddd880a72508a43a86a0933a48853adf9/geonode/security/registry.py#L238)).
With `permissions_registry.get_perms(instance=_resource)`, instead, I get the effective permissions needed here. So I think keeping this implementation is safer in this case.

geonode/security/handlers.py
sijandh35: Hi @mattiagiupponi ,

This block is executed only on resource creation (`created=True`), so it does not run when permissions are later updated manually (where `created=False`).

**That said, your concern about overriding previous handler output even during creation also is valid. I update this to apply defaults only when the target group is missing instead of always reassigning to make it extra safe for created=True part also.**

geonode/geoserver/manager.py
mattiagiupponi: fixed, was the registry that was not returing the correct perms. suggested approach has been implemented

mattiagiupponi (approved):

PR #14156, opened 2026-04-22
[Fixes #14152] Restrict the creation of remote resources to administrators only by default

### LINKED ISSUE #14152: Restrict the creation of remote resources to administrators only by default
### Overview
GeoNode offers several means to add resources obtained from remote services and URLs to its catalogue. This simplifies the integration of documents and datasets from external sources, but requires strong attention to security. 

GeoNode implements several security layers to keep insecure resources out of its perimeter, but keeping the surface attack as limited as possible is the best guarantee of safety.

For this reason, we're aiming to introduce a restriction to the catalogue functionality related to remote resources. The plan is to restrict them to administrators only by default. A configuration will allow them to revert to the current behaviour.
The affected features are:

  - Remote documents
  - Remote datasets (COGs, FGBs, 3dtiles)
  - Remote Services (WMS, ArcGIS)

### Technical proposal
For the moment, a single global configuration will switch the restriction on/off. Let's call it `REGISTERED_USERS_CAN_ADD_REMOTE_RESOURCES` for the moment.

This configuration will be read inside the `PermissionsHandlerRegistry.get_db_perms_by_user()` method and a new synthetic `add_remote_resource` will be added to the permissions for the current user.

The set of permissions will be returned automatically by the /users API, and the client can use it to handle the frontend tools (like action menus for Remote Services, and the "Add from URL" tab inside the upload pages). 

In the future we can decide to handle the settings as a real permission that can be assigned to users and groups.

nrjadkry:
Fixes #14152 

## Checklist

For all pull requests:

### DIFF
--- a/geonode/services/views.py
+++ b/geonode/services/views.py
@@ -34,7 +34,7 @@ from django.contrib.auth.decorators import login_required
 from geonode.base.models import ResourceBase
 from geonode.harvesting.models import Harvester
 from geonode.security.views import _perms_info_json
-from geonode.security.utils import get_visible_resources
+from geonode.security.utils import get_visible_resources, check_add_remote_resource_perm
 from django.core.cache import caches
 
 from .models import Service
@@ -59,6 +59,8 @@ def services(request):
 
 @login_required
 def register_service(request):
+    check_add_remote_resource_perm(request.user)
+
     service_register_template = "services/service_register.html"
     if request.method == "POST":
         form = forms.CreateServiceForm(request.POST)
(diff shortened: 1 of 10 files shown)


geonode/services/views.py
```diff
@@ -59,6 +59,8 @@ def services(request):
 
 @login_required
 def register_service(request):
+    check_add_remote_resource_perm(request.user)
```
giohappy, line 62: @nrjadkry I think this check should go under the POST case

giohappy: my fault @nrjadkry I thought it was another section of the UI. Of course we don't want to present the form to non-admins by default.
Forget it.

giohappy (approved):

PR #14176, opened 2026-04-28
[Fixes #14172] Hide Remote Services creation actions if user lacks permissions

### LINKED ISSUE #14172: Hide Remote Services creation actions if user lacks permissions
GeoNode now defines a permission to control whether a user has permission to create Remote Services.
The new `add_remote_resource` permission is returned along with the other global user permissions inside the `perms` property of the user endpoint response.

The following button will be visible only if `add_remote_resource` is set

Image

nrjadkry:
Fixes #14172 

## Checklist

For all pull requests:


geonode/services/views.py
```diff
@@ -49,11 +49,16 @@
...
 
     return render(
         request,
         "services/service_list.html",
-        {"services": Service.objects.all(), "can_add_resources": request.user.has_perm("base.add_resourcebase")},
+        {
+            "services": Service.objects.all(),
+            "can_add_resources": request.user.has_perm("base.add_resourcebase"),
+            "can_add_remote_resources": "add_remote_resource" in perms,
+        },
```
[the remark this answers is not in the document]
giohappy: @nrjadkry, this comment looks good. Let's do this optimization

giohappy (changes requested):

giohappy (changes requested):
@nrjadkry there are a couple of things that must be corrected:

- `request.user.has_perm("base.add_resourcebase")}` must be replaced with `PermissionsHandlerRegistry.user_has_perm(...)`. The first was already there, I know, but it was there because the registry didn't exist yet.
- Also `permissions_registry.get_db_perms_by_user` must be replces with `PermissionsHandlerRegistry.user_has_perm(...)`. The first is deprecated, and should also be changed [here](https://github.com/GeoNode/geonode/blob/master/geonode/security/utils.py#L170). `get_db_perms_by_user` could probably be dropped, it was an old compatibility method but apparently it's been replaced everywhere now, except the legacy views for remote services.

geonode/security/registry.py
```diff
@@ -203,61 +174,22 @@ def get_perms(
...
-            if user_cache_key and group_cache_key:
-                cache.set(user_cache_key, result["users"][user], settings.PERMISSION_CACHE_EXPIRATION_TIME)
-                cache.set(group_cache_key, result["groups"][group], settings.PERMISSION_CACHE_EXPIRATION_TIME)
-        elif use_cache and cache_key:
-            cache.set(cache_key, result, settings.PERMISSION_CACHE_EXPIRATION_TIME)
-        else:
-            pass
+        if not instance:
+            if not user:
+                return []
```
mattiagiupponi: i would raise an error rather than returning an empty list.
If one of the two is required, also the docstring should specify it

geonode/services/views.py
```diff
@@ -49,17 +51,28 @@
 
 def services(request):
     """This view shows the list of all registered services"""
+    can_add_resources = permissions_registry.user_has_perm(request.user, perm="add_resource")
+    can_add_remote_resources = False
+    if can_add_resources:
+        can_add_remote_resources = permissions_registry.user_has_perm(request.user, perm="add_remote_resource")
```
mattiagiupponi: i think we should update the permissions registry in order to accept both strings and lists in the `user_has_perm `method.
So, like in this case, the call will be: `permissions_registry.user_has_perm(request.user, perm="add_remote_resource", "add_resource"])`

the registry can be updated as follow:

```python
    def user_has_perm(self, user, instance=None, perm="", include_virtual=False):
        """
        Returns True if the user has the defined perm (permission)
        """
        if not perm:
            return False
        if isinstance(perm, list):
            return all(
                perm
                in self.get_perms(instance=instance, user=user, include_virtual=True, include_user_add_resource=True)
                for perm in perm
            )
        return perm in self.get_perms(
            instance=instance, user=user, include_virtual=True, include_user_add_resource=True
        )
```

the following part can be:

```python
 {
            "services": Service.objects.all(),
            "can_add_remote_resources": permissions_registry.user_has_perm(request.user, perm=["add_remote_resource", "add_resource"]),
        },
```

mattiagiupponi (changes requested):

mattiagiupponi (approved):

giohappy (approved):

PR #14192, opened 2026-05-04
[Fixes #14177] Implement storage and handling of authentications for external resources

### LINKED ISSUE #14177: Implement storage and handling of authentications for external resources
**Description**

GeoNode currently supports Basic Auth for remote services through the `Service` model, and the backend uses this information when harvesting remote OWS resources.

This works for the current remote OWS service flow, but it makes authentication handling tightly coupled to a specific use case and difficult to extend. GeoNode also has rule-handler-based extension points that may need to generate rules based on different authentication types.

Going forward, we may need a more generic and reusable authentication layer for outgoing requests, so that authentication configuration can be reused consistently across backend service integrations and future remote resource types.

The goal is to abstract and decouple authentication configuration and runtime logic so that both rule handlers and other backend services can use the same auth framework.

## High-level design
We need a generic and extensible authentication layer built on top of a Django `AuthConfig` model.

The goal is to decouple:

- how authentication settings are stored in the database
- how authentication is applied to outgoing HTTP requests
- whether authentication is expressed as a [requests-compatible auth object](https://requests.readthedocs.io/en/latest/api/#requests.auth.AuthBase) or as direct request enrichment

This would give us a single, uniform way to instantiate an auth handler from a DB record and then use it to prepare outgoing requests.
Authentication configurations are stored in Django as a model, like:

```python
class AuthConfig(models.Model):
    id: Number
    type: "Basic" | "Token" | "SASUrlsConfig" # extensible to other auth types
    payload: json
```

Where:

- type identifies the auth strategy
- payload stores strategy-specific configuration
- each auth strategy knows how to extract its own fields from payload
- each auth strategy can apply itself to a `requests.Request`

We also need a way to associate URL patterns with an AuthConfig for request-handler-based integrations. A possible approach is to introduce a separate model that maps a URL pattern to an auth configuration, so that request handler configuration can be stored in the database instead of being hardcoded in settings.

```
class URLPatternAuthConfig(models.Model):
    auth_config: AuthConfig
    pattern: string
```

sijandh35:
Fixes #14177 

## Checklist

For all pull requests:


geonode/services/utils.py
```diff
@@ -135,3 +135,30 @@ def parse_services_types():
         custom_services_type_module = import_string(services_type_path)
         custom_services_types = {**custom_services_types, **custom_services_type_module.services_type}
     return custom_services_types
+
+
+def get_basic_auth_credentials(auth_config):
```
giohappy: @sijandh35 I would move these utils under the security module

sijandh35: Updated.

geonode/security/admin.py
```diff
@@ -0,0 +1,87 @@
...
+        super().__init__(*args, **kwargs)
+        self.fields["type"].choices = _get_auth_type_choices()
+        self.fields["password"].help_text = (
+            "Used for basic authentication type. It will be encrypted and stored in the payload password key."
+        )
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if cleaned_data.get("type") == "basic" and not self.instance.pk and not cleaned_data.get("password"):
```
giohappy: Call the registry to obtain the handler for the type, and deledate payload validation to t handler classmethod, like `AuthHandler.validate(cleaned_data)`

geonode/security/admin.py
```diff
@@ -0,0 +1,87 @@
...
+            raise forms.ValidationError("Password is required for basic authentication to encrypt the password.")
+
+        return cleaned_data
+
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+
+        if instance.type == "basic":
```
giohappy: Let's avoid conditions on specific types. This couples the Admin to the implemented handlers.
For safety and simplicity, let's encrypt/decrypt the entire payload.

geonode/security/auth_registry.py
```diff
@@ -0,0 +1,51 @@
...
+        auth_handler_cls = self.registry.get(config.type)
+        if auth_handler_cls is None:
+            raise ValueError(f"Unsupported auth config type '{config.type}'")
+        return auth_handler_cls(config)
+
+
+auth_handler_registry = AuthHandlerRegistry()
+
+
+def build_auth_handler(config):
```
giohappy: we don't need this util. The caller can just use `auth_handler_registry.build(config)` itself.

geonode/security/utils.py
```diff
@@ -52,6 +52,33 @@
...
+    from geonode.security.auth_registry import build_auth_handler
+
+    auth_handler = build_auth_handler(auth_config)
+    if not isinstance(auth_handler, BasicAuthHandler):
+        raise ValueError(f"AuthConfig type '{auth_config.type}' is not supported for service basic credentials")
+
+    return (auth_handler.username, auth_handler.password)
+
+
+def create_basic_auth_config(username, password):
```
giohappy: This can be a class method on the `AuthBasicHandler`

geonode/security/utils.py
```diff
@@ -52,6 +52,33 @@
 logger = logging.getLogger(__name__)
 
 
+def get_basic_auth_credentials(auth_config):
```
giohappy: This can be a class methos on BasicAuthHandler.
However, see my comment inside services...

geonode/services/serviceprocessors/wms.py
```diff
@@ -288,12 +289,16 @@ def __init__(self, url, geonode_service_id=None, *args, **kwargs):
...
+        password = None
+        if service.auth_config:
+            username, password = get_basic_auth_credentials(service.auth_config)
         _parsed_service = get_service_handler(
             cleaned_url.geturl(),
             service.type,
             service.id,
-            username=service.username if service.needs_authentication else None,
-            password=service.get_password() if service.needs_authentication else None,
+            username=username,
```
giohappy: I wouldn't pass username/password around anymore. 
owslib.WebMapService supports passing an auth object (see the signature [here](https://github.com/geopython/OWSLib/blob/master/owslib/wms.py#L22)). So [here](https://github.com/GeoNode/geonode/blob/issue_14194/geonode/harvesting/harvesters/wms.py#L79) and [here](https://github.com/GeoNode/geonode/blob/issue_14194/geonode/harvesting/harvesters/wms.py#L93) we can pass an auth object.
I suggest adapting the handler to handle an auth object in place of username and password.

giohappy (changes requested):

geonode/services/serviceprocessors/wms.py
```diff
@@ -95,11 +97,14 @@ def get_cleaned_url_params(url):
     @property
     def parsed_service(self):
         cleaned_url, service, version, request = WmsServiceHandler.get_cleaned_url_params(self.url)
+        auth = self.kwargs.get("auth")
+        auth_config = self.kwargs.get("auth_config")
+        if auth is None and auth_config is not None:
+            auth = BasicAuthHandler(auth_config).get_request_auth()
```
giohappy: This shouldn't be tied to the Basic auth handler. Any auth type could be used for a service. We must retrieve the correct auth handler from the registry, based on the config.

giohappy (changes requested):

geonode/services/forms.py
```diff
@@ -80,13 +83,21 @@ def clean(self):
...
         url = self.cleaned_data.get("url")
         service_type = self.cleaned_data.get("type")
+        username = self.cleaned_data.get("username", None)
+        password = self.cleaned_data.get("password", None)
         if url is not None and service_type is not None:
             try:
+                auth_config = None
+                if username is not None or password is not None:
+                    payload = {"username": username, "password": password}
+                    BasicAuthHandler.validate(payload)
```
giohappy: Same here @sijandh35 . We create the config from username/passowrd (we only suport basic auth when creating a new service for the moment), but the auth handler can be obtained from the config instead of hardcodign the basichandler

giohappy (changes requested):

geonode/thumbs/thumbnails.py
```diff
@@ -266,9 +266,12 @@ def _get_auth_info(dataset: "Dataset") -> dict:
     """Gets authentication info for a dataset if it's a remote service requiring auth."""
     auth_info = {}
     if dataset.remote_service and dataset.remote_service.needs_authentication:
+        from geonode.security.auth_registry import auth_handler_registry
```
giohappy, line 269: For thumbnails, we can make similar changes to what has been done for the WMSService.
Instead of passing around the username/password, and so limit them to Basic Auth, we can send an auth dowin the chain.
At the end of the chain, you will see that [`owslib.OpenURL` is used](https://github.com/GeoNode/geonode/blob/master/geonode/thumbs/utils.py#L420), which accepts an auth parameter, which we currently set to `None`. We can use it, instead of [preparing the basic auth headers ourselves.](https://github.com/GeoNode/geonode/blob/master/geonode/thumbs/utils.py#L207-L212)

giohappy (changes requested):

giohappy (approved):

PR #14292, opened 2026-06-04
[Fixes #14291] Add parameter for beat DB file location

### LINKED ISSUE #14291: The celery-cmd script should have a parameter for celery-beat-db location
The celery-cmd script currently writes the celery-beat-db file in the same directory, there is no option to specify a different location. This is problematic when running as non-root.

The celery-cmd script does not have a parameter for defining the location of the celery-beat-db file

**Describe alternatives you've considered**
n/a

**Additional context**
I have tested a change in this script, a pull request follows soon for this issue.

jwkaltz:
The celery-cmd script currently writes the celery-beat-db file in the same directory, which is problematic when running as non-root. This pull request adds an option to be able to specify the file location to be used.

## Checklist

For all pull requests:

- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in the documentation) -> small change, so I suppose a CLA is not necessary

### DIFF
--- a/celery-cmd
+++ b/celery-cmd
@@ -16,6 +16,7 @@ CELERY__WORKER_CONCURRENCY=${CELERY__WORKER_CONCURRENCY:-"4"}
 
 # Celery beat settings
 CELERY__BEAT_SCHEDULE=${CELERY__BEAT_SCHEDULE:-"celery.beat:PersistentScheduler"}
+CELERY__BEAT_DB=${CELERY__BEAT_DB:-"/tmp/celerybeat-schedule"}
 CELERY__BEAT_LOG=${CELERY__BEAT_LOG:-"/var/log/celery_beat.log"}
 
 # Harvester settings
@@ -43,6 +44,7 @@ fi
 
 echo "Starting Celery Beat..."
 $CELERY_BIN -A $CELERY_APP beat --scheduler=$CELERY__BEAT_SCHEDULE \
+    --schedule=$CELERY__BEAT_DB \
     --loglevel=$CELERY__LOG_LEVEL -f $CELERY__BEAT_LOG --pidfile=/tmp/celerybeat.pid &
 
 echo "Starting Default Celery Worker..."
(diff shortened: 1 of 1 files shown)


celery-cmd
```diff
@@ -16,6 +16,7 @@ CELERY__WORKER_CONCURRENCY=${CELERY__WORKER_CONCURRENCY:-"4"}
 
 # Celery beat settings
 CELERY__BEAT_SCHEDULE=${CELERY__BEAT_SCHEDULE:-"celery.beat:PersistentScheduler"}
+CELERY__BEAT_DB=${CELERY__BEAT_DB:-"/tmp/celerybeat-schedule"}
```
Gpetrak: @jwkaltz Thank you for the PR.
The definition of `CELERY__BEAT_DB` makes perfect sense. My only concern is the `/tmp` folder.
I think it's risky because if the container restarts, `/tmp` is wiped, meaning Beat loses its task history.
I would propose to use the statics volume instead, which is already writeable by the celery user:
`CELERY__BEAT_DB=${CELERY__BEAT_DB:-"/mnt/volumes/statics/celerybeat-schedule"}"`

jwkaltz: Thanks for checking @Gpetrak , what you wrote makes sense to me, I have adapted the PR according to your suggestion, can you please check?
Another thing (not directly related): it seems there is a copy-paste error here:
https://github.com/GeoNode/geonode/blob/c9ef9d1cb65d93f92cb2a8744ed003fe5546727a/celery-cmd#L25
I think the right-hand part of the assignment should be CELERY__HARVESTER_MAX_MEMORY_PER_CHILD, similar to the lines above? Should I fix that in this PR as well?

Gpetrak: @jwkaltz thanks for the update. 

> Another thing (not directly related): it seems there is a copy-paste error here:
[geonode/celery-cmd](https://github.com/GeoNode/geonode/blob/c9ef9d1cb65d93f92cb2a8744ed003fe5546727a/celery-cmd#L25)
Line 25 in [c9ef9d1](https://github.com/GeoNode/geonode/commit/c9ef9d1cb65d93f92cb2a8744ed003fe5546727a)
CELERY__HARVESTER_MAX_MEMORY_PER_CHILD=${CELERY__MAX_MEMORY_PER_CHILD:-"500000"} 
I think the right-hand part of the assignment should be CELERY__HARVESTER_MAX_MEMORY_PER_CHILD, similar to the lines above? Should I fix that in this PR as well?

I would prefer you to open a separate issue / PR for this so that it is easier for us to track. In the future we are planning more updates to this functionality, so let's keep it clear.

I'll merge this PR as it is.

Gpetrak (approved):

PR #14328, opened 2026-06-12
[Fixes #14309] Generalize remote service type registration and auth handling

### LINKED ISSUE #14309: Generalize remote service registration and auth handling
## Description

GeoNode currently supports registering and importing resources from a fixed set of remote service types, such as WMS and ArcGIS services. These service types are defined statically in the service processor [`get_available_service_types`](https://github.com/GeoNode/geonode/blob/3ad3aa8ae002c5076ca40a65a6f273a98c8841ec/geonode/services/serviceprocessors/__init__.py#L32).

This makes it difficult to register additional remote service types dynamically, while reusing the existing GeoNode `Service` and `AuthConfig` models.

The goal of this issue is to generalize remote service registration in GeoNode by introducing a dynamic service registry.

The registry should allow service types and their handlers to be registered in a common way. Existing GeoNode service types should continue to work as before.

This work should also generalize remote resource authentication handling so that remote resources can reuse `AuthConfig` consistently, instead of keeping authentication logic tied to a specific remote resource type such as WMS.

## Proposed approach

- Introduce a dynamic service registry for remote service types.
- Register existing GeoNode service types through the registry.
- Allow additional service types and handlers to be registered dynamically.
- Keep existing service creation behavior working for WMS and ArcGIS services.
- Move common remote authentication payload handling from [WMS-specific logic](https://github.com/GeoNode/geonode/blob/3ad3aa8ae002c5076ca40a65a6f273a98c8841ec/geonode/upload/handlers/remote/wms.py#L119-L144) to the common remote resource handler.
- Allow remote resources to receive an `AuthConfig` either from the upload payload or from a matching registered remote service.

## Expected outcome

GeoNode should provide a generic mechanism for registering remote service types and reusing authentication configuration for remote resources.

sijandh35:
Fixes #14309 

## Checklist

For all pull requests:


geonode/upload/handlers/common/remote.py
```diff
@@ -110,18 +119,105 @@ def extract_params_from_data(_data, action=None):
...
+    @staticmethod
+    def extract_auth_config_from_data(_data, payload):
+        authentication = _data.pop("authentication", None)
+        if authentication:
+            auth_type = authentication.get("type")
+            auth_payload = authentication.get("payload") or {}
+            auth_handler_cls = auth_handler_registry.get_handler_class(auth_type)
+            if auth_handler_cls is None:
+                raise ValueError(f"Unsupported authentication type '{auth_type}'")
+            auth_handler_cls.validate(auth_payload)
```
mattiagiupponi: maybe we can wrap this into a "create method"

```
auth_handler_cls.create(auth_type, auth_payload)

def create(self, auth_type, auth_payload):
    self.validate(auth_payload)
    auth_config = AuthConfig(type=auth_type)
    auth_config.payload = auth_payload
    auth_config.save()
return auth_config
```

geonode/upload/handlers/common/remote.py
```diff
@@ -91,7 +94,13 @@ def is_valid_url(url, **kwargs):
         and if the url is valid
         """
         try:
-            r = requests.get(url, timeout=10)
+            auth = None
+            if kwargs.get("execution_id"):
```
mattiagiupponi: This part of the code is common between Basic remote, cog,3dtiles and FGB, can't we just place it into a common function anc call it via self?

mattiagiupponi (changes requested):

geonode/upload/handlers/common/remote.py
sijandh35: Updated.

geonode/upload/handlers/common/remote.py
sijandh35: @mattiagiupponi  I now wrapped this in a `create_auth_config` classmethod on `AuthHandler`, so it can be reused by all auth handlers.

geonode/upload/handlers/common/remote.py
```diff
@@ -110,18 +114,114 @@ def extract_params_from_data(_data, action=None):
...
+        for service in services:
+            if not service.service_url:
+                continue
+            service_url = service.service_url.rstrip("/")
+            if url == service_url or url.startswith(f"{service_url}/") or url.startswith(f"{service_url}?"):
+                return service
+        return None
+
+    @staticmethod
+    def get_auth_config_for_import(_exec, service=None, to_update=None):
```
giohappy: @sijandh35 I would simply name it `get_auth_config_from_execution()`

geonode/upload/handlers/common/remote.py
```diff
@@ -110,18 +114,114 @@ def extract_params_from_data(_data, action=None):
...
+            return AuthConfig.objects.filter(pk=auth_config_id).first()
+
+        if service and service.auth_config:
+            if to_update is not None:
+                to_update["auth_config_id"] = service.auth_config_id
+            return service.auth_config
+        return None
+
+    @staticmethod
+    def get_request_auth_for_import(execution_id):
```
giohappy: @sijandh35 I would name it `get_request_auth_from_execution()`

geonode/upload/handlers/common/remote.py
```diff
@@ -110,18 +114,114 @@ def extract_params_from_data(_data, action=None):
...
+    @staticmethod
+    def get_auth_config_for_import(_exec, service=None, to_update=None):
+        if not _exec or not _exec.input_params:
+            return None
+
+        auth_config_id = _exec.input_params.get("auth_config_id")
+        if auth_config_id:
+            return AuthConfig.objects.filter(pk=auth_config_id).first()
+
+        if service and service.auth_config:
```
giohappy: This method requires the exec id otherwise it returns at line 186. But this conditional branch doesn't require the original exection id. 

There's something strange in the signature. You ask for unrelated parameters, where the first is required, but not for the path that uses the second parameter?

geonode/upload/handlers/common/remote.py
sijandh35: Updated.

geonode/upload/handlers/common/remote.py
sijandh35: Updated.

geonode/upload/handlers/common/remote.py
sijandh35: Updated. The service fallback is now handled by the caller, and this method now only resolves values coming from the execution.

geonode/upload/handlers/common/remote.py
```diff
@@ -110,18 +114,112 @@ def extract_params_from_data(_data, action=None):
...
+        _exec = orchestrator.get_execution_object(exec_id)
+        auth_config_id = _exec.input_params.get("auth_config_id")
+        if auth_config_id:
+            AuthConfig.objects.filter(
+                pk=auth_config_id,
+                authconfigresources__isnull=True,
+                url_patterns__isnull=True,
+            ).delete()
 
     def pre_validation(self, files, execution_id, **kwargs):
```
giohappy: @sijandh35  What is this method for? I see it is only used inside tests.

geonode/upload/handlers/common/remote.py
```diff
@@ -110,18 +114,112 @@ def extract_params_from_data(_data, action=None):
...
+        if not auth_config and service and service.auth_config:
+            auth_config = service.auth_config
+            to_update["auth_config_id"] = service.auth_config_id
+        if service and auth_config:
+            to_update["remote_service_id"] = service.id
+            _exec.input_params.update(to_update)
+            _exec.save()
+
+    @staticmethod
+    def find_matching_service(url, user):
```
giohappy, line 171: @sijandh35 this is called by pre_validation, which is not called outside tests.
Please clarify.

Moreover, why "pre-validation"? Is this the code that is supposed to match a service auth-config? In that case I wouldn't call it "pre-validation"....

geonode/upload/handlers/common/remote.py
mattiagiupponi: The [pre-validation](https://github.com/GeoNode/geonode/blob/fe0f3197808cd0db8d87567f7673919b19358404/geonode/upload/celery_tasks.py#L325) is a hook available in the importer, is used to give the possibility to preform some operation on the data, to prepare them for the validation

giohappy: Ok, it's the pre-validation hook that is called by the imported. 
I think pre-validation is not the right naming for what it actually does. Maybe we can create another hook @mattiagiupponi called "pre-config", os something similar?

giohappy: Ok, it's the pre-validation hook that is called by the imported. 
I think pre-validation is not the right naming for what it actually does. Maybe we can create another hook @mattiagiupponi called "pre-config", os something similar?

geonode/upload/handlers/common/remote.py
mattiagiupponi: The [pre-validation](https://github.com/GeoNode/geonode/blob/fe0f3197808cd0db8d87567f7673919b19358404/geonode/upload/celery_tasks.py#L325) is a hook available in the importer, is used to give the possibility to preform some operation on the data, to prepare them for the validation.
In this case, better to use the hook "[prepare_import](https://github.com/GeoNode/geonode/blob/fe0f3197808cd0db8d87567f7673919b19358404/geonode/upload/celery_tasks.py#L330)" which gives the hook to prepare everything that is required to prepare the import of the resource

geonode/upload/handlers/common/remote.py
mattiagiupponi: we can use the "prepare_import" or the "preprocessing" hook instead of the pre-validation one

mattiagiupponi (approved):

geonode/security/auth_handlers.py
```diff
@@ -39,6 +40,9 @@ def _init_from_config(self):
     def get_request_auth(self) -> AuthBase:
         raise NotImplementedError
 
+    def get_gdal_config(self, url):
```
etj: This method name in a generic Auth class is *very* specific.
I would call it `get_config()` and handlers looking for gdal config will call `get_config(...).get('headers', None)` or `get_config(...).get('gdal', None)` or something like that

geonode/services/serviceprocessors/registry.py
```diff
@@ -0,0 +1,85 @@
...
+from django.utils.module_loading import import_string
+from django.utils.translation import gettext_lazy as _
+
+from geonode.services import enumerations
+
+
+class ServiceTypeRegistry:
+    def __init__(self):
+        self.registry = OrderedDict()
+        self._initialized = False
```
etj: Just a style note: we may safely remove the `_initialized` field, and consider the instance not initialized when `registry is None`.

`if self._initialized:` --> `if registry is not None`

`reset()` would become
```py
    def reset(self):
        self.registry = None
```

geonode/services/models.py
```diff
@@ -32,15 +32,18 @@
...
+
 
 logger = logging.getLogger("geonode.services")
 
 
 class Service(ResourceBase):
     """Service Class to represent remote Geo Web Services"""
 
-    type = models.CharField(max_length=10, choices=service_type_as_tuple)
+    type = models.CharField(max_length=10, choices=get_service_type_choices)
```
etj, line 47: Having a dynamic `choices` may require a migration every time the service types list changes.

geonode/services/serviceprocessors/registry.py
```diff
@@ -0,0 +1,85 @@
...
+from collections import OrderedDict
+
+from django.conf import settings
+from django.utils.module_loading import import_string
+from django.utils.translation import gettext_lazy as _
+
+from geonode.services import enumerations
+
+
+class ServiceTypeRegistry:
```
etj, line 28: There's should be either a lazy initialization or not.
Some methods call `self.init_registry()`, other do not.

Let's call init_registry in every method that uses `self.registry` and remove any other external initialization (like the one in `ready()`)

geonode/services/apps.py
```diff
@@ -63,6 +63,9 @@ def ready(self):
         super().ready()
         # Let's make sure the signals are connected to the App
         from . import signals  # noqa
+        from geonode.services.serviceprocessors.registry import service_type_registry
+
+        service_type_registry.init_registry()
```
etj: According to the review comment in `ServiceTypeRegistry` this init should be useless

geonode/services/models.py
```diff
@@ -102,12 +105,15 @@ def service_url(self):
...
     def service_type(self):
         # Return the gxp ptype that should be used to display layers
-        return [x for x in service_type_as_tuple if x[0] == self.type][0][1]
+        service_type = get_available_service_types().get(self.type)
+        if service_type:
+            return service_type["label"]
+        return self.type
 
     def get_absolute_url(self):
         return "/services/%i" % self.id
```
etj: while we're at it, probably here a `reverse` would be good.

geonode/services/tests.py
```diff
@@ -1025,38 +1024,82 @@ def test_will_use_multiple_service_types_defined(self):
...
             },
-        }
-        self.assertDictEqual(expected, elems)
+            registry.registry["CUSTOM"],
+        )
+
+    def test_service_type_registry_should_unregister_service_type(self):
+        registry = ServiceTypeRegistry()
+        registry.register("CUSTOM", handler="path.to.CustomServiceHandler", label="Custom Service")
+
```
etj, line 1087: add and `assertIn`, in order to make sure the `unregister` really unregistered something

geonode/upload/handlers/common/remote.py
etj: @mattiagiupponi you introduced `pre_processing()` but forgot to remove the `pre_validation()` declaration

etj (changes requested):

geonode/services/models.py
sijandh35: Hi @etj, since Django 5.0, field choices can be declared using a callable:
https://docs.djangoproject.com/en/6.0/releases/5.0/#more-options-for-declaring-field-choices

So here the choices are not stored as the evaluated service type list. The migration only [references](https://github.com/GeoNode/geonode/pull/14328/changes#diff-8ed98a700ad06380872937aff5c7375b03873753b496f397dd640f104ca6d2f3R17) the callable method, and the choices are resolved at runtime.So will not make migration everytime service type changes.

geonode/security/auth_handlers.py
sijandh35: Updated with method `get_extra_config` to make it more generic.

geonode/services/serviceprocessors/registry.py
sijandh35: Updated.

geonode/services/serviceprocessors/registry.py
sijandh35: Updated.

geonode/services/apps.py
sijandh35: Updated.

geonode/services/models.py
sijandh35: Updated.

geonode/services/tests.py
sijandh35: Updated.

geonode/upload/handlers/common/remote.py
sijandh35: Hi @etj, I think it was not removed because [BaseHandler](https://github.com/GeoNode/geonode/blob/fe0f3197808cd0db8d87567f7673919b19358404/geonode/upload/handlers/base.py#L44) currently does not define this method, but the upload chain calls it here:
https://github.com/GeoNode/geonode/blob/fe0f3197808cd0db8d87567f7673919b19358404/geonode/upload/celery_tasks.py#L325

If we remove it from `BaseRemoteResourceHandler`, we should add a default `pre_validation()` method to `BaseHandler`, otherwise remote handlers could break when the importer calls the hook. Should i update this ?

etj: Both are called:
https://github.com/GeoNode/geonode/blob/fe0f3197808cd0db8d87567f7673919b19358404/geonode/upload/celery_tasks.py#L323-L325

I would remove the call to `pre_validation()` altogether, but IDK if external core refers to it. Maybe @mattiagiupponi knows better.
If we want to remove the call in a safe way, we may go this way:
```py
if getattr(_datastore, "pre_validation", None):
   _datastore.pre_validation(**kwargs) 
```
but I would prefer to remove the related code if possibile

mattiagiupponi: There are few handlers that uses it (Like the metadata one) but i think we can move the logic into the pre_processing one.
If we want to keep it safe from regressions, we should keep as suggested by @etj

etj: @sijandh35 so, I would go for something like this, in order to track down the origin of the call:
```py
if getattr(_datastore, "pre_validation", None):
   logger.warning("Calling deprecated method pre_validation()", exc_info=True)
   _datastore.pre_validation(**kwargs) 
```

- remove from BaseRemoteResourceHandler
- rename any existing method from subclasses into pre_processing

etj (approved):

PR #14400, opened 2026-07-02
[Fixes #14394] remove support of GoogleCloudStorage

### LINKED ISSUE #14394: Stop the support for Google storage
GeoNode includes optional support for Google backend storage. We do not have direct experience with it, and we have never received feedback from users on it.

For these reasons, Google storage support will be removed from GeoNode.

We invite users who are using it to contribute to its maintenance and, in case, reintroduce its support in the future.

This issue is a follow up of the previous issue https://github.com/GeoNode/geonode/issues/14339

mattiagiupponi:

## Checklist

For all pull requests:


cesarbenjamindotnet:
I believe there hasn't been much progress because this area is not fully implemented yet. Many parts of the system still depend on direct disk access, and currently there is no way to use Google Cloud Storage seamlessly. More importantly, supporting GCS should really be viewed as supporting cloud storage in general, regardless of the underlying provider.

At the moment, many components—especially the importer—still rely on physical file system access. This prevents the application from working properly with cloud storage backends.

Django already has the `django-storages` package, which I believe should be the foundation for provider-agnostic cloud storage support. Rather than removing cloud storage support, the storage layer should be refactored so that it no longer depends on physical disk paths.

I believe the proposal to remove cloud storage support is fundamentally misguided. In reality, cloud storage support is still incomplete, and removing it would be a regression rather than an improvement. In my opinion, it is a very poor proposal.

giohappy:
@cesarbenjamindotnet you're right; the storage still heavily relies on disk access. But I don't get your point against removing the dependency. I know the dependency has never been fully used (same for boto). Removing it doesn't mean saying that the alternative storage will work. It just removes a dependency that is becoming problematic.

giohappy (approved):

PR #14520, opened 2026-08-14
[Fixes #14519] GroupProfile.access default is "public'", not a valid choice

### LINKED ISSUE #14519: GroupProfile.access default is "public'" (stray apostrophe), matching none of its own GROUP_CHOICES
## Expected Behavior

A `GroupProfile` created without an explicit `access` should get one of the field's own `GROUP_CHOICES`, so that code filtering on `access="public"` can see it.

## Actual Behavior

`geonode/groups/models.py`:

```python
access = models.CharField(
    _("Access"), max_length=15, default="public'", choices=GROUP_CHOICES, help_text=access_help_text
)
```

The default is `"public'"` — with a trailing apostrophe. `GROUP_CHOICES` is `[("public", …), ("public-invite", …), ("private", …)]`, so the default is not a valid choice. It is also baked into `geonode/groups/migrations/24_initial.py` as `default=b"public'"`.

The typo dates back to 796add43c ("move groups to core", 2014) and is still on `master`, 4.4.x and 5.0.x.

Most of GeoNode is unaffected because it keys on `access == "private"` rather than `== "public"` (`security/utils.py`, `groups/views.py`, `api/authorization.py`, `people/models.py`). The one place in core that does test for `"public"` is `geonode/people/utils.py`:

```python
# get_available_users
GroupProfile.objects.filter(Q(access="public") | Q(group__in=user.groups.all()))
```

That arm can never match a group created with the default. On an install where the `REGISTERED_MEMBERS_GROUP_NAME` fallback in the same function is populated, the result looks correct by accident, which is presumably why this has gone unnoticed.

## Steps to Reproduce

```python
>>> from geonode.groups.models import GroupProfile
>>> g = GroupProfile.objects.create(slug="demo", title="demo")
>>> g.access
"public'"
>>> GroupProfile.objects.filter(access="public").filter(pk=g.pk).exists()
False
```

Any code path that creates groups without passing `access` is affected. `geonode_ldap`'s `updateldapgroups` is one — on our AD-backed instance all 276 mirrored `GroupProfile` rows carry `"public'"`:

```
GroupProfile count: 276
by access: [{'access': "public'", 'n': 276}]
```

## Your Environment

Observed on 4.4.x (GeoNode 4.4.x base image, GeoServer 2.24.4, Postgres/PostGIS, LDAP via `geonode_ldap`). Verified present on `master` and on the 5.0.x branch by reading `geonode/groups/models.py`.

Fixing the data alone is not enough — the rows get rewritten with the bad value on the next group sync while the model default stands, so it has to be fixed at the model plus a data migration for existing installs. PR follows.

christianbraun:
Fixes #14519.

`GroupProfile.access` was declared with `default="public'"` — a stray apostrophe, matching none of the field's own `GROUP_CHOICES`. Any group created without an explicit `access` was stored with that value and became invisible to code filtering on `access="public"`, notably `geonode.people.utils.get_available_users`. The typo has been there since 796add43c (2014) and is also in `24_initial.py`.

Three changes:

- `geonode/groups/models.py` — `default="public'"` → `default="public"`.
- `geonode/groups/migrations/0036_fix_groupprofile_access_default.py` — the `AlterField` for the new default, plus a `RunPython` that rewrites existing `"public'"` rows to `"public"`. Existing installs need the data fix; the model fix alone leaves every group already created still broken. The `RunPython` is `noop` in reverse, since undoing it would mean writing an invalid value back.
- `geonode/groups/tests.py` — `test_default_access_is_a_valid_choice`, asserting a `GroupProfile` created without an `access` is both a valid choice and returned by `filter(access="public")`.

How we hit it: on an AD-backed instance, `geonode_ldap`'s `updateldapgroups` creates `GroupProfile`s without passing `access`, so all 276 mirrored groups carried `"public'"` and a group picker filtering on `"public"` showed none of them.

Most of core is unaffected because it keys on `access == "private"` instead, which the typo leaves alone — that is why this has survived so long.

## Checklist

For all pull requests:

- [x] PR title must be in the form "[Fixes #`<issue_number>`] Title of the PR"

Happy to backport to 4.4.x / 5.0.x once this lands. I could not run the full suite locally, so CI is the first real execution of the migration.

### DIFF
--- a/.clabot
+++ b/.clabot
@@ -89,6 +89,7 @@
     "jwkaltz",
     "dsuren1",
     "Valyrian-Code",
-    "brynsofz"
+    "brynsofz",
+    "christianbraun"
   ]
 }
--- a/geonode/groups/migrations/0036_fix_groupprofile_access_default.py
+++ b/geonode/groups/migrations/0036_fix_groupprofile_access_default.py
@@ -1,0 +1,38 @@
+from django.db import migrations, models
+
+# The old default, with a stray apostrophe that matches none of the field's own
+# GROUP_CHOICES. Every GroupProfile created without an explicit access — which is
+# what geonode_ldap's group mirroring does — was stored with this value.
+BAD_DEFAULT = "public'"
+
+
+def repair_access(apps, schema_editor):
+    GroupProfile = apps.get_model("groups", "GroupProfile")
+    GroupProfile.objects.filter(access=BAD_DEFAULT).update(access="public")
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("groups", "0035_remove_modeltranslation"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="groupprofile",
+            name="access",
+            field=models.CharField(
+                choices=[
+                    ("public", "Public"),
+                    ("public-invite", "Public (invite-only)"),
+                    ("private", "Private"),
+                ],
+                default="public",
+                help_text="Public: Any registered user can view and join a public group.<br>Public (invite-only):Any registered user can view the group.  Only invited users can join.<br>Private: Registered users cannot see any details about the group, including membership.  Only invited users can join.",
+                max_length=15,
+                verbose_name="Access",
+            ),
+        ),
+        # Reversing would mean writing the invalid value back, so this only goes
+        # forward; the AlterField above is reversible on its own.
+        migrations.RunPython(repair_access, migrations.RunPython.noop),
+    ]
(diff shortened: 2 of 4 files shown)


christianbraun:
The CLA is signed and sent. As the bot notes, @christianbraun still needs adding on your side — could a maintainer take care of that?

Meanwhile `validate (3.12)`, both CodeQL analyses and the Docker image build are green; Smoke Tests are still running, and that job is the first real execution of the new migration anywhere, since `0036` depends on `0035_remove_modeltranslation` and I had no local environment on master to run it against. Happy to adjust if the `AlterField` turns out not to match what `makemigrations` regenerates — I modelled the serialised `help_text` on `0028_auto_20200114_1651.py`.

giohappy:
Thanks @christianbraun for this long-stainding silent bug.
Can you please include the .clabot file with your account appended?

mattiagiupponi (approved):

mattiagiupponi:
backport PR https://github.com/GeoNode/geonode/pull/14531

PR #14404, opened 2026-07-02
[Fixes #14381] Relax Remote Service URL uniqueness

etj:
This PR changes the identity model of remote services from **endpoint URL uniqueness** to **service record uniqueness**.

Before:
- A remote endpoint (`base_url`) implicitly identified one service.
- Duplicate endpoint registrations were rejected/assumed invalid.
- Some lookups and cache entries could collide when the same URL was reused.

After:
- `base_url` is no longer a unique identifier.
- Multiple `Service` objects can intentionally reference the same endpoint (e.g., different owners/use-cases/credentials).
- Service handler caching is keyed by service context (including auth fingerprint), so same-URL services with different credentials are isolated.
- Map layer dataset resolution is stricter: ambiguous URL-based fallback no longer fails silently.
- Registering the same remote endpoint multiple times no longer risks name collisions in harvester/service records; naming is now uniqueness-safe across both models.
 
## What this PR changes

- Removes the uniqueness constraint from `Service.base_url`.
- Allows registering multiple `Service` records pointing to the same remote endpoint.
- Updates service registration/lookup logic and related tests to support duplicate `base_url` values.
- Adds a DB migration to alter the `Service.base_url` field.
- Makes service-handler cache keys credential-aware to avoid collisions across different auth contexts.
- Makes `geoserver_pre_save_maplayer` fallback strict on ambiguous dataset matches (logs + raises on `MultipleObjectsReturned`).
- Extracts service-handler caching into a `ServiceHandlerCache` class (`serviceprocessors/cache.py`), mirroring the existing `ServiceTypeRegistry` pattern, instead of loose module-level functions and a duplicated cache handle in `views.py`.
- Fixes a pre-existing bug in `GeoNodeServiceHandler.parsed_service` (predates this PR) that read auth/type/id off a variable that is actually just the raw `service=` query-string value, not a Service model -- this would raise `AttributeError` the moment the property ran (e.g. registering or probing a GeoNode-type remote service). Auth now resolves the same way `WmsServiceHandler.parsed_service` does.

## Additional fix: remote Datasets have no `store`

Duplicate registrations crashed harvesting on the `(store, workspace, name)` uniqueness, because `store` was derived from the remote URL. `store` is a GeoServer grouping with no meaning for remote datasets, so it is now NULL (migration `layers/0047`, harvesters + indexed service handlers). PostgreSQL treats NULLs as distinct, so remote datasets are exempt from the constraint.

[...]

PR #14532, opened 2026-08-17
[Fixes #14524] Remote-resource detection is inconsistent across harvesters

### LINKED ISSUE #14524: Remote-resource detection is inconsistent across harvesters
Working on #14381 highlighted that `store` and `workspace` are meaningless for Datasets coming from remote services. 
PR #14404 allowed `store` to be NULL. 
`workspace` cannot follow yet, because it is used for unrelated checks inside GeoNode.

The blocker is that "is this dataset remote?" has three answers that disagree, depending on the harvester:

| harvester | `sourcetype` | `subtype` | `remote_service` | `is_remote_resource()` |
|---|---|---|---|---|
| WMS | REMOTE | `remote` | set | True |
| ArcGIS | REMOTE | `remote` | set | True |
| GeoNode legacy | REMOTE | `vector`/`raster` | set | **False** |
| GeoNode current | **LOCAL** | `vector`/`raster` | **never set** | **False** |

Causes:
- `is_remote_resource()` (`resource/utils.py`) keys off `subtype`, but the GeoNode harvesters overwrite it with the remote's real subtype.
- `sourcetype`/`subtype` are only set to REMOTE as a side effect of the remote *not* providing a lat/lon bbox (`harvesting/harvesters/base.py:197-201`).
- `remote_service`/`remote_typename` are only assigned when `sourcetype == REMOTE` (`resource/utils.py:261`), so GeoNode-current-harvested datasets never get them.

Consequence: datasets harvested from a remote GeoNode look local, so they bypass the guards added in `4bc1154ba` and the local GeoFence/GeoServer is called for them. `get_dataset_workspace()` returns the *remote's* workspace name (commonly the literal `geonode`) so `set_permissions`/`remove_permissions` can write and **delete** rules belonging to a local layer with the same name.

**Repro:** register a remote GeoNode service, harvest a dataset whose name matches an existing local layer, change its permissions, inspect GeoFence rules for `geonode:<name>`. Then delete the harvested dataset and observe the local layer's rules disappear.

Tasks in scope, **in this order**:
1. Make remote detection single-sourced: `is_remote_resource()` off `sourcetype`/`remote_service` instead of `subtype`; stop deriving `sourcetype` from `lonlat_extent`; set `remote_service`/`remote_typename` for all harvested datasets.
2. Guard the GeoFence/GeoServer paths accordingly (`geoserver/manager.py`, `geoserver/security.py`).
3. Only then: stop defaulting `get_dataset_workspace()` to `DEFAULT_WORKSPACE` for remote datasets, and audit its call sites (`thumbs/thumbnails.py`, `geoserver/manager.py`, `geoserver/security.py`).
4. Set `workspace` to NULL for remote datasets; and change the ArcGIS `alternate` derivation, which currently builds `f"{workspace}:{name}"` (`harvesting/harvesters/arcgis.py:209,311`).

Note: this is a **pre-existing** bug, not introduced by #14381 - reproducible with a single remote GeoNode service and no duplicate URLs. 
**TBD if needs backport**

Two notes on process: worth labelling this security-adjacent given the rule-deletion path, and step 1 is a behavior change for every existing GeoNode-harvested dataset (they'd flip to `sourcetype=REMOTE`), so it may want a data migration.

etj:
This PR is built upon #14404 since it needs its fixes.

## Checklist

For all pull requests:


etj:
Superseded by https://github.com/GeoNode/geonode/pull/14536

PR #14536, opened 2026-08-18
[Fixes #14381] Service URL uniqueness  [Fixes #14524] Remote-resource detection

### LINKED ISSUE #14381: Relax Remote Service URL uniqueness
Currently, Remote Service assume uniqueness on the remote service URL.
This was for historical reasons, when services were considered a global resource available for everyone, and where any user could harvest resources from the service.

GeoNode is moving in a direction where a Remote Service will be considered a private resource. Every user (where enabled) can create and manage their own remote service and publish resources from it.

This move requires dropping the assumption of URL uniqueness, because multiple users can create their own service for the same remote resource.

The implications of these changes must be investigated carefully, in particular on the connection to the underlying harvestes.

### LINKED ISSUE #14524: Remote-resource detection is inconsistent across harvesters

etj:
Coalesce the commits from:
- https://github.com/GeoNode/geonode/pull/14404
- https://github.com/GeoNode/geonode/pull/14532

## Checklist

For all pull requests:


mattiagiupponi (approved):

etj:
There are still some unrelated pre-existing bugs for arcgis that will be fixed in https://github.com/GeoNode/geonode/issues/14544

PR #14597, opened 2026-09-14
[Fixes #14377] Honor skip_existing_layers flag

### LINKED ISSUE #14377: skip_existing_layers is documented in the upload API as functional but the code says otherwise
Hello team, 

Just want to clear up which of these statement is correct.

We have this line in the code:

https://github.com/GeoNode/geonode/blob/27a4e5d84809098a02f288ab4dd1942fa4887f5d/geonode/geoserver/management/commands/importlayers.py#L77

But the API doc still [list skip_existing_layer as a functional flag. ](https://docs.geonode.org/en/latest/development/api/#datasets)

FWIW, the flag does not work for me when uploading a raster and some json.

```python
endpoint= os.getenv('BASE_URL_GEONODE') + "/uploads/upload/"
upload = {
         'base_file': (os.path.basename(path), open(file, 'rb'), 'application/octet-stream')
      }

payload = {'action': 'upload', 'skip_existing_layers':'true'}
res = requests.post(endpoint, auth=auth, files=upload, verify=False, data=payload)
``` 

The data sent: 

```shell
POST /api/v2/uploads/upload/ HTTP/1.1
User-Agent: python-requests/2.32.5
Accept-Encoding: gzip, deflate, br, zstd
Accept: */*
Connection: keep-alive
Content-Length: 4725
Content-Type: multipart/form-data; boundary=51d3766f5769ab6091be732eb4851968
Authorization: Basic YWRtaW5AbG9jYWxob3N0OmFkbWlu
Host: vspar-dataviz-pp-geoserver-01.afp.com

--51d3766f5769ab6091be732eb4851968
Content-Disposition: form-data; name="action"

upload
--51d3766f5769ab6091be732eb4851968
Content-Disposition: form-data; name="skip_existing_layers"

true
--51d3766f5769ab6091be732eb4851968
Content-Disposition: form-data; name="base_file"; filename="20250405_050646_62_24d1.json"
Content-Type: application/octet-stream
``` 

The response: 

```shell
HTTP/1.1 201 Created
Server: nginx/1.28.0
Date: Mon, 22 Jun 2026 15:43:01 GMT
Content-Type: application/json
Content-Length: 55
Connection: keep-alive
Vary: Accept, Authorization, Accept-Language, Cookie, origin
Allow: POST
Strict-Transport-Security: max-age=3600; includeSubDomains
X-Content-Type-Options: nosniff
Referrer-Policy: strict-origin-when-cross-origin
Cross-Origin-Opener-Policy: same-origin
Content-Language: en-us
access-control-allow-origin: *
Set-Cookie: sessionid=t0k890w2nasgm95cq9de8uxy8q89sebu; expires=Mon, 06 Jul 2026 15:43:01 GMT; HttpOnly; Max-Age=1209600; Path=/; 
SameSite=Lax
Access-Control-Allow-Credentials: false
Access-Control-Allow-Headers: Content-Type, Accept, Authorization, Origin, User-Agent
Access-Control-Allow-Methods: GET, POST, PUT, PATCH, OPTIONS
```

ostyhar:
Fixes #14377.

The upload API exposes `skip_existing_layers`, while the importer checks `skip_existing_layer`. This change normalizes the flag before the execution request is created and adjusts the expected layer count when existing layers are skipped.

It also fixes the raster and 3D Tiles skip paths and adds regression tests for full and partial skips.

## Checklist

For all pull requests:


ostyhar:
The CLA was sent to info@osgeo.org

mattiagiupponi (changes requested):
Hi @ostyhar 
I dont agree much with the implementation, i would suggest to approach it in a different way:
1) the main issue since the parameters is `skip_existing_layer` instead of the one in the serializer `skip_existing_layers` i agree that we should do it, but I think the change should be done on the `extract_params_from_data` for each handler rather than an IF in the view. Understandable that at least is common for everyone, but goes againts the principle to have each handler responsible for the preparation of the parameters required for the execution
2) Rather than leaving the funciton as it is and move logic in the datastore to evaluate the execution, since we want to skip some dataset, i suggest to move the control `should_be_imported` into the `_select_valid_layers` which checks if the layers are valid. Ofc if i need to skip it is not valid, then should be removed from the selected valid layers. We can add a log into the function to evaluate it. In this way when it reaches the `import_resource` there are not valid layers, so then the execution stops.

ostyhar:
@mattiagiupponi , got it. I moved the flag handling into `extract_params_from_data` and did the vector filtering in `_select_valid_layers`.

Just one thing I'm not sure about: when `skip_existing_layers=true` and all valid layers already exist, what should happen? After moving the check into `_select_valid_layers`, the list will be empty, so currently it looks like it would end with No valid layers found.

Should this case finish successfully with 0 imported resources, or should it fail? And should raster and 3D Tiles behave the same way when everything is skipped?

